PR #24, opened 2015-10-10
Expr->closure

trestletech:
- [ ] Need to support expressions for backwards compatibility.
 - [ ] Update the programmatic page on docs to use functions, not expressions.


trestletech:
Manually squash-merged with backwards compat.

Issue #50, opened 2016-07-25
Support parsing JSON on POST


FvD:
As far as I can tell, the only way to POST values to a plumber API is using `&` separated strings. ie:

     curl -i -X POST http://plumber.tres.tl/append/append -d "val=50" 

and there is no support yet for JSON formatted data, ie.:

    curl -H "Content-Type: application/json" -X POST  http://plumber.tres.tl/append/append -d '{"val":50}'

The function `postBodyFilter` filters using `parseQS` that splits on `&` and `=` and does not yet check for JSON formats (perhaps based on an initial `{` and ending `}`.

PR #51, opened 2016-08-24
Add support for simple JSON 

FvD:
This adds basic JSON support to solve issue #50. Following the example in the readme, the following will work after adding this patch:

     curl --data '{"a":4, "b":5}' http://localhost:8000/sum

Since the call is on jsonlite to parse the content of the querystring, more complex JSON should also be possible, but I have not tested that yet.

### DIFF
--- a/R/query-string.R
+++ b/R/query-string.R
@@ -8,29 +8,36 @@ queryStringFilter <- function(req){
 #' @importFrom utils URLdecode
 #' @noRd
 parseQS <- function(qs){
-  if (is.null(qs) || length(qs) == 0 || qs == ""){
+  # Is there data in the request?
+  if (is.null(qs) || length(qs) == 0 || qs == "") {
     return(list())
   }
-  if (stri_startswith_fixed(qs, "?")){
+ # Is it JSON data?
+  if (stri_startswith_fixed(qs, "{")) {
     qs <- substr(qs, 2, nchar(qs))
+    #TODO: add handling of JSON
+  } else {
+    # If not handle it as & separated strings
+    if (stri_startswith_fixed(qs, "?")) {
+      qs <- substr(qs, 2, nchar(qs))
+    }
+
+    qs <- gsub("+", " ", qs, fixed = TRUE)
+
+    parts <- strsplit(qs, "&", fixed = TRUE)[[1]]
+    kv <- strsplit(parts, "=", fixed = TRUE)
+    kv <- kv[sapply(kv, length) == 2] # Ignore incompletes
+
+    keys <- sapply(kv, "[[", 1)
+    keys <- unname(sapply(keys, URLdecode))
+
+    vals <- sapply(kv, "[[", 2)
+    vals[is.na(vals)] <- ""
+    vals <- unname(sapply(vals, URLdecode))
+
+    ret <- as.list(vals)
+    names(ret) <- keys
   }
-
-  qs <- gsub("+", " ", qs, fixed=TRUE)
-
-  parts <- strsplit(qs, "&", fixed=TRUE)[[1]]
-  kv <- strsplit(parts, "=", fixed=TRUE)
-  kv <- kv[sapply(kv, length) == 2] # Ignore incompletes
-
-  keys <- sapply(kv, "[[", 1)
-  keys <- unname(sapply(keys, URLdecode))
-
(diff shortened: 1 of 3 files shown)


R/query-string.R
```diff
@@ -8,29 +8,36 @@ queryStringFilter <- function(req){
...
+  if (is.null(qs) || length(qs) == 0 || qs == "") {
     return(list())
   }
-  if (stri_startswith_fixed(qs, "?")){
-    qs <- substr(qs, 2, nchar(qs))
-  }
-
-  qs <- gsub("+", " ", qs, fixed=TRUE)
+ # Is it JSON data?
+  if (stri_startswith_fixed(qs, "{")) {
```
trestletech: This wouldn't handle having any leading whitespace in front of the query string. I'm not sure if that's common or expected, but would it be worthwhile to trim whitespace from either end above this?

R/query-string.R
```diff
@@ -8,29 +8,36 @@ queryStringFilter <- function(req){
...
+    # Handle JSON with jsonlite
+    ret <- jsonlite::fromJSON(qs)
+    } else {
```
trestletech: Nit: Looks like the indentation is off here. Mind shifting this back one indent?

R/query-string.R
```diff
@@ -8,29 +8,36 @@ queryStringFilter <- function(req){
 #' @importFrom utils URLdecode
 #' @noRd
 parseQS <- function(qs){
-  if (is.null(qs) || length(qs) == 0 || qs == ""){
+  # Is there data in the request?
```
trestletech: Now that this function supports arbitrary data -- not just query strings, what would you think about renaming the function `parseBody`? Or even leaving the old parseQS, introducing a new function for parsing JSON, and then having a higher-level function that could delegate to the appropriate parser as needed?

trestletech:
Looks great! Thanks for contributing. Just a couple of minor concerns.

Please feel free to add yourself as an author in the DESCRIPTION file, if you'd like.

R/query-string.R
FvD: Not at all :)

R/query-string.R
FvD: Do you mean we should not allow arbitrary data in the querystring? I've let this sink, and my feeling is that it perhaps it makes sense prescribe the use of POST for including JSON data in the body, and only accept querystrings in the GET request. 

Perhaps that is what you mean by suggesting to move this functionality to a new function called parseBody. If I understand correctly the higher function would need to check the kind of request (GET or POST) and pass the check for a query string in case of a GET request, and check for a body in case of POST. It does [not seem to make sense to post data with a GET request](http://stackoverflow.com/questions/978061/http-get-with-request-body).

R/query-string.R
FvD: I would only expect a data payload on POST in the form of eg. `-d {"val":50}`. It goes hand in hand with you other remark on whether or not we should keep this in parseQS.

FvD:
Thanks for you comments and your invitation to the DESCRIPTION file! I'm closing this pull request, as I have incorporated the changes in a new branch called `issue50`. I you a pull request when I'm finished.

R/query-string.R
trestletech: The original motivation for this feature was to support browser `<form>`s which tend to use the key/value (querystring) format in a POST format. I think it's important to maintain support for that so browser forms can easily get data in. I do agree that most APIs would expect JSON, though.

R/query-string.R
trestletech: The original motivation for this feature was to support browser <form>s which tend to use the key/value (querystring) format in a POST format. I think it's important to maintain support for that so browser forms can easily get data in. I do agree that most APIs would expect JSON, though.

I think your detection algorithm is reasonable, my comment was just about the name of the function. `parseQS` sounds like it only parses query strings. In reality, this `parseQS` function now parses either query strings or JSON. I think that's the right behavior, but the function name is now a misnomer. I was just hoping to bring the function name up to speed with what the current behavior of the function is.

I think the behavior you have now is right -- users should be able to POST query strings or JSON and have it just work.

R/query-string.R
trestletech: 👍

PR #62, opened 2016-11-06
Add an entrypoint for width to the png processor

statwonk:
I'd like to expose parameters like `width` to [the `png` processor](https://github.com/trestletech/plumber/blob/c82f6b3418858b07c6ea60f19c175967fc6fd708/R/processor-image.R#L23).

My idea here is to use a `filter` like,

```r
#* @filter plot-width
function(req, width=480){
  req$width <- as.numeric(width)
  forward()
}
```

I'm still figuring out a nice general way to expose processor params anyone have ideas about how to do that?

### DIFF
--- a/R/processor-image.R
+++ b/R/processor-image.R
@@ -24,7 +24,12 @@ PlumberProcessor$new(
   function(req, res, data){
     t <- tempfile()
     data$file <- t
-    png(t)
+    if(!is.null(req$width)) {
+      width <- req$width
+    } else {
+      req$width <- 480
+    }
+    png(t, width = req$width)
   },
   function(val, req, res, data){
     dev.off()


trestletech:
That's an interesting question; I don't have a great answer off the top of my head. It seems reasonable that you might want to pass some data from a request into a processor. 

Are you explicitly looking to make the width variable? Or would it be enough to be able to accept the params as a hard-coded property of the annotation e.g.

```
#* Get a graph of the values
#* @png (width=420)
#* @get /graph
function()...
```

statwonk:
@trestletech I'm looking to make width variable. I bet there are many other use cases that require this pattern. My initial thought was to attach the argument to the output of the function as an attribute. Seems then you'd need a closure to hold that bit of logic. I'm  not yet seeing an elegant path.

statwonk:
We also need a way to send the arguments to the correct function rigorously.

trestletech:
I don't love the idea of having a magic object hanging off of the req that allows you to manage this.  For now, the best option I've found is to write a custom serializer if you need to support dynamic widths.

I've documented the current state of the world here, so closing this PR for now.
https://www.rplumber.io/docs/rendering-and-output.html#customizing-image-serializers

Thanks again for your input and contributing! Feel free to open an issue if you want to discuss further about how one might approach this./

PR #64, opened 2016-11-28
Fix invalid R function example in documentation

Marlin-Na:
Add parentheses to the return function.


trestletech:
Thanks for contributing!

PR #73, opened 2017-02-01
Added custom pm2 interpreter info

sellorm:
Updated to include info on using a custom interpreter with pm2. In the earlier version of the docs a shell script wrapper was used for the service, however, in current versions of pm2, this prevents proper monitoring as the R code is run in a sub-shell and is therefore detached from the monitoring system. If a custom interpreter is specified however, it is possible to implement a pure R solution that also works on the command line via Rscript.


trestletech:
Looks great! Thanks for contributing, Mark.

Issue #75, opened 2017-02-28
Provide a clean way to upload files via an endpoint


sellorm:
Posting whole files to an endpoint, using the "multipart/formdata" enctype is not currently well supported.

It is possible to retrieve a posted file (see https://github.com/sellorm/plumber-uploader for an example), but this is hacky and fragile at best. This approach basically rips the file name and content from `req$postBody`.

It would be useful to have something like a `req$FILES` that could be parsed to obtain filenames and content, for processing by the endpoint author.

Not sure how you'd handle things like accepted file types, or max permitted file size and so on though.

FvD:
I had not seen your ticket when I sent some code for review yesterday (see #77). In my notes I suggest something that seems similar to what you are suggesting:

>We move out all the code related to feather, so that the raw binary body is returned to be handles by the function written in the endpoint. We can return the name of the handle and the name of the file, as both are available in the multiparse object and the tmpfile location of the binary file (as set by tempfile())

That would pretty much do what you suggest right? If so I will change the submission to make it generic, so that it just returns the binary, to be handled by the endpoint.

sellorm:
My use case was different, but I think that any solution will end up being applicable to both use cases.

trestletech:
I think #77 is the ground work for this feature. This would be really nice but I think there's a decent chunk of work to get us there.

liori:
Is there any problem with using `Rook::Multipart` for that? I've just tried doing:

```
#' @post /echo
function(req){
  list(formContents = Rook::Multipart$parse(req))
}
```

And it seems to work with `curl -v -F foo=bar -F upload=@somefile.txt  http://localhost:8993/echo` just fine.

ZJUguquan:
@liori So cool !!
And we can just get any kinds of data by doing like:
```r
somefile <- readLines(con = formContents$upload$tempfile)
```
If the file is some kind of **binary file**, then we can do:
```r
somefile <- readBin(con = formContents$upload$tempfile, "raw", 
                    n = file.size(formContents$upload$tempfile))
```

jiangnight:
@liori  @ZJUguquan How can i deal with multiple file from front end

alimanager:
i can parse my object file with : `Multipart$parse`.i can't pasre my json tempfile to read with `readLines` or `capture.output `. 
i wonder if there are a official fonction now in plumber to read upload text file?

stewartthomasj:
[...]

PR #77, opened 2017-03-01
Added support for feather files in POST

FvD:
Hi Jeff, 

This was not an open request in the issues, but was something that came up in a plumber use case: the need to be able to send feather (binary) files. I would much appreciate your review, as I think this may not be the most elegant solution. But it works, provided you upload the file as multipart/form-data:

     curl -X POST http://localhost:9080/ -F 'myfile=@mydata.feather'   

It then uses `webutils` from Jeroen Ooms to parse the multipart content into its different components and writes out a temporary file to read it back in using feather. 

What I have not been able to do (yet) is write a test for this. I've tried to copy an example multipart upload out of the debug session (by transforming rawToChar()) but for some reason I cannot get it right. Any pointers there are welcome.


FvD:
For some reason the installation of `feather` failed in Travis CI. I'll have a look why. 

But before making any changes, i'd like to know you opinion on generalizing this to more (all?) binary formats. Either we move out all the code related to feather, so that the raw binary body is returned to be handles by the function written in the endpoint. We can return the name of the handle and the name of the file, as both are available in the multiparse object and the tmpfile location of the binary file (as set by `tempfile()`

The other option is to include the handling of files here, either based on `Content-type` (that has to be included then by the user in the POST request) or the file extension (that has to be correct, otherwise it will fail, but this is seldom an issue). The trouble there is that there are many different formats, and why should plumber enforce how to handle XLSX files or ODS files, for instance. It may be much better to leave all that to the code in the endpoint.

R/post-body.R
```diff
@@ -24,3 +30,27 @@ parseBody <- function(body){
...
+  if (is.null(body) || length(body) == 0 || body == "") {
+    return(list())
+  }
+  # This is something odd with webutils::parse_multipart. For some reason it
+  # does not work when the boundary is set to the content type as per the
+  # webutils docs.
+  boundary <- "-----------------"
+  parsed_binary <- webutils::parse_multipart(body, boundary)
+  file_name <- parsed_binary[[1]][[1]]
+  file_data <- parsed_binary[[1]][[2]]
```
trestletech: Am I understanding correctly that we could iterate here to support more than one file?

R/post-body.R
```diff
@@ -24,3 +30,27 @@ parseBody <- function(body){
...
+
+  tmpfile <- tempfile()
+  writeBin(file_data, tmpfile)
```
trestletech: Would be neat if we could keep this off disk using textConnections, but I know some packages struggle with connections and prefer files, so maybe landing this stuff on disk is the more generic solution.

DESCRIPTION
```diff
@@ -18,7 +18,9 @@ Imports:
     R6 (>= 2.0.0),
     stringi (>= 0.3.0),
     jsonlite (>= 0.9.16),
-    httpuv (>= 1.2.3)
+    httpuv (>= 1.2.3),
+    feather,
```
trestletech: I'd hope to keep feather off of `imports` and in `suggests`. See how I handle the `htmlwidgets` package in this PR: https://github.com/trestletech/plumber/pull/82/files

We could just suggest and then, if a user goes to use a feature that requires a particular package and doesn't have it, we'd abort at runtime.

R/post-body.R
```diff
@@ -24,3 +30,27 @@ parseBody <- function(body){
...
+  # easy to extend if more binary formats are required
+  # Is there data in the request?
+  if (is.null(body) || length(body) == 0 || body == "") {
+    return(list())
+  }
+  # This is something odd with webutils::parse_multipart. For some reason it
+  # does not work when the boundary is set to the content type as per the
+  # webutils docs.
+  boundary <- "-----------------"
+  parsed_binary <- webutils::parse_multipart(body, boundary)
```
trestletech: I'm unfamiliar with the specs here but this feels pretty hacky. I wonder if this is something that we should just fix in webutils directly?

R/post-body.R
FvD: For my purpose I would prefer to be able to land the files on disk. If we use textConnections, the data would not be persistent at all, and not be available after a restart of the service. 

I can imagine use cases where persistence would not be an issue. Maybe this is something to make a separate ticket of, to add as an option later.

DESCRIPTION
FvD: Absolutely! I had not looked carefully enough at how you are handling this. Your solution is elegant. Also, perhaps I should focus on a more generic solution to POST binary files in plumber and leave the handling of feather or other formats to the API code.

R/post-body.R
FvD: There is nothing wrong with webutils. It was just poorly implemented on my part. The full boundary is part of the req$CONTENT_TYPE so we can fetch each current boundary like so:

    boundary <- strsplit(req$CONTENT_TYPE, split = "=")[[1]][2]

R/post-body.R
FvD: I'm  not sure yet. The list indexes used here are not the best way of accessing the parsed content. This looks better (`myfile` is a name assigned by `webutils::parse_multipart()`)

    parsed_binary$myfile$filename
    parsed_binary$myfile$value

For readability I've now implemented it as named lists, and have yet to figure out yet how to parse multiple files.

FvD:
Hi Jeff,

I've pushed the update based on your review. I  could not  implement everything yet, perhaps I can set up tickets to tackle them in the future.

Things changed:
- [x] Made a generic solution for binary files
- [x] Made a reference to the boundary in CONTENT_TYPE for use in `webutils::part_multipart()`
- [x] Do not clutter imports but move additional packages to suggest and warn user when suggested package is required.
- [x] Added specific use case of feather files to an example

Things no changed (yet)
- [] Accept and parse multiple files
- [] Leave data in-memory until user decides to write to disk
- [] Specify supported (accepted?) extensions in the call to `$run()`

I think this will be an answer to #75 and #126 that is workable. But by all means send this back for changes if you a need for further improvements before merging.

FvD:
Travis fails with the message:
```
checking package dependencies ... ERROR
Namespace dependency not required: ‘crayon’
```
Crayon is outside of the scope of the changes I have made. Should I merge with upstream and commit again?

yangbenfa:
any example about upload the file?



@post /upload
function(req){
  cat("Files uploaded: ", names(req$files))
}
curl -X POST http://localhost:1371/upload -F 'req=./test.txt'

FvD:
HI @yangbenfa , I included an example of binary uploads in the pull request. You can find it in the folder [plumber/inst/examples/13-binary-upload/](https://github.com/FvD/plumber/tree/feather/inst/examples/13-binary-upload). Please note the following two things:

1. For it to work you have to install from the feather branch in my repository - this pull request is still under review and has not been merged yet.
2. I have only tested this with binary files. I'm not sure if a text file would work, but I would love to know whether it does. If it does not work, please let me know.

The examples look as follows:

```  r
# To upload and use the content of the file in a function:
# curl -X POST http://localhost:9080/upload -F 'myfile=@test.feather'

#* @post /upload
function(name, data) {
  # work with binary files after upload (in this example a feather file)
  content <- read_feather(data)
  return(content)
}

# To upload and use the properties of the file in a function:
# curl -X POST http://localhost:9080/inspect -F 'myfile=@test.feather'

#* @post /inspect
function(name, data) {
  file_info <- data.frame(
    filename = name,
    mtime = file.info(data)$mtime,
    ctime = file.info(data)$ctime
    )
  return(file_info)
}
```

R/post-body.R
```diff
@@ -1,11 +1,21 @@
 postBodyFilter <- function(req){
-  body <- req$rook.input$read_lines()
-  args <- parseBody(body)
+
+  if (!is.null(req$CONTENT_TYPE) && grepl("multipart/form-data; boundary=",
+                                          req$CONTENT_TYPE, fixed = TRUE)) {
+
+    boundary <- strsplit(req$CONTENT_TYPE, split = "=")[[1]][2]
```
trestletech: Is this vulnerable to having an `=` sign in the body of the message?

R/post-body.R
```diff
@@ -24,3 +34,33 @@ parseBody <- function(body){
...
+        stop("The webutils package is not available but is required in order
+          to use the functionality to POST binary files",
+             call. = FALSE)
+      }
+      # Is there data in the request?
+      if (is.null(body) || length(body) == 0 || body == "") {
+        return(list())
+      }
+      parsed_binary <- webutils::parse_multipart(body, boundary)
+      file_name <- parsed_binary$myfile$filename
```
trestletech: I'm having a hard time finding where `myfile` comes from? Is that baked in to `parse_multipart`'s return value?

R/post-body.R
```diff
@@ -24,3 +34,33 @@ parseBody <- function(body){
...
+      file_data <- parsed_binary$myfile$value
+
+      tmpfile <- tempfile()
```
trestletech: Hm. It's unfortunate that we're not able to clean up these temp files when we're done. But if we added an `on.exit` here the file would be gone before we could access the file downstream. Maybe I need to add a hook to the response object so that we can clean up any associated resources when it's done.

R/post-body.R
```diff
@@ -24,3 +34,33 @@ parseBody <- function(body){
...
+      if (!requireNamespace("webutils", quietly = TRUE)) {
+        stop("The webutils package is not available but is required in order
+          to use the functionality to POST binary files",
+             call. = FALSE)
+      }
+      # Is there data in the request?
+      if (is.null(body) || length(body) == 0 || body == "") {
+        return(list())
+      }
+      parsed_binary <- webutils::parse_multipart(body, boundary)
```
trestletech: Are there limits built in to any of these on how big of a file we'd attempt to parse? Anything to prevent a client from sending a 10GB file and running the server out of RAM?

R/post-body.R
```diff
@@ -24,3 +34,33 @@ parseBody <- function(body){
...
+      file_name <- parsed_binary$myfile$filename
+      file_data <- parsed_binary$myfile$value
+
+      tmpfile <- tempfile()
+      writeBin(file_data, tmpfile)
+
+      ret <- NULL
```
trestletech: Is this any different than `ret <-list(name = file_name, data=tmpfile)`?

It surprised me that you could assign to a field on a NULL objec.t

trestletech:
Oh and the one other high-level concern I had was around the hard-coding (I think) of the file name. I'd love to make that more generic before we bring it in.

EDiLD:
Any updates on this feature?

meztez:
#550 adds support for any type of file, this should cover this use case and I think this one can be closed. Documentations to follow

PR #78, opened 2017-03-03
Update query-string.R

pietervosnl:
This line removes any '+' character from both the url and the body. In my case it removes the '+' from my JSON what I post in the body. Perhaps it's better to replace the %20 with a space?

### DIFF
--- a/R/query-string.R
+++ b/R/query-string.R
@@ -15,7 +15,7 @@ parseQS <- function(qs){
     qs <- substr(qs, 2, nchar(qs))
   }
 
-  qs <- gsub("+", " ", qs, fixed = TRUE)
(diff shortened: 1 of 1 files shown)


trestletech:
Thanks for contributing this -- sorry it took a while to get to it.

I agree that the current behavior is wrong, so we probably need to get that fixed. plumber definitely shouldn't be stripping `+` characters from your JSON body. I'm thinking I should maybe just delete that line and not do any special parsing (or do wholistic URL decoding).

Since I'm not sure what solution we want to implement, I'm going to go ahead and close this PR for now, but I definitely want to investigate this problem with you so I opened issue #81. Can we continue the conversation there?

Specifically, if you're able to share your JSON body that's not being parsed correctly, that would be a great start.

Thanks again.

PR #80, opened 2017-03-07
Change from jsonlite to rjson

teekaay:
This pull request changes the default JSON deserializer from `jsonlite` to `rjson`. 
The former deserializes everything to lists by default, even when just providing a single
number or string. The `auto_unbox` option is not a solution since sometimes you really 
have a list or vector of one element.

As `rjson` has the same `toJSON` function and API, the switch was made through only renaming.

Now, e.g. `3` is serialized as `"3"` and not `"[3]"` and vectors are also handled properly.


Bart6114:
I'm following this package's notifications a bit from the side line, so I have no say in this whatsoever. But, this seems like a mistake to me. I don't believe there are good arguments to choose `rjson` over a very mature `jsonlite`.

What you are probably referring to is the need to be able to specify the `auto_unbox` argument if the default value does not suit your needs.

trestletech:
Hi @teekay. Thanks for taking the time to write this up and contribute it.

Unfortunately I'm not totally sold on this change. First, it's a backwards incompatible change that could break all existing plumber APIs out there, so it would require a major version bump and might create a lot of headaches for a lot of users.

I do agree that plumber needs a way to handle deserializing scalars in a more explicit way, though. I'm just not sure that this is the approach that I want to take to accomplish it. 

I'll chew on this issue a bit more and see what I can come up with. Thanks again, and I hope you'll continue to consider contributing next time!

PR #86, opened 2017-04-07
Swagger definition generation

trestletech:
When a plumber file is executed with `run(swagger=TRUE)`, create a swagger file available at `/swagger.json` documenting the API, and also host the swagger UI tool at `/__swagger__/` to create an interactive client that allows you to interact with your API in the browser.

### Before Merging

 - [x] parameter definitions, perhaps `@param flarg:int* Flarg`

```
#' @param id The ID of the car to get list(
#'   required = TRUE,
#'   type = "integer"
#' )
#' @param id:int* The ID of the car to get
#' @param id* The ID of the car to get list(type="integer")
```

But then we might mis-parse any comment that ends in `list()` (maybe that's ok) Alternatively...

```
#' @param id The ID of the car to get `
#'   required = TRUE,
#'   type = "integer"
#' `
#' @param id* The ID of the car to get `type="integer"`
```

That's a lot of quoting...

 - [x] Figure out inclusion of Apache 2 code in MIT project
 - [x] Overarching API description and title fields
 - [x] Tests
 - [x] Response code definitions
 - [x] Put swaggerui on CRAN or resolve pointers to GitHub repo in local DESCRIPTION
 - [x] Update the default URL in the swagger UI
 - [x] What's the syntax to bring your own Swagger API?
 - [x] Enable based on if(interactive)
 - [x] Fix the cars example

### After Merging 

 - [ ] Docs
 - [ ] Tag definitions & endpoint tags for organization in UI
 - [x] Extended API metadata fields -- license, contact, etc.
 - [x] Host/port customization in the swagger file
 - [x] Host/port templating in the index.html file that we load in the swagger UI -- don't always assume port 8000
 - [ ] security definitions using filters
 - [ ] Schema definitions of objects -- perhaps using R6? -- to reduce the boilerplate/redundancy of POST/PUT param definitions
 - [ ] Generate "produces" definition on each endpoint based on serializer
 - [ ] `consumes` field definition
 - [ ] Customization of `in` -- where the variables are expected.
 - [ ] Infer the `default` parameter for the swagger endpoint based on the function signature.


trestletech:
sounds like I'm fine to include Apache 2 in MIT, but still unclear how to call it out appropriate in either the README or in CRAN policies.

From https://www.quora.com/Is-the-MIT-license-compatible-with-the-Apache-License-Version-2-APLv2 and https://www.whitesourcesoftware.com/whitesource-blog/top-10-apache-license-questions-answered/?utm_content=Is-the-MIT-license-compatible-with-the-Apache&utm_medium=social&utm_source=quora&utm_term=blog-top-10-apache-license-questions-answered

> MIT is one of the most permissive free software licenses. Basically, you can do whatever you want with a software licensed under the MIT license - just make sure that you add a copy of the original MIT license and copyright notice to it.
> The Apache License is also a permissive license. However, it has stringent terms when it comes to modifications. It requires you to explicitly list out all the modifications that you’ve done in the original software, i.e., you’re required to preserve modification notices. The Apache License also states clearly that you can’t name your product in any way that hints at the product being endorsed by Apache. So you can call your product “SuperWonderServer powered by Apache” but not “Apache SuperWonderServer”. The MIT license doesn’t impose any such terms.
The MIT license is also gaining popularity with developers due to its short and clear license agreement, in contrast to the Apache license agreement. Although the Apache License is nearly identical in terms of what you can and can't do, it is much more heavily "lawyered" and is significantly more verbose. Heck, the appendix alone, which explains how to apply the license, is longer than the entire MIT license.

trestletech:
Alternative suggestion to move Apache 2 code to another package that's explicitly licensed. There is some prior art here: https://twitter.com/noamross/status/863193899170508800

It could even be made an optional `Suggests` in the interest of keeping the install size small. I suppose it depends on how broadly this feature is going to be used.

trestletech:
There are so many properties of a Swagger parameter that I can't imagine it would make sense to try to break them all down into Roxygen-style docs. I think it would make more sense to just take arbitrary JSON/YAML on a parameter (maybe as the first thing after the name) and then just pass that through.

We could do reasonable inference internally to e.g. add required to path args and set the type for anything which is typed for us, but otherwise maybe the less magic the better. Don't reinvent swagger just let users control their swagger definition themselves.

PR #102, opened 2017-05-24
Let plumber take either a file or a directory

scottmmjackson:


R/plumber.R
```diff
@@ -493,7 +493,13 @@ plumber <- R6Class(
 
 #' @rdname plumber
 #' @export
-plumb <- function(file){
+plumb <- function(file, dir){
+  if(!xor(missing(file), missing(dir))){
+    stop("plumber needs only one of a file or a directory with a `plumber.R` file in its root.")
+  } else if (missing(file)){
+    stri_trim_right(dir, pattern = "[/]")
```
trestletech: There's no way you could have known this, but I'm hoping to eliminate the stringi dependency in the future. Can we use R primitives for this instead?

I think:

```
dir <- sub("/$", "", dir)
```

R/plumber.R
```diff
@@ -493,7 +493,13 @@ plumber <- R6Class(
...
+    file <- stri_join(dir,"plumber.R", sep="/")
```
trestletech: Same `stri` concern, and I'm not sure if this will work on Windows. Let's use `file.path` instead.

tests/testthat/test-plumber.R
```diff
@@ -37,6 +37,21 @@ test_that("Invalid file fails gracefully", {
   expect_error(plumber$new("asdfsadf"), regexp="File does not exist.*asdfsadf")
 })
 
+test_that("#plumb accepts a directory with a `plumber.R` file", {
+  # works without trailing slash
+  r <- plumb(dir = 'files')
+  expect_equal(length(r$endpoints), 1)
```
trestletech: You don't need to care about the internal details, but this test doesn't actually mean much about what was found in the file. Can we add an additional test after this (and line 45) for:

`expect_equal(length(r$endpoints[[1]]), 4)`

to check the contents of this length-1 list?

R/plumber.R
```diff
@@ -493,7 +493,13 @@ plumber <- R6Class(
 
 #' @rdname plumber
 #' @export
-plumb <- function(file){
+plumb <- function(file, dir){
```
trestletech: As you already observed, we'll need docs for dir. https://github.com/trestletech/plumber/blob/master/R/plumber.R#L25

R/plumber.R
```diff
@@ -23,6 +23,7 @@ stopOnLine <- function(private, line, msg){
 #' See \url{http://plumber.trestletech.com/docs/programmatic/} for additional
 #' details on the methods available on this object.
 #' @param file The file to parse as the plumber router definition
+#' @param dir The directory containing the `plumber.R` file to parse as the plumber router definition
```
trestletech: Nit: in RStudio can you run a `Code > Reflow Comment` on this?

R/plumber.R
```diff
@@ -493,7 +494,13 @@ plumber <- R6Class(
 
 #' @rdname plumber
 #' @export
-plumb <- function(file){
+plumb <- function(file, dir){
+  if(!xor(missing(file), missing(dir))){
+    stop("plumber needs only one of a file or a directory with a `plumber.R` file in its root.")
+  } else if (missing(file)){
+    dir <- sub("/$", "", dir)
+    file <- base::file.path(dir,"plumber.R")
```
trestletech: No need for `base::`

aronatkins:
@scottmmjackson Probably should have added yourself to `DESCRIPTION` as an author!

PR #116, opened 2017-06-19
Relicense to Apache 2.

trestletech:


gaborcsardi:
What's wrong with MIT?

trestletech:
Nothing. I can't tell any meaningful difference between the two, but I want to include some assets from swagger-ui which is apache 2. At this point it seemed easier to relicense than to create a separate assets package. Unless I hear any objections...

gaborcsardi:
Not that I know much about licenses, but I think:
* you can mix the two under MIT, you only need to list the modifications to the original Apache-licensed piece (once you modify it). I would not switch licenses just because of this.
* while the two licenses are basically equivalent, imo MIT is better, because it is shorter and has somewhat less restrictions. But this is just an opinion. Others might like Apache better.
* apache seems to be a pain for people who would take (some of) your codebase, because of the "list the modifications" clause.

trestletech:
Was told by legal counsel that it's acceptable to bundle an Apache 2 asset in an MIT project. So closing this out.

Still need to keep all the other assets on a CDN since the minified JS represents copyright claims from dozens (hundreds?) of others that CRAN would want represented in the DESCRIPTION.

PR #122, opened 2017-07-07
entrypoint.R support.

### LINKED ISSUE #106: plumb(dir) should look for entrypoint.R
When plumbing a directory, we should look for another magic file named `entrypoint.R` which would allow you to define custom logic for starting your API.

For instance, if you want to define an encrypted cookie in your API, you could rely on `entrypoint.R`to setup your cookie key and return a `run`nable plumber router.

trestletech:
Add support for the `entrypoint.R` file which can be used when parsing a directory. In that case, rather than naively trying to parse the plumber.R file in the directory, we'll entrust the entrypoint.R to generate a plumber router for us. Presumably a user would do this in order to do some extra configuration on the front-end.

@scottmmjackson do you mind reviewing?

Closes #106

### DIFF
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -386,6 +386,17 @@ plumb <- function(file, dir){
     dir <- sub("/$", "", dir)
     file <- file.path(dir,"plumber.R")
   }
-  plumber$new(file)
+
+  if (file.exists(file.path(dir, "entrypoint.R"))){
+    old <- setwd(dir)
+    on.exit(setwd(old))
+
+    # Expect that entrypoint will provide us with the router
+    x <- source("entrypoint.R")
+    # source returns a list with value and visible elements, we want the (visible) value object.
+    x$value
+  } else {
+    plumber$new(file)
+  }
 }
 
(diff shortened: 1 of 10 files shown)


scottmmjackson:
```
stop("plumber needs only one of a file or a directory with a `plumber.R` file in its root.")
``` <- + "or `entrypoint.R`"

File: R/plumber.R [hunk 1]
```diff
@@ -386,6 +386,17 @@ plumb <- function(file, dir){
...
-  plumber$new(file)
+
+  if (file.exists(file.path(dir, "entrypoint.R"))){
+    old <- setwd(dir)
+    on.exit(setwd(old))
+
+    # Expect that entrypoint will provide us with the router
+    x <- source("entrypoint.R")
+    # source returns a list with value and visible elements, we want the (visible) value object.
+    x$value
```
scottmmjackson: Do we want to trap for malformed `entrypoint.R`?

scottmmjackson:
One last thing: Docstring that Entrypoint.R takes precedence over Plumber.R if they're both present.

R/plumber.R
```diff
@@ -381,11 +383,35 @@ plumber <- R6Class(
...
+
+    old <- setwd(dir)
+    on.exit(setwd(old))
+
+    # Expect that entrypoint will provide us with the router
+    x <- source("entrypoint.R")
+
+    # source returns a list with value and visible elements, we want the (visible) value object.
+    pr <- x$value
+    if (!("plumber" %in% class(pr))){
```
scottmmjackson: Yes! This is awesome.

PR #151, opened 2017-07-31
Set ByteCompile on installation

FvD:
At UseR!2017 @csgillespie emphasized the importance of adding the BytCompile flag to packages. It is a an easy way to increase performance of the package. See [Efficient R Programming](https://csgillespie.github.io/efficientR/7-4-the-byte-compiler.html) for reference.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,4 +1,3 @@
-
 Encoding: UTF-8
 Package: plumber
 Type: Package
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -1,3 +1,4 @@
+
 Encoding: UTF-8
 Package: plumber
 Type: Package
@@ -26,6 +27,7 @@ Imports:
     httpuv (>= 1.2.3),
     crayon
 LazyData: TRUE
+ByteCompile: TRUE
 Suggests:
     testthat (>= 0.11.0),
     XML,


DESCRIPTION
```diff
@@ -1,3 +1,4 @@
+
```
trestletech: Do you mind removing the extra line here?

csgillespie:
Since I was mentioned above, I'll chime in.

A few R versions ago, there was a potential downside (see https://stackoverflow.com/questions/10106736/possible-shortcomings-for-using-jit-with-r)

However, the JIT is now implemented by default in R 3.4. I suspect that byte compiling will be enabled by default in the near future. 

All recommended R packages are byte compiled by default.

Downsides are slightly longer install time and larger binary. I suppose this has implications for CRAN, but not for the user

trestletech:
Thanks for the info. https://stackoverflow.com/questions/37932467/when-is-rs-bytecompile-counter-productive was helpful too.

Yeah, I'm open to trying this out. I just released to CRAN so we can merge this in now and let it soak for a bit. If the debugging or slow-installs are causing us too much pain when developing I might consider backing this back out unless the performance gain for serving is really great, but let's give it a shot.

FvD:
@trestletech I removed the stray newlines!

trestletech:
Thanks!

PR #157, opened 2017-08-08
respect setErrorHandler by setting debug as an option

slopp:
Tricky part was ensuring that a function call like `run(debug=TRUE)` didn't have the side-affect of changing the option if it was otherwise set. This might not matter that much, in which case you could simplify the code.

### DIFF
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -182,7 +182,7 @@ plumber <- R6Class(
 
     },
     run = function(host='127.0.0.1', port=getOption('plumber.port'), swagger=interactive(),
-                   debug= getOption('plumber.debug') %||% interactive() ){
+                   debug=interactive() ){
       port <- findPort(port)
 
       message("Starting server to listen on port ", port)
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -194,7 +194,7 @@ plumber <- R6Class(
       # Set and restore the wd to make it appear that the proc is running local to the file's definition.
       if (!is.null(private$filename)){
         cwd <- getwd()
-        on.exit({ setwd(cwd) })
+        on.exit({ setwd(cwd) }, add = TRUE)
         setwd(dirname(private$filename))
       }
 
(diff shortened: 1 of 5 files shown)


R/plumber.R
```diff
@@ -182,12 +182,14 @@ plumber <- R6Class(
 
     },
     run = function(host='127.0.0.1', port=getOption('plumber.port'), swagger=interactive(),
-                   debug=interactive()){
+                   debug= getOption('plumber.debug') %||% interactive() ){
```
trestletech: I'm unfamiliar with `%||%`. Do you have a reference?

R/utils.R
```diff
@@ -0,0 +1 @@
+`%||%` <- function(l, r) if (is.null(l)) r else l
```
trestletech: Can we add tests for this, I don't think the logic is right. Wouldn't it fail on `FALSE %||% TRUE`?

R/plumber.R
trestletech: Oh, I see it's custom. I feel like I'm going to get tripped up on this every time I come across it and I think there's a bug in the implementation. And since this is part of the public function signature I'd rather not have this internal function leak out.

Can we just ignore the option here and leave it as `interactive()`? Then if someone wants to override it they can do so in the `run()` invocation.

slopp:
@trestletech - addressed your feedback

trestletech:
@slopp tests failing on utils.R in collate field. Can you remove that? Otherwise LGTM.

slopp:
Should be all set, shame on me for not running the check before pushing.

R/plumber.R
```diff
@@ -182,12 +182,14 @@ plumber <- R6Class(
...
     },
     run = function(host='127.0.0.1', port=getOption('plumber.port'), swagger=interactive(),
-                   debug=interactive()){
+                   debug=interactive() ){
       port <- findPort(port)
 
       message("Starting server to listen on port ", port)
 
-      private$errorHandler <- defaultErrorHandler(debug)
+      on.exit({ options('plumber.debug' = getOption('plumber.debug')) })
```
trestletech: Shoot. Sorry. Last thing. I just noticed this is going to get clobbered by the `on.exit()` call on line 197. Can you throw an `add=TRUE` onto 197?

slopp: Good catch, updated

trestletech:
Thanks!

PR #165, opened 2017-09-05
Parsing query string arrays

mhairi:
When a query string has multiple of the same element. For example ‘keyword=hello&keyword=goodbye’, these form an array and should probably be parsed in R as a vector e.g. keyword = c(‘hello’, ‘goodbye’). 

Currently having multiple elements of the same name in a query string just raises an error. The pull request aims to fix this and has added a test for it.


trestletech:
Awesome! Thanks so much for contributing, and sorry for the slow response.

PR #179, opened 2017-10-13
Update plumber.R

vfulco:
minor misspelling


trestletech:
Thanks for contributing!

Issue #212, opened 2017-11-23
Support for negative numbers


etiennebr:
It seems like providing a negative number is not mapped to a `numeric`.

For instance:
``` r
#' @get /type/<x:numeric>
function(x){
  list(
    value = x,
    type = typeof(x)
  )
}
```
Then `curl https://localhost/type/-1.2` returns `{"error":["404 - Resource Not Found"]}`. Would changing the regex for numeric from `\d*\.?\d*` to `-?\d*.\d*` break anything? Also, since the numeric wasn't working I tried to receive `x` as character and parse it within the code, but dots(`.`) are explicitly excluded by `[^\./]+`. Would it be possible to allow dots in characters?

trestletech:
Closed by #213

PR #213, opened 2017-11-23
Support `.` in characters and negative numbers

etiennebr:
This is in relation to #212. I don't know how is your code coverage for tests, but it seems to pass. I was expecting to break more things, especially the `.` in characters since it was explicitly excluded. I'd be happy to change the pull request if needed.

### DIFF
--- a/R/query-string.R
+++ b/R/query-string.R
@@ -70,9 +70,9 @@ createPathRegex <- function(pathDef){
 }
 
 typeToRegex <- function(type){
-  re <- rep("[^\\\\./]+", length(type))
-  re[type == "int"] <- "\\\\d+"
-  re[type == "double" | type == "numeric"] <- "\\\\d*\\\\.?\\\\d*"
+  re <- rep("[^/]+", length(type))
+  re[type == "int"] <- "-?\\\\d+"
+  re[type == "double" | type == "numeric"] <- "-?\\\\d*\\\\.?\\\\d*"
   re[type == "bool" | type == "logical"] <- "[01tfTF]|true|false|TRUE|FALSE"
 
   re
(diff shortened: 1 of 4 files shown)


R/query-string.R
```diff
@@ -70,9 +70,9 @@ createPathRegex <- function(pathDef){
 }
 
 typeToRegex <- function(type){
-  re <- rep("[^\\\\./]+", length(type))
-  re[type == "int"] <- "\\\\d+"
-  re[type == "double" | type == "numeric"] <- "\\\\d*\\\\.?\\\\d*"
+  re <- rep("[^/]+", length(type))
+  re[type == "int"] <- "-?\\\\d+"
+  re[type == "double" | type == "numeric"] <- "-?\\\\d*\\\\.?\\\\d*"
```
trestletech: Unrelated to your code, but I believe this is actually a bug, as it matches `""` as currently written. Do you mind swapping one of the `\\\\d*`s with a `\\\\d+` so that we can be sure that there's at least one digit in the string (as we do with int)?

etiennebr: Ok, you're right. I'll use `-?\d*\.?\d+` which would match e.g. `.23`, which can be interesting, I guess.

trestletech:
Everything looks great except that the `tests/testthat/files/include/test.html` seems to have been accidentally committed. If you don't mind removing that from the branch, I'll merge. Thanks again!

etiennebr:
Sorry about the `test.html`. Seems like everything is ready, but tell me if not. Thanks for the plumber package and thanks for making it so easy to contribute.

trestletech:
Awesome! Thanks again.

PR #215, opened 2017-12-04
cookie expires and max-age arguments format

sambaala:
Hi,
I corrected the cookie format to make its _expires_ and _max-age_ arguments work.
Regards


trestletech:
🤦‍♂️ How'd I miss this?

Thanks for contributing. Change looks great. Closes #216

PR #228, opened 2018-01-17
Add support for PATCH verb

cw25:
Helpful for those of us who use PATCH to differentiate partial resource update operations


trestletech:
Thanks for contributing!

PR #244, opened 2018-03-26
Change to using url_decode for query parameters to handle +

cliveseldon:


schloerke:
@cliveseldon Do you have an example where the current implementation breaks?

PR #248, opened 2018-03-30
Add support for returning promises from endpoints

trestletech:
TODO:

 - [x] support filters returning promises, at which point we promote the rest of the calls into a promise chain?
 - [x] tests
 - [ ] docs
 - [x] NEWS update
 - [x] httpuv version dependency update


kahuna-blake:
Are promises/a-sync processing supported in plumber yet? 

 I somehow need to have a-sync processing on Google App Engine/Flex with a docker container.  I looked into docker-compose and docker swarm, but it does not appear GAE supports those yet.

shrektan:
Any future plans on this? It would be great to have concurrent connections.

schloerke:
@shrektan Yes! This is definitely on our radar.  I am really pushing for this to happen before the next CRAN release.

R/async.R
```diff
@@ -0,0 +1,131 @@
...
+  runStep <- function() {
+    if (nextStepPos > step_count) {
+      return(x)
+    }
+
+    nextStep <- steps[[nextStepPos]]
+    nextStepPos <<- nextStepPos + 1L # TODO pass in as value? multisession issue
+    # if NULL is passed in (or not a function), it is skipped
+    if (!is.function(nextStep)) {
+      return(runStep())
```
jcheng5: I'd do this for NULL, but throw if non NULL and not a function.

R/async.R
```diff
@@ -0,0 +1,131 @@
...
+    function(x) {
+      # always continue
+      FALSE
+    },
+    steps
+  )
+}
+
+#' @import promises
+runStepsUntil <- function(initialValue, errorHandlerStep, conditionFn, steps) {
```
jcheng5: This is lovely.

R/plumber-step.R
```diff
@@ -6,8 +8,11 @@
 #' itself and no other filters or endpoints should be evaluated for this
 #' request.
 #' @export
-forward <- function(){
-  .globals$forwarded <- TRUE
+forward <- function() {
+  structure(forward_class, class = forward_class)
```
jcheng5: This is changing the semantics of forward, isn't it? Previously you just had to call forward() somewhere in your filter but now you have to make sure that forward() is what's returned?

R/plumber.R
```diff
@@ -391,53 +413,117 @@ plumber <- R6Class(
       invisible(self)
     },
 
-    serve = function(req, res){
+    serve = function(req, res) {
       hookEnv <- new.env()
 
-      private$runHooks("preroute", list(data=hookEnv, req=req, res=res))
+      errorHandlerStep <- function(error, ...) {
+        str(list(error, ...))
```
jcheng5: Was this str just for debugging?

schloerke: Correct! Removed.

R/plumber-step.R
schloerke: Correct.  I've noted this in breaking changes within the NEWS.  All examples show it as a final step, so I'm hoping this will be minimal impact.

There probably needs to be more documentation around this.

R/async.R
schloerke: 🕺

R/plumber-step.R
jcheng5: Oh, can't we just make it work?

R/async.R
schloerke: fixed.

R/plumber-step.R
trestletech: I worry that'd be a notable breaking change. Might require a major version bump if we can't get it to work IMHO.

schloerke: fixed!

R/plumber.R
```diff
@@ -749,3 +877,9 @@ plumber <- R6Class(
     }
   )
 )
+
+
+
+bounceErrorStep <- function(error, ...) {
```
schloerke: replace with stop

tests/testthat/test-async.R
```diff
@@ -0,0 +1,341 @@
...
+wait_for_async <- function() {
+  while (!later::loop_empty()) {
+    later::run_now()
+    Sys.sleep(0.00001)
+  }
+}
+
+
+get_result <- function(result) {
+  if (!promises::is.promise(result)) {
```
schloerke: is.promise -> is.promising

PR #258, opened 2018-05-04
RStudio integration work

gtritchie:
## Version bump
Bumped version to 0.4.5

## Callback
Added optional `swaggerCallback` parameter to `run()`.

If supplied, needs to be a function that will be invoked with the URL of the swagger page.

May also be provided by setting the R option `plumber.swagger.url`.

## New Project Template
Implemented simple New Project Template for RStudio. This works in RStudio 1.1.28 or newer.

Shows up in New Project dialog, and creates an RStudio project with a sample `plumber.R` in specified directory.


R/plumber.R
```diff
@@ -196,7 +196,7 @@ plumber <- R6Class(
 
     },
     run = function(host='127.0.0.1', port=getOption('plumber.port'), swagger=interactive(),
-                   debug=interactive() ){
+                   debug=interactive(), swagger.url=getOption('plumber.swagger.url', NULL)){
```
trestletech: Do you mind renaming this to make it more clear that it's a callback? As written, I feel like it implies that you get to specify the URL using this argument.

And I've usually been using camelCase for parameter names.

Maybe this could be `swaggerCallback`?

gtritchie: Sure, that'll work. I'll change it on Monday and update the PR.

R/new-rstudio-project.R
```diff
@@ -0,0 +1,62 @@
...
+    "",
+    "#* Return the sum of two numbers",
+    "#* @param a The first number to add",
+    "#* @param b The second number to add",
+    "#* @post /sum",
+    "function(a, b){",
+    "  as.numeric(a) + as.numeric(b)",
+    "}",
+    ""
+  )
```
trestletech: Is there compelling reason not to put this in a file in that new inst directory and copy it over, or `readlines` it in here? If not, I may have a slight preference for one of those just for maintainability.

PR #259, opened 2018-05-06
Body encoding

farrjere:
This should fix #153.  I added a test to validate that it does handle encoding issues, also I tested locally on my windows partition that it was broken and isn't now. 

I couldn't see any rules for contributing so please let me know if there is something else I need to do.  I did sign the agreement though.

PR #263, opened 2018-06-01
Handle non base64 encoded javascript from pandoc

schloerke:
Javascript is not being encoded by default anymore. When reading the saved widget file, the new lines must be preserved to keep the javascript functioning.

In the tests, checking for base64 encoded images is still valid within the htmlwidgets test to make sure it is self contained.

Reference:
"SelfContained: don't use data URIs for script or style.": jgm/pandoc@7c0a80c


schloerke:
This PR fixes the testing error in #259 and #264

PR #265, opened 2018-06-01
Smaller test include html file and avoid stomping test file

schloerke:
Moved include test.html to test-html.html and update the appropriate include route file.  Then removed all css and javascript code within the html to reduce bloat. 

This prevents the updating of the html file for every tweak done within pandoc or code.  The test is only testing to see if html is returned, not compiled the exact same way.


schloerke:
Merge #263 to pass travis test

Issue #219, opened 2017-12-18
Automated Swagger docs for programmatic routes


nuest:
I've changed my implementation from annotated routes to programatically defined one (see https://github.com/r-lib/debugme/issues/30).

Can I somehow manually provide the (parameters) information for the Swagger UI generation?

trestletech:
Not obviously, no. I'll keep this open to consider how we could support that.

schloerke:
bump

kurt-o-sys:
I guess this still doesn't work? I'd like to use plumber in a package, but that's an issue:
* when using files and annotations for defining the routes, packaging doesn't work properly (can't point to files when packaging, one should always point to objects)
* when defining the routes programmatically, the generated openapi misses the parameters (and probably other stuff)

meztez:
Ok, for everyone looking to this here is an example, we dot it by passing arguments via `...` to function like `pr_get`, `pr_handle`.

Here is the structure of the arguments you can use with `...`.

`params` : a named list of named lists
```R
params[[{parameter name}]] <- list(desc= {description}, type={OpenAPIType}, required={logical}, isArray={logical})
```
See [OpenAPIType](https://www.rplumber.io/dev/articles/annotations.html#more-details-on-param)

`comments` : a character string

`responses` : a named list of named list
```R
responses[[{response label}]] <- list(description={response description})
```
You could probably fill `responses` with a valid openapi spec and it will generate correctly.
`tags` : a character vector

So let see an example

This
```R
#* @get /bob
#* @param a:[int]* parameter a
#* @param b:chr parameter b
#* @tag uno
#* @tag dos
#* @response 201 Created
function(a, b) {
  ans <- list()
  ans[[b]] <- sum(as.numeric(a))
}
```

Can be written programmatically like this
```R
f <- function(a, b) {
  ans <- list()
  ans[[b]] <- sum(as.numeric(a))
}
pr() %>% pr_get("/bob", f, tags = c("uno", "dos"), params = 
                  list(a = list(
                      desc = "parameter a",
                      type = "integer",
                      required = TRUE,
                      isArray = TRUE),
                    b = list(
                      desc = "parameter b",
                      type = "string",
                      required = FALSE,
                      isArray = FALSE))) %>% pr_run()
```

schloerke:
Work around provided above. 

Marking as not-completed as plumber did not change it's code.

PR #274, opened 2018-06-19
update to generate swagger docs for mounted routers

bradleyhd:
Hi,

I'm not sure if plumber is currently taking unsolicited PRs, but I'm new to both R and plumber and really wanted the automatic swagger doc generation to work with mounted routers. These changes seem to work for me (at least for walking this list of endpoints for the router children of the root router). Open to any/all feedback/criticism and happy to revise.

Thanks!

(I _think_ this addresses https://github.com/trestletech/plumber/issues/219)

### DIFF
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -676,19 +676,20 @@ plumber <- R6Class(
 
       private$ends[[preempt]] <- c(private$ends[[preempt]], ep)
     },
-    swaggerFileWalkMountsInternal = function(router, parentPath = ""){
+    swaggerFileWalkMountsInternal = function(router, parentPath=""){
 
       parentPath <- sub("[/]$", "", parentPath)
       endpoints <- lapply(router$endpoints, function(endpoint){
 
         endpointEntries <- lapply(endpoint, function(endpointEntry){
+          endpointEntry <- endpointEntry$clone()
           endpointPath <- sub("^[/]", "", endpointEntry$path)
           endpointPath <- paste(parentPath, endpointPath, sep="/")
           endpointEntry$path <- endpointPath
           endpointEntry
         })
         
-        endpoint
+        endpointEntries
       })
 
       mounts <- router$mounts
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -532,7 +532,30 @@ plumber <- R6Class(
       private$addFilterInternal(filter)
     },
     swaggerFile = function(){ #FIXME: test
-      endpoints <- prepareSwaggerEndpoints(self$endpoints)
+      endpoints <- self$endpoints
+
+      if (length(self$mounts) > 0){
+        for (path in names(self$mounts)){
+
+          mount <- self$mounts[[path]]
+          mount_endpoints <- mount$endpoints
+          path <- sub("[/]$", "", path)
+
+          if (length(mount_endpoints) > 0) {
+            mount_endpoints <- lapply(mount_endpoints, function(i){
+              i <- lapply(i, function(j){
+                subpath <- sub("^[/]", "", j$path)
+                j$path <- paste(path, subpath, sep="/")
+                j
+              })
+              i
+            })
+            
+            endpoints <- c(endpoints, mount_endpoints)
+          }
+        }
+      }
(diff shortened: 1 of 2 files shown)


schloerke:
Hi @bradleyhd !

Thanks for the PR!  Good catch!  (Unsolicited PRs are welcome anytime!)

Couple things to address.  
1. Could the for loop be changed to a 
```r
mountedEndpoints <- unlist(lapply(names(self$mounts), function(path) {
...
  if (length(mountEndpoints) == 0) return(NULL)
...
}))
endpoints <- c(endpoints, mountedEndpoints)
```
`NULL` values  with `unlist()` are removed.  When possible, avoid `answer <- c(answer, newItem)` as it does not scale well.

2. Formatting
* The inner logic seems good for calculating `mount_endpoints`.  Could you upgrade the variable names (`i`,`j`) to something more descriptive? (`mountEndpoint`, `mountEndpointEntry` or something similar)
* Could you change the code to use `camelCase` (vs `snake_case`)?

3. We should make the swaggerFile mount calculations recursive.  In the current state, it is missing swaggerFile information for `p3` (`/sub2/sub3`)
```r
pr <- plumber$new()
pr$handle("GET", "/nested/path/here", function(){})
pr$handle("POST", "/nested/path/here", function(){})

stat <- PlumberStatic$new(".")

pr2 <- plumber$new()
pr2$handle("POST", "/something", function(){})
pr2$handle("GET", "/", function(){})

pr3 <- plumber$new()
pr3$handle("POST", "/else", function(){})
pr3$handle("GET", "/", function(){})

pr$mount("/static", stat)
pr2$mount("/sub3", pr3)
pr$mount("/sub2", pr2)

pr$swaggerFile()
```

Thank you in advance,
Barret

schloerke:
Yes, it solves #219

bradleyhd:
Thanks for the pointers/feedback! I've updated to walk routers/mounts recursively and it seems to work on your example and a few others I tried out. However, if you repeatedly call `pr$swaggerFile()`, the paths are repeatedly prepended with their parent paths. I'm not sure what's idiomatic here for preventing this from happening but something should probably be done 😄

bradleyhd:
@schloerke maybe I'm not looking in the right place, but I don't see any review changes. Mind re-adding them?

schloerke:
My fault.. never hit "submit review".  (GitHub review is new to me)

bradleyhd:
No worries at all. I've pushed the test and clone changes, let me know if there's anything else!

R/plumber.R
```diff
@@ -673,6 +675,31 @@ plumber <- R6Class(
...
+      })
+
+      mounts <- router$mounts
+      mountedEndpoints <- unlist(lapply(names(mounts), function(mountPath){
+        mountedSubrouter <- router$mounts[[mountPath]]
+        mountPath <- sub("^[/]", "", mountPath)
+        mountedEndpoints <- private$swaggerFileWalkMountsInternal(mountedSubrouter, paste(parentPath, mountPath, sep="/"))
+      }))
+
+      c(endpoints, list(mountedEndpoints))
```
alandipert: Is there a reason you call `list` here? My understanding is that at this point `endpoints` is a list and `mountedEndpoints` is a character vector, so the result will definitely be a list.

bradleyhd: @alandipert I'm a bit fuzzy on the types but I think endpoints is a list of lists and mountedEndpoints without the extra `list` call is just a list. If you remove that call and try to run swagger documentation, you'll get `Error in for (e in fil) { : invalid for() loop sequence`. When I print out the result of calling `c` above without modification, I get entries like
```
[[1]]`$__no-preempt__14`
<PlumberEndpoint>
...
```

alandipert: Ah, that makes sense, thanks for clarifying.

schloerke:
Merged in #280 

Thank you @bradleyhd!!

turiaso:
Hi, a quick question. Do you upload the version with this feature 0.4.6.9000 (based on versión downloaded by install_github) to CRAN repositories. Same case on 0.5.0 version, based on changelogs. I understand that both versions are more similars, one is a snapshot another is a release.
Regards!

PR #280, opened 2018-07-16
Swagger support for recursive router mounts

schloerke:
Continuation of #274 

Changed behavior of internal method to capture the endpoint entries and not the nested entries within the endpoint.  The names of the nested structure have meaning and should be done properly with a `flatten(pr)` method.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -64,4 +64,4 @@ Collate:
     'serializer.R'
     'session-cookie.R'
     'swagger.R'
-RoxygenNote: 6.0.1.9000
+RoxygenNote: 6.0.1
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -64,4 +64,4 @@ Collate:
     'serializer.R'
     'session-cookie.R'
     'swagger.R'
-RoxygenNote: 6.0.1
+RoxygenNote: 6.0.1.9000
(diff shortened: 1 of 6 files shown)


DESCRIPTION
```diff
@@ -64,4 +64,4 @@ Collate:
     'serializer.R'
     'session-cookie.R'
     'swagger.R'
-RoxygenNote: 6.0.1
+RoxygenNote: 6.0.1.9000
```
alandipert: Is the `.9000` bit OK to check in?

R/swagger.R
```diff
@@ -17,34 +17,32 @@ plumberToSwaggerType <- function(type){
...
 
-  for (fil in routerEndpoints){
-    for (e in fil){
-      # TODO: we are sensitive to trailing slashes. Should we be?
-      cleanedPath <- gsub("<([^:>]+)(:[^>]+)?>", "{\\1}", e$path)
-      if (is.null(endpoints[[cleanedPath]])){
-        endpoints[[cleanedPath]] <- list()
-      }
+  for (e in routerEndpointEntries){
+    # TODO: we are sensitive to trailing slashes. Should we be?
```
alandipert: I think we should be, but I can imagine it being an option.

DESCRIPTION
schloerke: i'll take it off. typically is though

R/swagger.R
schloerke: that is an old comment from https://github.com/trestletech/plumber/pull/280/files#diff-32592aa9d638876dfcfa9d16765e4311L25 

I think we should be as well

PR #292, opened 2018-08-13
We can now install plumber from CRAN

mtoto:
By installing plumber from CRAN in the Dockerfile we can make the image slightly smaller.


schloerke:
Thanks!

PR #297, opened 2018-08-20
Multiline POST body collapsed

robertdj:
Re: https://github.com/trestletech/plumber/issues/270

### DIFF
--- a/R/post-body.R
+++ b/R/post-body.R
@@ -1,7 +1,7 @@
 postBodyFilter <- function(req){
   handled <- req$.internal$postBodyHandled
   if (is.null(handled) || handled != TRUE){
-    body <- req$rook.input$read_lines()
+    body <- paste0(req$rook.input$read_lines(), collapse = "\n")
     charset <- getCharacterSet(req$HTTP_CONTENT_TYPE)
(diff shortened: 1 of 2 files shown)


boglesby03:
Not everything posted is single line.  We're using plumber to post csv's so we need the returns in the body of the message.

R/post-body.R
```diff
@@ -1,7 +1,7 @@
 postBodyFilter <- function(req){
   handled <- req$.internal$postBodyHandled
   if (is.null(handled) || handled != TRUE){
-    body <- req$rook.input$read_lines()
+    body <- paste0(req$rook.input$read_lines(), collapse = "\n")
```
boglesby03: What if post body has valid returns in it?  Such as a csv, or a text field that has intentional returns in it?  Blanket removing returns isn't the correct solution.

schloerke: `read_lines()` should only be returning a character vector for each line of text.  If the text has a newline value in it, it should be escaped and be interpreted as regular text.  

By collapsing the text with `\n`, the original new line locations are preserved.  This makes the postBody a single character value (correct), not a vector of character values (incorrect).

Issue #296, opened 2018-08-17
Chinese characters are not supported


shangfr:
Is there any way to set the character set so that the plumber can support Chinese characters?
谢谢！

schloerke:
@shangfr Do you have a small example plumber file?  This way I can reproduce / fix the error you're seeing.

Thank you in advance!

shangfr:
> @shangfr Do you have a small example plumber file? This way I can reproduce / fix the error you're seeing.
> 
> Thank you in advance!

shrektan:
@shangfr It should be fixed via the PR #312 and #314. You can try after the two PRs being merged. 😄

shangfr:
Thank you. I appreciate your help. Thanks.

PR #312, opened 2018-10-05
plumber() and Plumber$new() will read source file as UTF-8 encoded

shrektan:
Related to #296

### DESCRIPTION

R uses the encoding "unknown" or "native" when reading text files by default. However, the fact that the default encoding varies (not UTF-8) on Windows, leads to a  big headache for our multiple-bytes charset users on Windows, which is **we have to store the source file as the native encoding** in order to use plumber. 

Thanks to the great work of RStudio (especially, @yihui), the common practice of handling this issue is to **read the file as UTF-8 encoding by default**(may provide a param "encoding" allowing users to change) and this PR is going to address this.

### WHAT THIS PR DOES?

- add the new param `encoding` for `plumber()` and `Plumber$new()`
- enable the UTF-8 as the default

As mentioned in #296, there's another issue that the function input should be marked as UTF-8, which  I would like file another PR after this being merged.

Thanks.


yihui:
YES! Use UTF-8, or we can't be friends.

Perhaps in **knitr** 2.0 and **rmarkdown** 2.0, I'll just remove the `encoding` arguments, and force UTF-8 like what Pandoc has been doing for years, which is something I have envied for long.

schloerke:
@yihui Do you think it's a good idea remove the argument and just force UTF-8?

shrektan:
It’s a good idea to me actually.

yihui:
@schloerke Yes. That is what I have been doing in bookdown (since 2016) and blogdown (2017). In future packages, I'll always do this. The `encoding` argument brings enormous pain to developers, whereas it only bugs users a tiny bit when they are asked to save a text file in UTF-8 (they can set this as the default in RStudio).

schloerke:
@shrektan Mind removing it from the args and adding it as a global package variable w/ comment to this PR? (Feel free to copy from below)

```r
# Hard code UTF-8 file encoding
# Removes encoding headache at minor cost of setting encoding in editor
# https://github.com/trestletech/plumber/pull/312
utf8Encoding <- "UTF-8"


plumb <- function(file, dir = "."){
# ...
x <- source(entrypoint, encoding = utf8Encoding)
# etc...
}
```

shrektan:
@schloerke I have removed it.

R/plumber.R
```diff
@@ -177,10 +182,17 @@ plumber <- R6Class(
       private$notFoundHandler <- default404Handler
 
       if (!is.null(file)){
-        private$lines <- readLines(file)
-        private$parsed <- parse(file, keep.source=TRUE)
+        con <- file(file, encoding=utf8Encoding)
+        on.exit(close(con), add=TRUE)
+
+        private$lines <- readLines(con)
+        Encoding(private$lines) <- utf8Encoding
```
schloerke: I believe this line is not needed as the `con` already has the encoding put in.

schloerke:
```
> b <- url("https://raw.githubusercontent.com/trestletech/plumber/b8cbe3afd0b4a5d8f012f4e06e5248e45535e785/tests/testthat/files/multibytes.R", encoding = "UTF-8")
> readLines(b)
[1] "#* @get /echo"  "function() {"   "  \"中文消息\"" "}"     
```

R/plumber.R
```diff
@@ -177,10 +182,17 @@ plumber <- R6Class(
...
+
+        srcfile <- srcfilecopy(file, private$lines, file.info(file)[1, "mtime"],
```
schloerke: Any specific reasoning for the `mtime` over `Sys.time()` (when the file is 'copied')?

R/plumber.R
```diff
@@ -177,10 +182,17 @@ plumber <- R6Class(
...
+                               isFile=TRUE)
+        private$parsed <- parse(text=private$lines, n =-1, srcfile=srcfile)
```
schloerke: `n` is not need here. `NULL` (default) behaves the same as a negative number

schloerke: To be safe, also set `keep.source = TRUE`

R/plumber.R
shrektan: You are right about this.

R/plumber.R
shrektan: I don't think we need this. Changed.

shrektan:
@schloerke I've made the changes you requested and add a NEWS file (I made a forced-push in order to have a clean history). 

Can it be merged? It's very inconvenient to store the file as encoding other than UTF-8. Moreover, it's consistent with shiny, which forces the UTF-8 encoding [in the very early stage](https://github.com/rstudio/shiny/blob/master/NEWS.md#shiny-0101).

> # shiny 0.10.1
> Added Unicode support for Windows. Shiny apps running on Windows must use the UTF-8 encoding for ui.R and server.R (also the optional global.R) if they contain non-ASCII characters. See this article for details and examples: http://shiny.rstudio.com/gallery/unicode-characters.html (#516)

In addition, there's a minor issue about using `utf8Encoding` to represent the literal string "UTF-8" (see the quote below). Although I made the change, I don't support it. As you can see in other packages like [shiny](https://github.com/rstudio/shiny/search?l=R&q=UTF-8) or in [pkgload](https://github.com/r-lib/pkgload/search?l=R&q=UTF-8), they almost always are written directly.

> @shrektan Mind removing it from the args and adding it as a global package variable w/ comment to this PR? (Feel free to copy from below)
> 
> ```r
> # Hard code UTF-8 file encoding
> # Removes encoding headache at minor cost of setting encoding in editor
> # https://github.com/trestletech/plumber/pull/312
> utf8Encoding <- "UTF-8"
> 
> 
> plumb <- function(file, dir = "."){
> # ...
> x <- source(entrypoint, encoding = utf8Encoding)
> # etc...
> }
> ```

PR #313, opened 2018-10-05
use inherits() and expect_is() rather than "xxx" %in% class(obj)

shrektan:
I think useing inherits() and expect_is() should be a better practice.


schloerke:
Thank you for the PR @shrektan !

Could you change the `expect_is` to `expect_s3_class`?

shrektan:
I've made the change. Just notice that: 😺 

> expect_is() is an older form. I'd recommend using expect_s3_class() or expect_s4_class() in order to more clearly convey intent.

PR #314, opened 2018-10-05
the query string supports multibyte chars

### LINKED ISSUE #296: Chinese characters are not supported

### LINKED ISSUE #324: plumber UI process chinese string
Sorry my poor English ><
The Chinese string pain me two days long.

Then we find the source code ,referenced by 
https://github.com/trestletech/plumber/blob/master/R/query-string.R

In that query-string.R,
focus on line 14 to line 47 .
I found that plumber must call function (utils::URLdecode) to process querystring.

But we found that ,for example,
`%E8%81%B7%E5%A0%B4`    ( in Chinese `職場` )
it might output `職場`,
but we got `<e8> <e5>`

Therefore we do that,
```
trace(URLdecode,edit = T)
```
And we just do nothing and then return it as following,
```
function (URL) 
{
  URL
}
```
it means we can get the original URL like `%E8%81%B7%E5%A0%B4`

and we process the URL string by using function `httr::GET()`,

Eventually, we can solve Chinese string .

Full code reference by https://www.rplumber.io/docs/
```
# plumber.R

#' Echo the parameter that was sent in
#' @param msg The message to echo back.
#' @get /echo
#' @json
function(msg=''){
  require(dplyr)
  rawdata <- httr::POST('http://www.convertstring.com/zh_TW/EncodeDecode/UrlDecode',body = list('input'= msg,
                                                                                                'outputtype'= 'outputstring')) %>% httr::content()
  
  decodedstring <- rawdata %>% rvest::html_nodes('#output') %>% rvest::html_text()
  list(msg = paste0("The message is: '", decodedstring , "'"))
```
------------------------------------------------------------------------------------
```
# runplumber.R

trace(URLdecode,edit = T) # function(URL){URL} then save
plumber::plumb('plumber.R')$run()

```

shrektan:
Fixes #296
Fixes #324

`parseQS()` uses `base:: URLdecode()` which will not mark the results as `UTF-8` and causes trouble for Windows users. 

Thanks.


shrektan:
Note, it's kind of simliar to my previous [PR](https://github.com/rstudio/htmltools/pull/95) on rstudio/htmltools .

https://github.com/rstudio/htmltools/pull/95/files#diff-07c5eb6cd2976a58b1a5e1ebc2e4f90cR49

tests/testthat/test-query-with-multibytes.R
```diff
@@ -0,0 +1,14 @@
+context("test-query-with-multibytes")
+
+test_that("Support multi-bytes queries", {
+  r <- plumber$new("files/query-with-multibytes.R")
+  res <- PlumberResponse$new()
+
+  req <- make_req("GET", "/msg", "\u53c2\u65701=\u4e2d\u6587&\u53c2\u65702=\u4f60\u597d")
```
wch: The query string cannot have raw UTF-8 characters in it. They must be URL-encoded. Also, this should have a leading `?`. So the query string should be:

```
"?%E5%8F%82%E6%95%B01=%E4%B8%AD%E6%96%87&%E5%8F%82%E6%95%B02=%E4%BD%A0%E5%A5%BD"
```

Same below.

tests/testthat/test-query-with-multibytes.R
```diff
@@ -0,0 +1,14 @@
...
+  out <- r$serve(req, res)$body
+  expect_identical(out, jsonlite::toJSON("\u4e2d\u6587-\u4f60\u597d"))
```
wch: `identical` automatically does encoding conversion, so this should compare the output of `charToRaw()`. Either that, or check that both strings are UTF-8 encoded and then do the identical check.

R/query-string.R
```diff
@@ -22,13 +22,20 @@ parseQS <- function(qs){
   parts <- strsplit(qs, "&", fixed = TRUE)[[1]]
   kv <- strsplit(parts, "=", fixed = TRUE)
   kv <- kv[sapply(kv, length) == 2] # Ignore incompletes
+  if (length(kv) == 0L) return(list())
 
   keys <- sapply(kv, "[[", 1)
   keys <- unname(sapply(keys, URLdecode))
+  Encoding(keys) <- "UTF-8"
+  # utils::URLdecode() will not mark the encoding
+  # Usually, the original query is UTF-8 encoded.
```
wch: Slight clarification of the comment:

> The query string (after URL decoding) is usually UTF-8, but this is not always true. For simplicity, we'll just enforce that it has to be UTF-8, and not support other encodings. We also need to ensure that R understands that it's UTF-8.

tests/testthat/test-querystring.R
```diff
@@ -23,4 +23,18 @@ test_that("incomplete query strings are ignored", {
 
 test_that("query strings with duplicates are made into vectors", {
   expect_equal(parseQS("a=1&a=2&a=3&a=4"), list(a=c("1", "2", "3", "4")))
-})
\ No newline at end of file
+})
+
+test_that("parseQS() will mark UTF-8 explicitly", {
+  out <- parseQS("\u53c2\u65701=\u4e2d\u6587")
```
wch: `parseQS` expects a URL-encoded string (not raw UTF-8).

tests/testthat/test-query-with-multibytes.R
wch: For reference:

> Octets must be encoded if they have no corresponding graphic character within the US-ASCII coded character set

From https://tools.ietf.org/html/rfc1738

wch:
@schloerke and I looked into this, and we think that it's not even possible to call a function with argument names that can't be represented in the R session's native encoding. 

For example:

```R
f <- function(`参`) {
  `参`
}

f(123)
```

On my Mac (where the native encoding is UTF-8), this prints out:

```
[1] 123
```

On my Windows machine (US locale), it gives an error when I try to define the function:

```
> f <- function(`参`) {
Error: \uxxxx sequences not supported inside backticks (line 1)
```

Now, it is possible to put the function in a file and then call `source()` or our `sourceUTF8()` function on it, and then it will end up with a working function, but with messed up names.

```
> source('test.R')
> f
function(`å‚`) {
  `å‚`
}
```

If you try to call `f` with named arguments, it doesn't work. Here are a bunch of different ways we tried:

```
> f(`参` = 1)
Error: \uxxxx sequences not supported inside backticks (line 1)
> f("参" = 1)
Error in f(`<U+53C2>` = 1) : unused argument (`<U+53C2>` = 1)
> f(参 = 1)
Error: unexpected input in "f(\"
> 
> args <- list("参" = 1)
> names(args) <- "参"      # Needed to make the name encoding UTF-8
> Encoding(names(args))
[1] "UTF-8"
> do.call(f, args)
Error in (function (`å\217‚`)  : unused argument (`<U+53C2>` = 1)
```

Plumber uses `do.call()` to invoke the function, and it doesn't seem to work.

Note: It is possible to run the function by putting this in a file and then sourcing it, but so far that's the only way we've found.

```
f(`参` = 1)
```

shrektan:
@wch I don't think people are going to use multi-bytes names as parameters... and in my option, they shouldn't do that, either...

So I will simply change the test to the one that only use ASCII names as the parameters.

tests/testthat/test-query-with-multibytes.R
shrektan: Thanks, I will make the change.

R/query-string.R
```diff
@@ -22,13 +22,20 @@ parseQS <- function(qs){
   parts <- strsplit(qs, "&", fixed = TRUE)[[1]]
   kv <- strsplit(parts, "=", fixed = TRUE)
   kv <- kv[sapply(kv, length) == 2] # Ignore incompletes
+  if (length(kv) == 0L) return(list())
 
   keys <- sapply(kv, "[[", 1)
   keys <- unname(sapply(keys, URLdecode))
+  Encoding(keys) <- "UTF-8"
```
wch: @schloerke Perhaps this should give a warning instead if `keys` is non-ASCII. I believe that the encoding will be `unknown` iff the text contains only ASCII characters, but I'm not 100% sure.

shrektan: Yes it’s unknown for only ASCII strings but set it to UTF8 will cause no damage - it will stay in unknown.

wch: From my previous comment:

> we think that it's not even possible to call a function with argument names that can't be represented in the R session's native encoding

So if you have a function with a Chinese argument name and your machine is a Windows machine with a Chinese locale, you will not encounter this problem. You could run the same code on a Linux machine, and you won't have the problem. However, if you run the exact same code on a Windows machine with a non-Chinese locale, then you will end up with code that doesn't work, and error message that are very confusing.

My idea here is to warn people when they try to use non-ASCII characters so that they don't end up with that situation.

wch:
Might be relevant: rstudio/httpuv#185.

PR #315, opened 2018-10-08
Accept NA handling arg for serializers

### LINKED ISSUE #279: Fields missing from json when the field is sometimes NA
I'm trying to output a `data.frame` as json.  This works fine for the columns that are always populated.  However, it seems that if the column is `NA` for any row, then the entire column is missing from the resulting json. Is there a fix or configuration option I'm missing?

namelessjon:
Happy to receive any comments or feedback

This allows the use of `#' json(na = ...)` in a comment block, which is handled appropriately.  It requires the use of `purrr::partial` to apply the arguments, as otherwise the fact plumber uses R6 classes breaks relying on using simple closures.

This fixes #279 

I think this code could be generalised to allow specification of any `toJSON` arguments.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -24,7 +24,8 @@ Imports:
     stringi (>= 0.3.0),
     jsonlite (>= 0.9.16),
     httpuv (>= 1.2.3),
-    crayon
+    crayon,
+    purrr (>= 0.2.0)
 LazyData: TRUE
 ByteCompile: TRUE
 Suggests:
--- a/R/parse-block.R
+++ b/R/parse-block.R
@@ -115,7 +115,7 @@ parseBlock <- function(lineNum, file){
       }
     }
 
-    shortSerMat <- stringi::stri_match(line, regex="^#['\\*]\\s*@(json|html)")
+    shortSerMat <- stringi::stri_match(line, regex="^#['\\*]\\s*@(json|html)([\\s\\(].*)?\\s*$")
     if (!is.na(shortSerMat[1,2])){
       s <- stri_trim_both(shortSerMat[1,2])
       if (!is.null(serializer)){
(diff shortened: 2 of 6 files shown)


schloerke:
Hi @namelessjon

Thank you for tackling this problem!

I believe we will not keep up if we make new serializers for different argument settings.  However, I really like the idea of passing the arguments directly for the json serializer (similar to the jpeg, and png serializers).  

Instead of just the `na` argument, is it possible to add `...` as well? This would allow for all arguments to be passed through.

```r
# untested
serializer_json <- function(..., na = "null") {
  purrr::partial(
    ..., na = na,
    .lazy = FALSE,
    ...f = function(val, req, res, errorHandler, ...) {
      tryCatch({
        json <- jsonlite::toJSON(val, ...)
## rest of function
```

With `serializer_unboxed_json` accepting arguments and trying to keep code short, it could be upgraded to:

```r
# untested
serializer_unboxed_json <- function(..., auto_unbox = TRUE) {
  serializer_json(..., auto_unbox = auto_unbox)
}
```


--------------------------------------


Assuming we can get the above to work, the regex for html and json should be split, as html does not allow for any arguments at all.  Given that the rest of the code in the `parseBlock` function is copypasta'ed many times, I'd just duplicate the code block and trim out the necessary parts for each code chunk. (ex: knowing it's only `@html`, could go back to `serializer <- .globals$serializers[[s]]()`)

Also because this feature is so useful and `@serializer unboxedJSON` is so common, I'd like to add in a `@unboxedJSON` to work like `@json` , but calls the unboxedJSON serializer.  

--------------------------------

Thanks again for solving a VERY helpful TODO!

\-Barret

wch:
The latest version of R6 (2.3.0, released last week) made a change to how members that are functions but *not* methods are treated. Previously, when cloned, the new R6 object would turn these functions into methods, but that no longer happens. That fix might eliminate the need for `purrr::partial`.

See https://github.com/r-lib/R6/pull/156 for more information.

namelessjon:
Hi @schloerke 

Thanks for the feedback.  I completely agree on not keeping up with the arguments, and did wonder about adding support for `...`.  I don't *think* that would break anything, and I'm willing to look at doing that.  I can also try and refactor the parse block code to avoid too much copy/paste.

@wch I will look into that.

shrektan:
It will be great to add support for `...` because I often need to set the `digits` param in `jsonlite::toJSON()` (the default value is only 4 digits). At present, I have to register a new serializer by calling `addSerializer()` every time, which is not very elegant.

namelessjon:
Hi, just to let you know I've not forgotten this.  I've just also been super busy.  I hope to get back to it over the next few days

schloerke:
@namelessjon I'm going to purse the remaining parts of this PR. Thank you for the idea and kick start on the code!

PR #318, opened 2018-10-09
Removes the httpuv connection: close workaround

atheriel:
This removes the workaround introduced in 4901c5e to avoid rstudio/httpuv#49, which has been fixed (in my estimation) since at least `httpuv` 1.4.0.

I've tentatively confirmed that environments are no longer shared between connections even when they are `keep-alive` by using the original integration test added in the referenced commit.

As a nice bonus, this improves `plumber`'s baseline latency by about 10% on my machine.

### DIFF
--- a/DESCRIPTION
+++ b/DESCRIPTION
@@ -23,7 +23,7 @@ Imports:
     R6 (>= 2.0.0),
     stringi (>= 0.3.0),
     jsonlite (>= 0.9.16),
-    httpuv (>= 1.2.3),
+    httpuv (>= 1.4.0),
     crayon
 LazyData: TRUE
 ByteCompile: TRUE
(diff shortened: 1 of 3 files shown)


wch:
Looks good to me.

trestletech:
There's one more file in the commit you referenced: https://github.com/trestletech/plumber/commit/4901c5e0f613b37476792996e563226a56a90bd2#diff-68678c8c847a2f72255e66a25514f0abR34 should we also update that one?

schloerke:
@atheriel Thank you for the change!   Could you please sign the CLA so that I may merge this PR in a couple days? Link: https://cla-assistant.io/trestletech/plumber?pullRequest=318

PR #319, opened 2018-10-12
need to set the LC_TIME to C to ensure the Date is formatted in English

shrektan:
Hi, the `PlumberResponse` will attach `Date` to the header. However, the output of the below line will be different on a different computer (see the example below), which is problematic because normally the non-English date will not be parsed correctly (for example, [httr will parse the date as NA](https://github.com/r-lib/httr/pull/544))...

With this PR, we can ensure the date is always in English.

https://github.com/trestletech/plumber/blob/8c25b8bfba9774bd422022dc29662ffc17e170bc/R/response.R#L19

### The example

``` r
date_fmt <- function() {
  cat("Previous: ", format(Sys.time(), "%a, %d %b %Y %X %Z", tz="GMT"))
  cat("\n")
  old_lc_time <- Sys.getlocale("LC_TIME")
  Sys.setlocale("LC_TIME", "C")
  on.exit(Sys.setlocale("LC_TIME", old_lc_time), add = TRUE)
  cat("Now: ", format(Sys.time(), "%a, %d %b %Y %X %Z", tz="GMT"))
}
date_fmt()
#> Previous:  周五, 12 十月 2018 7:04:22 GMT
#> Now:  Fri, 12 Oct 2018 07:04:22 GMT
```

<sup>Created on 2018-10-12 by the [reprex package](https://reprex.tidyverse.org) (v0.2.1)</sup>

### DIFF
--- a/R/response.R
+++ b/R/response.R
@@ -17,10 +17,13 @@ PlumberResponse <- R6Class(
       h <- self$headers
       # httpuv doesn't like empty headers lists, and this is a useful field anyway...
       # need to set the LC_TIME to C to ensure the Date is formatted in English
-      old_lc_time <- Sys.getlocale("LC_TIME")
-      Sys.setlocale("LC_TIME", "C")
-      on.exit(Sys.setlocale("LC_TIME", old_lc_time), add = TRUE)
-      h$Date <- format(Sys.time(), "%a, %d %b %Y %X %Z", tz="GMT")
+      english_time <- function() {
+        old_lc_time <- Sys.getlocale("LC_TIME")
+        Sys.setlocale("LC_TIME", "C")
+        on.exit(Sys.setlocale("LC_TIME", old_lc_time), add = TRUE)
+        format(Sys.time(), "%a, %d %b %Y %X %Z", tz="GMT")
+      }
+      h$Date <- english_time()
 
       # Due to https://github.com/rstudio/httpuv/issues/49, we need each
       # request to be on a separate TCP stream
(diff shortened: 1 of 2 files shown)


wch:
I think it would be better to implement this without setting the locale. I think that can potentially cause problems if there are multiple threads.

Here's an implementation of the same thing I wrote in C++:

https://github.com/rstudio/httpuv/blob/a830a0d/src/utils.h#L197-L247

wch:
Here's an R function that creates a HTTP date string from a POSIXct object. (I wrote this in the process of creating tests for httpuv).

```R
# Given a POSIXct object, return a date string in the format required for a
# HTTP Date header. For example: "Wed, 21 Oct 2015 07:28:00 GMT"
http_date_string <- function(time) {
  weekday_names <- c("Sun", "Mon", "Tue", "Wed", "Thu", "Fri", "Sat")
  weekday_num <- as.integer(strftime(time, format = "%w", tz = "GMT"))
  weekday_name <- weekday_names[weekday_num + 1]

  month_names <- c("Jan", "Feb", "Mar", "Apr", "May", "Jun",
                   "Jul", "Aug", "Sep", "Oct", "Nov", "Dec")
  month_num   <- as.integer(strftime(time, format = "%m", tz = "GMT"))
  month_name <- month_names[month_num]

  strftime(time,
    paste0(weekday_name, ", %d ", month_name, " %Y %H:%M:%S GMT"),
    tz = "GMT"
  )
}


http_date_string(Sys.time())
#> [1] "Wed, 12 Dec 2018 05:54:57 GMT"
```

shrektan:
@wch  I've changed the PR based on your code. Thanks.

PR #325, opened 2018-10-29
Protect from unsafe JSON parsing behavior.

trestletech:
I believe the only way that this could have been exploited would be if the API used encrypted cookies and an attacker knew the encryption key.

We were already sniffing for JSON by reading the first char and matching against '{', so we wouldn't have been vulnerable to any attack in standard usage.


R/json.R
```diff
@@ -0,0 +1,9 @@
+#' @importFrom jsonlite validate
+#' @importFrom jsonlite fromJSON
```
schloerke: with `jsonlite` already as a imports, I believe these two lines are not necessary as we are only calling the pkg functions directly with `::`.

Do you mind dropping these two lines and re-documenting?

trestletech: Hm. I thought we needed to either import all of `jsonlite` in NAMESPACE or the specific functions we wanted. But it looks like R CMD check does indeed work with those lines deleted, so I must be wrong. I'll push the commit and we'll see how CI does.

trestletech:
Oh, look at you! One step ahead of me :) Thanks

r2evans:
I know it's late in the game (congrats on 1.0.0!), is there an advantage to using this over `jsonlite::parse_json()`? It is safer than `fromJSON` and does mostly the same think (lacking `jsonlite::validate`). (Note, it defaults to not simplifying ... other than that, they are identical to `fromJSON`.) (Companion function: `jsonlite::read_json` for explicit file paths.

schloerke:
@r2evans https://github.com/rstudio/plumber/blob/47100cdeb268bf06b4e12ca317a58b426fa65f84/R/json.R#L4 now uses `parse_json` and defaults simply to `FALSE`. I believe this should have the same behavior as before. 

Never too late to bring up discussion and to double check things!  Thank you! 😊😊

r2evans:
Thanks @schloerke. I have been bitten (local-code only, not within plumber) by the flexibility of `fromJSON` before, so I happened to see those functions. I've been shying away from polymorphism that includes filenames, as it can definitely be more problematic, especially cases like these.

But mostly ... I should have realized that even though I read the new-release `NEWS` *today*, that doesn't mean that this PR was at all *recent* (vice almost 2y ago). Oops, I see your `safeFromJSON` now, sorry for the noise. :-)

PR #328, opened 2018-11-01
uft8 source file support

schloerke:
Fixes: #312 

This stems from wanting to read utf8 files on a non utf8 locale.  On windows, the locale is set to the system default which typically does not support utf8 out of the box.

`utf8.R` functions are adopted from the [RStudio Shiny repo](https://github.com/rstudio/shiny/blob/6ede0194c6e52d18be3ec13514defd8e0f22a3c1/R/utils.R#L1459-L1532) in an effort to use code that has already be validated to work.

The end result is that `source(file, encoding="UTF-8")` does not work as expected and instead we should use `eval(parse(file, encoding = "UTF-8"))`

Other additions
* Use `testthat::test_path` when loading test files
* Add Barret as `ctb`
* Removes appveyor.yml build 
* Only source all expressions in a plumber file once - https://github.com/trestletech/plumber/pull/328/commits/cde0d3d2543a654fd0c5799b670767ccb0e22e35


R/plumber.R
```diff
@@ -177,10 +177,9 @@ plumber <- R6Class(
       private$notFoundHandler <- default404Handler
 
       if (!is.null(file)){
-        private$lines <- readLines(file)
-        private$parsed <- parse(file, keep.source=TRUE)
-
-        source(file, local=private$envir, echo=FALSE, keep.source=TRUE)
+        private$lines <- readUTF8(file)
+        private$parsed <- parseUTF8(file)
+        sourceUTF8(file, private$envir)
```
wch: I think we should get an understanding of why we're both parsing the file and sourcing it. That will help us understand whether we need to preserve the `keep.source=TRUE` behavior.

schloerke: I believe we were misled by the name of `keep.source` within `source`.  

`source` docs:
> echo: logical; if ‘TRUE’, each expression is printed after parsing, before evaluation.
> keep.source: logical: should the source formatting be retained when echoing expressions, if possible?

I believe it was used for debugging purposes when developing. (Set `echo` to true and display nicely on screen.)

--------------------

We are more concerned with `parse`'s `keep.source`.

`parse` docs:
> keep.source: a logical value; if ‘TRUE’, keep source reference information.

Which I don't know why it wouldn't be TRUE.  If errors were a big deal, users would be sanitizing them before responding.

schloerke: The extra sourcing is doe to populate local variables in the envir for the function.  Explanation: https://github.com/trestletech/plumber/issues/329

While maybe it is a bug that the code will be evaluated twice, this must be kept for desired behavior.

wch: This line can be removed once `activateBlock` is changed so that every expression given to it is `eval`ed.

tests/testthat/test-multibytes.R
```diff
@@ -0,0 +1,39 @@
...
+  force(expr)
+}
+
+expect_utf8 <- function(x) {
+  expect_equal(Encoding(x), "UTF-8")
+}
+
+test_that("support files with multibytes", {
+
+  with_locale("LC_CTYPE", "C", {
```
wch: Add comment about why `LC_ALL` isn't used.

tests/testthat/test-multibytes.R
```diff
@@ -0,0 +1,39 @@
...
+      # ?Quotes
+      # "Unicode escapes can be used to enter Unicode characters not in the current locale's
+      # charset (when the string will be stored internally in UTF-8)."
+      expected <- jsonlite::toJSON("\u4e2d\u6587\u6d88\u606f")
+
+      expect_identical(Sys.getlocale("LC_CTYPE"), "C")
+      expect_identical(Sys.getlocale("LC_COLLATE"), "C")
+      expect_utf8(expected)
+      expect_utf8(out)
+      expect_identical(out, expected)
```
wch: Use `charToRaw` for this test, since `identical` will automatically convert encodings.

wch:
We now understand why `source()` was being called. It shouldn't be necessary if `activateBlock()` is modified to do `eval(expr, envir)` when there's no plumber metadata associated with the expression.

PR #331, opened 2018-11-02
test that variables defined after a function may be used within the function

schloerke:
Test for https://github.com/trestletech/plumber/issues/330

### DIFF
--- a/tests/testthat/files/source_values.R
+++ b/tests/testthat/files/source_values.R
@@ -1,13 +1,24 @@
-#* @get /a
-(function(){
-  function() {
-    value_a
+#* @get /count
+{
+  if (!exists("count")) {
+    count <- 0
   }
-})()
-
-#* @get /b
-function() {
-  value_a
+  # will not init if variable exists,
+  # testing against sourcing an endpoint twice in the same envir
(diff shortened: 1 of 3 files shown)


tests/testthat/files/source_values.R
```diff
@@ -0,0 +1,13 @@
+#* @get /a
```
wch: Maybe replace this with a test that counts how many times the code has been evaluated.

```R
#* @get /count
{
  if (!exists("count")) {
    count <- 0
  }
  count <- count + 1
  function() {
    count
  }
}
```

tests/testthat/files/source_values.R
```diff
@@ -0,0 +1,13 @@
...
+(function(){
+  function() {
+    value_a
+  }
+})()
+
+#* @get /b
+function() {
```
wch: Replace this with:

```R

#* @get /static_count
function() {
  static_count
}

if (!exists("static_count")) {
  static_count <- 0
}
static_count <- static_count + 1
```

schloerke:
Test will fail as the current implementation of this branch is faulty.

PR #332, opened 2018-11-05
Atheriel remove connection close workaround

schloerke:
Fixes: #318

Added commit commenting on tcp test path originally created when bug was found.

PR #333, opened 2018-11-06
atheriel - Remove connection close workaround

schloerke:
Trying #332 with no squash


schloerke:
Was already merged. Am not merging in commits directly

PR #341, opened 2018-11-13
moved definition of private to initialize method in plumber object to…

shapenaji:
… prevent locking

This brings the serializer in line with the `private$errorHandler` and `private$notFoundHandler` in the plumber class in that they are defined during initialization rather than as part of the R6 definition (which appears to lock any functions)

This fixes https://github.com/trestletech/plumber/issues/339

### DIFF
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -168,6 +168,10 @@ plumber <- R6Class(
       }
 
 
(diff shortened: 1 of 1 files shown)


shapenaji:
Closing in lieu of a new PR with a proper email which will allow accepting the CLA

Issue #348, opened 2018-11-20
Error when type of dynamic route parameter is not supplied


blairj09:
Consider the following:
```r
library(plumber)

#* @param msg message to return to client
#* @get /msg/<msg>
function(msg = "") {
  list(msg = msg)
}
```

When this file is "plumbed", the following error occurs: 
```
Error in if (type == "bool" || type == "logical") { : 
  missing value where TRUE/FALSE needed
```
If the endpoint is changed to provide a type for the parameter, then it runs successfully. Both of the following work as expected.

```r
#* @param msg:character message to return to client
#* @get /msg/<msg>
function(msg = "") {
  list(msg = msg)
}
```

```r
#* @param msg message to return to client
#* @get /msg/<msg:chr>
function(msg = "") {
  list(msg = msg)
}
```

<details>
<summary>Session Info</summary>

``` r
devtools::session_info(pkgs = "plumber")
#> ─ Session info ──────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 3.5.1 (2018-07-02)
#>  os       macOS  10.14.1              
#>  system   x86_64, darwin15.6.0        
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  ctype    en_US.UTF-8                 
#>  tz       America/Denver              
#>  date     2018-11-19                  
#> 
#> ─ Packages ──────────────────────────────────────────────────────────────
#>  package  * version  date       lib source        
#>  BH         1.66.0-1 2018-02-13 [1] CRAN (R 3.5.0)
#>  crayon     1.3.4    2017-09-16 [1] CRAN (R 3.5.0)
#>  httpuv     1.4.5    2018-07-19 [1] CRAN (R 3.5.0)
#>  jsonlite   1.5      2017-06-01 [1] CRAN (R 3.5.0)
#>  later      0.7.5    2018-09-18 [1] CRAN (R 3.5.1)
#>  magrittr   1.5      2014-11-22 [1] CRAN (R 3.5.0)
#>  plumber    0.4.6    2018-06-05 [1] CRAN (R 3.5.0)
#>  promises   1.0.1    2018-04-13 [1] CRAN (R 3.5.0)
#>  R6         2.3.0    2018-10-04 [1] CRAN (R 3.5.1)
#>  Rcpp       1.0.0    2018-11-07 [1] CRAN (R 3.5.0)
[...]

Issue #347, opened 2018-11-20
Preempt fails for dynamic routes


blairj09:
Consider the following Plumber definition:
```r
library(plumber)

#* @filter nopass
function(req, res) {
  if (grepl("swagger", req$PATH_INFO, ignore.case = TRUE)) return(forward())
  list(text = "You shall not pass")
}

#* @param msg message to return to client
#* @preempt nopass
#* @get /msg
function(msg = "") {
  list(msg = msg)
}

#* @param msg message to return to client
#* @preempt nopass
#* @get /msg2/<msg:chr>
function(msg = "") {
  list(msg = msg)
}
```

The `/msg` endpoint properly preempts the `nopass` filter, while `/msg2` fails to skip the filter.

<details>
<summary>Session Info</summary>

``` r
devtools::session_info(pkgs = "plumber")
#> ─ Session info ──────────────────────────────────────────────────────────
#>  setting  value                       
#>  version  R version 3.5.1 (2018-07-02)
#>  os       macOS  10.14.1              
#>  system   x86_64, darwin15.6.0        
#>  ui       X11                         
#>  language (EN)                        
#>  collate  en_US.UTF-8                 
#>  ctype    en_US.UTF-8                 
#>  tz       America/Denver              
#>  date     2018-11-19                  
#> 
#> ─ Packages ──────────────────────────────────────────────────────────────
#>  package  * version  date       lib source        
#>  BH         1.66.0-1 2018-02-13 [1] CRAN (R 3.5.0)
#>  crayon     1.3.4    2017-09-16 [1] CRAN (R 3.5.0)
#>  httpuv     1.4.5    2018-07-19 [1] CRAN (R 3.5.0)
#>  jsonlite   1.5      2017-06-01 [1] CRAN (R 3.5.0)
#>  later      0.7.5    2018-09-18 [1] CRAN (R 3.5.1)
#>  magrittr   1.5      2014-11-22 [1] CRAN (R 3.5.0)
#>  plumber    0.4.6    2018-06-05 [1] CRAN (R 3.5.0)
#>  promises   1.0.1    2018-04-13 [1] CRAN (R 3.5.0)
#>  R6         2.3.0    2018-10-04 [1] CRAN (R 3.5.1)
#>  Rcpp       1.0.0    2018-11-07 [1] CRAN (R 3.5.0)
#>  rlang      0.3.0.1  2018-10-25 [1] CRAN (R 3.5.0)
[...]

PR #351, opened 2018-11-21
Check the length of regex in a dynamic path

byzheng:
solves #347 and #348

### DIFF
--- a/R/query-string.R
+++ b/R/query-string.R
@@ -71,10 +71,11 @@ createPathRegex <- function(pathDef){
 
 typeToRegex <- function(type){
   re <- rep("[^/]+", length(type))
-  re[type == "int"] <- "-?\\\\d+"
-  re[type == "double" | type == "numeric"] <- "-?\\\\d*\\\\.?\\\\d+"
-  re[type == "bool" | type == "logical"] <- "[01tfTF]|true|false|TRUE|FALSE"
-
+  if (length(re) > 0) {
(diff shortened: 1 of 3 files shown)


schloerke:
@byzheng The PR is currently "failing" as code coverage is decreased. (even if < 1%).  This is not a stopper for me.

I really like the PR.  Could you add a couple tests to validate that type `chr` works?

Thank you!

byzheng:
@schloerke  Add one test for chr. Not sure whether it is enough for you.

R/query-string.R
```diff
@@ -71,10 +71,11 @@ createPathRegex <- function(pathDef){
 
 typeToRegex <- function(type){
```
wch: A little simpler and clearer to do something like:

```R
if (length(type) == 0)
  return(character(0))
# Do the rest of the logic here
```

R/swagger.R
```diff
@@ -1,6 +1,9 @@
 #' Parse the given plumber type and return the typecast value
 #' @noRd
 plumberToSwaggerType <- function(type){
+  if (is.na(type)) {
```
wch: Why is this necessary now?

R/query-string.R
```diff
@@ -71,10 +71,11 @@ createPathRegex <- function(pathDef){
...
   re <- rep("[^/]+", length(type))
```
wch: It would help to have a comment explaining which cases get this default value (like `NA` and `"chr"`?)

PR #355, opened 2018-12-07
using httpuv url encode / decode

schloerke:
Fixes: #244

Alteration of PR originally proposed by @cliveseldon in #244 to use urltools url_decode / url_encode.  

Using httpuv to avoid an extra dependency


R/query-string.R
```diff
@@ -24,11 +23,11 @@ parseQS <- function(qs){
   kv <- kv[sapply(kv, length) == 2] # Ignore incompletes
 
   keys <- sapply(kv, "[[", 1)
-  keys <- unname(sapply(keys, URLdecode))
+  keys <- unname(sapply(keys, decodeURI))
```
wch: `decodeURI` is vectorized  so the `sapply` shouldn't be necessary.

Also, I'd suggest replacing the `sapply`s that are nearby with `lapply` and `vapply` as appropriate, to make the code more robust and easier to reason about.

R/cookie-parser.R
```diff
@@ -25,5 +25,5 @@ parseCookies <- function(cookie){
   cookies <- lapply(cookieList, "[[", 2)
   names(cookies) <- sapply(cookieList, "[[", 1)
 
-  return(lapply(cookies, URLdecode))
+  return(lapply(cookies, decodeURI))
```
wch: As mentioned below, `decodeURI` is vectorized, so it's faster to not use `lapply` (though you may need `as.list` if it's important that the return value is a list instead of character vector).

PR #360, opened 2018-12-17
Automatically Run on Restart

pachamaltese:
small change to activate that option (just `-d` didn't work)

### DIFF
--- a/docs/docker/index.md
+++ b/docs/docker/index.md
@@ -92,9 +92,9 @@ Then you'd be able to use `docker run --rm -v `pwd`:/app myCustomDocker` to run 
 
 ## Automatically Run on Restart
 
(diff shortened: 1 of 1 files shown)


schloerke:
Notes: 

* Adds options `-i` and `-t`.
https://docs.docker.com/engine/reference/run/#foreground :
> For interactive processes (like a shell), you must use -i -t together in order to allocate a tty for the container process


* Restart policy: https://docs.docker.com/engine/reference/run/#restart-policies---restart
  * Maybe should be `docker run --restart= unless-stopped`?

schloerke:
I'll follow up with a PR. Thank you!

schloerke:
This is on the docs.  I don't believe this will be to the docs that are currently displayed.  

Going to hold off for now until pkgdown is implemented.

PR #370, opened 2019-01-11
appveyor pkg cache will bust when DESCRIPTION changes

schloerke:
https://www.appveyor.com/docs/build-cache/#cache-dependencies

Was running to a missing SHA error as the branch was removed (but not the commit)


schloerke:
Cache bust worked.  Currently failing due to a recent build of curl not compiling.  Merging

PR #376, opened 2019-01-29
update host from 0.0.0.0 to 127.0.0.1 for swagger url only

### LINKED ISSUE #375: Why doesn't allow swagger running on 0.0.0.0
https://github.com/trestletech/plumber/blob/5fcf2dc62ee462522f2f4f60421d565b18d6841a/R/plumber.R#L220-L222

Hi, I don't understand why host 0.0.0.0 has to be masked by 127.0.0.1 if swagger gets enabled?

This leads to the case that __swagger is not available in docker__.

For example, if I run this:

`docker run --rm -v "/Users/Flynn/Downloads/shinyproxy-experiements/plumber:/plumber" -p 8000:8000 rdocker R -e "plumber::plumb('/plumber/api1.R')\$run(host = '0.0.0.0', port = 8000, swagger = TRUE)"`

Not only http://127.0.0.1:8000/__swagger__/ returns error but also the normal APIs like http://127.0.0.1:8000/plot ... The cause is that the host 0.0.0.0 gets masked by 127.0.0.1 thus will not be exposed to the docker's host machine.

So I have to remove the trailing `swagger = TRUE` in the above code.

schloerke:
Fixes #375 

@shrektan Could you try this out to see if it fixes the issue?   Do you think it's fair that if we see `0.0.0.0` coming in as the host, we could change the message to just say "Running Swagger", rather than "Running Swagger UI at X.X.X.X"?

I agree that there is no need to update the host location given that swagger is enabled as the `openapi.json` file is loaded as a relative path from the UI location.


shrektan:
Thanks for the quick fix but I think there's a bug that if the host is set to 0.0.0.0 via `option(plumber.apiHost = '0.0.0.0')`, the swaggerHost is still 0.0.0.0 .

In addition, would you mind to share me the reason that Swagger can't be running on 0.0.0.0 (a.k.a has to be replaced by 127.0.0.1)? Moreover, If 0.0.0.0 cannot be used as the host of Swagger, is it possible to expose the Swagger UI from a docker container?

schloerke:
@shrektan Currently, Swagger UI finds the `openapi.json` spec with relative urls so the actual host location is irrelevant to Swagger UI execution.  If you're manually setting `options(plumber.apiHost = "0.0.0.0")` (host location to view the API info in Swagger UI), I don't see why we wouldn't say to visit `0.0.0.0:1234`, as you set that value yourself.  

Swagger UI does work with `0.0.0.0`.
```r
options(plumber.apiHost = "0.0.0.0")
f <- system.file("examples/01-append/plumber.R", package = "plumber")
plumb(f)$run("0.0.0.0")
```
screen shot 2019-01-30 at 9 19 46 am

Where it does become important is with RStudio IDE (`swaggerCallback(swaggerHost)` in code).  In the [daily IDE](dailies.rstudio.com), the Swagger UI is auto displayed when a plumber router is the last item in a console execution.  However, the IDE does not like it when `0.0.0.0` is used. 

```r
options(plumber.apiHost = "0.0.0.0")
f <- system.file("examples/01-append/plumber.R", package = "plumber")
plumb(f)$run("0.0.0.0")
```
> screenshot from rstudio daily ide

screen shot 2019-01-30 at 9 30 32 am


If the `0.0.0.0` address is changed to `127.0.0.1` for Swagger UI only (with the api running on `0.0.0.0`, a capture all traffic interface), everything works as expected.

```r
options(plumber.apiHost = NULL)
f <- system.file("examples/01-append/plumber.R", package = "plumber")
plumb(f)$run("0.0.0.0")
```
> screenshot from rstudio daily ide

screen shot 2019-01-30 at 9 27 05 am

---------------------------------------------

I'm going to switch up how the messages are to be something similar to this...

```r
options(plumber.apiHost = NULL)
f <- system.file("examples/01-append/plumber.R", package = "plumber")
plumb(f)$run("0.0.0.0")
#> Running plumber API at 0.0.0.0:8296
#> Running Swagger UI at  0.0.0.0:8296/__swagger__/
```
> but in RStudio IDE, `swaggerCallback()` receives and opens window for `127.0.0.1:8296/__swagger__/`

schloerke:
@shrektan Could you please check this again? If a `0.0.0.0` host is provided, it should never change to `127.0.0.1` unless it is going to swaggerCallback (which is only really used by the RStudio IDE).

Thank you

R/plumber.R
```diff
@@ -271,16 +267,26 @@ plumber <- R6Class(
...
-        message("Running the swagger UI at ", swaggerUrl, sep = "")
+
+        swaggerHost <- getOption("plumber.apiHost", host)
+        swaggerUrl = paste0(swaggerHost, ":", port, "/__swagger__/")
+        message("Running Swagger UI at  ", swaggerUrl, sep = "")
 
         # notify swaggerCallback of plumber swagger location
         if (!is.null(swaggerCallback) && is.function(swaggerCallback)) {
+          # if within the RStudio IDE...
+          if (rstudioapi::isAvailable()) {
```
shrektan: I believe we can get rid of the dependence on rstudioapi by changing to

`if (requireNamespace('rstudioapi', quietly = TRUE) && rstudioapi::isAvailable()) {`

R/plumber.R
```diff
@@ -271,16 +267,26 @@ plumber <- R6Class(
...
-        swaggerUrl = paste0(host, ":", port, "/__swagger__/")
-        if (!grepl("^http://", swaggerUrl)) {
-          # must have http protocol for use within RStudio
-          # does not work if supplying "127.0.0.1:1234/route"
-          swaggerUrl <- paste0("http://", swaggerUrl)
-        }
-        message("Running the swagger UI at ", swaggerUrl, sep = "")
+
+        swaggerHost <- getOption("plumber.apiHost", host)
+        swaggerUrl = paste0(swaggerHost, ":", port, "/__swagger__/")
```
shrektan: The `=` should be `<-` in term of consistency of style... 😃

DESCRIPTION
```diff
@@ -25,7 +25,8 @@ Imports:
     stringi (>= 0.3.0),
     jsonlite (>= 0.9.16),
     httpuv (>= 1.4.5.9000),
-    crayon
+    crayon,
```
shrektan: I think need to specify the depends on the dev version of swagger... because the use of `swagger::swagger_spec()`

shrektan:
@schloerke Thanks for such a quick fix. It works on my computer. It works on the docker, too. Now I can easily expose both the API and the Swagger UI at the same time. 

Cheers! Thank again!

DESCRIPTION
schloerke: It's in `Suggests`.  Added a check to make sure it's installed

R/plumber.R
schloerke: Asking `rstudioapi` maintainer about `rstudioapi::isAvailable` internals to avoid a dependency for a single function

R/plumber.R
```diff
@@ -275,16 +271,27 @@ plumber <- R6Class(
...
 
         # notify swaggerCallback of plumber swagger location
         if (!is.null(swaggerCallback) && is.function(swaggerCallback)) {
+          # if within the RStudio IDE...
+          if (is_running_in_rstudio()) {
+            if (grepl("0.0.0.0:", swaggerUrl, fixed = TRUE)) {
+              # RStudio IDE does NOT like 0.0.0.0 locations.
+              # Must use 127.0.0.1 instead.
+              swaggerUrl <- sub("0.0.0.0:", "127.0.0.1:", swaggerUrl, fixed = TRUE)
+              message("Displaying Swagger API at ", swaggerURL, sep = "")
```
wch: UI

wch:
I think it would be good to always convert 0.0.0.0 to 127.0.0.1 (not just in RStudio) for the client target URL, and also for IPv6, convert `::` to `::1`.

See https://github.com/rstudio/shiny/blob/95173f6/R/server.R#L782-L787 for an example.

PR #387, opened 2019-02-22
add serializer_rds and serializer_rds3

schloerke:
Fixes #345

- [x] change name to serializer_rds
- [x] add params to serializer fn
- [x] add new serializer_rds3 with `version = "3"`
- [x] in serializer_rds, stop if R version is less than 3.5 as version 3 will fail


schloerke:
@shrektan Does this PR work for you?

shrektan:
@schloerke I've tested it. It works great. Thanks!

PR #388, opened 2019-02-22
Consolidate swagger type information

schloerke:
Fixes #351 

This consolidates all swagger type information into one location.  If a new type is supported in the future, only one spot will need to be added.


schloerke:
Old:
```r
switch(type,
"bool" = ,	
"logical" = "boolean",

"double" = ,
"numeric" = "number",

"int" = "integer",

"character" = "string"
)
```


New:
```r
addSwaggerInfo(
  "boolean",
  c("bool", "boolean", "logical"),
  "[01tfTF]|true|false|TRUE|FALSE",
  as.logical
)
addSwaggerInfo(
  "number",
  c("dbl", "double", "float", "number", "numeric"),
  "-?\\\\d*\\\\.?\\\\d+",
  as.numeric
)
addSwaggerInfo(
  "integer",
  c("int", "integer"),
  "-?\\\\d+",
  as.integer
)
addSwaggerInfo(
  "string",
  c("chr", "str", "character", "string"),
  "[^/]+",
  as.character
)
```

schloerke:
I am worried about this change.  https://github.com/trestletech/plumber/pull/388/files#diff-c6fcef256c9135dbef4ee90414a8074aR54

Should the regex be updated to only look for known types? Or should the regex look for a pattern of `<VAR:TYPE>` and complain if `TYPE` is unknown.

R/swagger.R
```diff
@@ -1,20 +1,85 @@
...
-plumberToSwaggerType <- function(type){
-  switch(as.character(type),
 
-    "bool" = ,
-    "logical" = "boolean",
 
-    "double" = ,
-    "numeric" = "number",
+# calculate all swagger type information at once and use created information throughout package
+swaggerTypeInfo <- (function() {
```
wch: Use `local()` instead -- it's a bit more idiomatic for R.

R/swagger.R
```diff
@@ -1,20 +1,85 @@
...
+  swaggerTypes <- c()
```
wch: Use one master list and nest the information inside of it.

schloerke:
This currently will cause an error if a type isn't found.

```
> tmp <- tempfile("tmp")
> cat(file = tmp, "
+ #* @get /barret/<barret:int>
+ function(req, res) {
+   'barret-int!'
+ }
+ #* @get /barret/<barret:asdf>
+ function(req, res) {
+   'barret-asdf!'
+ }
+ ")
> plumb(tmp)$run(port = 1234)
Error in plumberToSwaggerType(types) : Unrecognized type: asdf
```

Before, I would at least be able to find the `/barret/<barret:int>` route.  The `/barret/<barret:asdf>` route was not reachable.

Issue #390, opened 2019-02-27
swagger function does not handle tags correctly


EDiLD:
plumber with new swagger functionality does not handle tags of lenght one correctly.

[Here is a reprex](https://gist.github.com/EDiLD/9da84e3d51931a10b96191fb4e29c3d4).

With only one tag in /users this produces:

![image](https://user-images.githubusercontent.com/849616/53509995-44434500-3abd-11e9-8308-567c59eab77c.png)

http://localhost:1234/openapi.json:
```json
{"openapi":"3.0.3","info":{"description":"An API","version":"0.1.1.9000","title":"An API"},"tags":[{"name":"tag1","description":"tag1"}],"paths":{"/users":{"get":{"summary":"get endpoint","tags":"tag1","description":"an endpoint","parameters":[{"name":"parameter","in":"query","description":"a parameters","required":false,"schema":{"type":"string"}}],"responses":{"204":{"description":"No match not found."}}}}}}
```

With two tags this works as expected: 

![image](https://user-images.githubusercontent.com/849616/53510158-9edca100-3abd-11e9-81ef-c955d60d5f35.png)

http://localhost:1234/openapi.json:
```json
{"openapi":"3.0.3","info":{"description":"An API","version":"0.1.1.9000","title":"An API"},"tags":[{"name":"tag1","description":"tag1"}],"paths":{"/users":{"get":{"summary":"get endpoint","tags":["tag1","tags2"],"description":"an endpoint","parameters":[{"name":"parameter","in":"query","description":"a parameters","required":false,"schema":{"type":"string"}}],"responses":{"204":{"description":"No match not found."}}}}}}
```

Digging into plumbers, the problems is [here](https://github.com/trestletech/plumber/blob/e3610056929753fb009bfad9fe34620eab87d36a/R/plumber.R#L291) which auto_unboxes all atomic vectors of length 1, which should not be done with tags.

schloerke:
I believe this is a bug within `yaml::read_yaml`.  The tags supplied in `spec` for the swagger function have already been wrapped with `I(tags)` to prevent unboxing.  The `read_yaml` function does not know if it should be an atomic value or a list when reading, so it chooses string value.

EDiLD:
@schloerke That's true for the decorators (see https://github.com/trestletech/plumber/blob/e3610056929753fb009bfad9fe34620eab87d36a/R/plumber-step.R#L170), but not if we supply a custom swagger function? Or is this in the swagger package?

[...]

PR #391, opened 2019-02-27
fix issue #390 (tags should be boxed in swagger-function)

EDiLD:
By recursivly prevent unboxing of tags.


cpsievert:
Thanks for the PR @EDiLD

plumber 1.0 has a new function, `validate_api_spec()`, which helps in finding and diagnosing issues like this one. Closing here since `validate_api_spec()` provides users a general way to address this class of problem without us having to solve every situation internally using heuristics.

PR #398, opened 2019-03-14
adds svg plot option

pachamaltese:
works smooth!


pachamaltese:
Hi @trestletech. I've been using this modification for ~1 year and it works ok, a client wanted an api to show plots, and I did it with ggplot2 + plumber

schloerke:
This makes sense.  

How far should this go?  All major graphics devices?

Current:
`jpeg`, `png`, (`svg)`

Possible formats baked into R:
`bmp`, `tiff`, `pdf`, (any others?)

I don't want to import `Cairo`.

Don't know if we should namespace the image type. Ex: 

`#' @image jpeg` √
`#' @image tiff` √
`#' @jpeg` // allowed as a shorthand version of `#' @image jpeg`
`#' @tiff` // not possible

pachamaltese:
I need to think more, I see more value in the CSV PR, but I'll dig into this soon

schloerke:
No rush. Am tied up with `shiny` for the month.  Plan to dive deep on `plumber` in March.

PR #404, opened 2019-03-22
Use sodium instead of PKI

schloerke:
Fixes #242


jcheng5:
Oh, very important: when actually setting the session cookie, we must use `HTTPOnly`. Ideally we'd set the `Secure` option as well, but, I don't know if we have a way to know if the browser is accessing us through https?

R/session-cookie.R
```diff
@@ -28,44 +118,170 @@ sessionCookie <- function(key, name="plumber", ...){
...
+
+  if (
+    nchar(key) != 64 ||
+    grepl("[^0-9a-fA-F]", key)
+  ) {
+    warning(
+      "\n",
+      "\n\t!! Legacy cookie secret 'key' detected!                                         !!",
+      "\n\t!! Support for legacy cookie secret 'key' is deprecated and will be removed.    !!",
+      "\n\t!! Please see follow the instructions in `?sessionCookie` for a new secret key. !!",
```
jcheng5: "Please see follow" => "Please see"

"for a new" => "for creating a new"

R/session-cookie.R
```diff
@@ -1,25 +1,115 @@
...
+#' \item Do NOT store keys in databases which can be queried by everyone.
+#' }
+#'
+#' Instead, please: \itemize{
+#' \item Use a key management system, such as
+#' \href{https://github.com/r-lib/keyring}{'keyring'} (preferred)
+#' \item (or) Store it on a disk that only you have access to.
+#' Such as modifying the file permissions to "user read only"
+#' (\code{chmod 400 myfile.txt}) to your file from being modified and prevent
+#' others from reading your file.
```
jcheng5: "on a disk that only you have access to", that sounds like a USB key or something.

The next sentence is awkwardly worded as well.

How about:

> (or) Store the secret in a file on disk with appropriately secure permissions, such as "user read/write only" (\code{Sys.chmod("myfile.txt", mode = "0600")}), to prevent others from reading it.

jcheng5:
Other than those wording tweaks, LGTM!

R/session-cookie.R
schloerke: Should this be 400 or 600 permissions? (read only vs read/write).  

My thought is to safe guard against the password accidentally changing (or changing because of a program run by the user) by removing write permissions.

schloerke: If 600, I'll change the example as well

jcheng5: I can see both sides but 400 seems like overkill to me. There's a much larger risk in any of your .R files changing accidentally but we don't encourage anyone to set 400 on those. And I thought there might be a chance that some people wouldn't actually know how to open permissions up if they do want to edit or delete that file.

PR #416, opened 2019-04-26
Find function when only name is provided

schloerke:
Use the evaluated function not the raw expression.

# Testing Notes

```r
remotes::install_github("trestletech/plumber#416")
p <- callr::r_bg(
  function(){ 
tmp <- tempfile(fileext = ".R")
cat(file = tmp, '
api_msg <- function(msg = "") {
  list(msg = paste0("The message is: \\"", msg, "\\""))
}

# API ----
#* @apiTitle Simple Plumber API

#* Echo back the input
#* @get /echo
api_msg
')

plumber::plumb(tmp)$run(port = 1234)
  }
)

Sys.sleep(1)
x <- httr::GET("127.0.0.1:1234/echo?msg=asdf")
print(httr::content(x))
#> $msg
#> $msg[[1]]
#> [1] "The message is: \"asdf\""
p$kill()
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -8,4 +8,4 @@
 plumber.Rcheck
 plumber_*.tar.gz
 yarn.lock
-node_modules/*
+node_modules
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,6 +13,7 @@ jobs:
       r: oldrel
       if: |
         repo = trestletech/plumber AND
+        type = push AND
         branch IN (master, travis, rc-v0.5.0)
       addons:
         apt:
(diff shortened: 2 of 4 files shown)


schloerke:
cc @blairj09

blairj09:
This works on my end! One thing I noticed given the following example:
```r
# Setup ----
library(plumber)

source("utils.R")
api_msg <- function(msg = "") {
}

# API ----
#* @apiTitle Simple Plumber API

#* Echo back the input
#* @get /utils_echo
utils_msg

#* Echo back the input
#* @get /api_echo
api_msg
```

Where `utils_msg` is defined in `utils.R`. Both endpoints work as expected. However, when I terminate the Plumber API process after running it, `utils_msg` is persisted in my global environment while `api_msg` is not. Is this expected?

schloerke:
@blairj09 Try adding `local = TRUE` to the source call.

```r
source("utils.R", local = TRUE)
```

blairj09:
Ahh that works! This all looks good to me 😄

PR #443, opened 2019-06-05
Work around `cloud-init` which clobbers /etc/apt/sources.list after s…

trestletech:
…ome number of seconds.

The edits we were making to add the CRAN repo were getting clobbered.

We also separated into multiple separate SSH commands, though we don't believe that this is actually necessary.


schloerke:
Fixed in #448

PR #455, opened 2019-07-03
add content Type and Disposition

ycphs:
```r
reprex::reprex({
  library(plumber)
library(openxlsx)
#* @serializer contentTypeDisposition list(type="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet",disposition='attachment; filename="tmp.xlsx"')

#* @get /xlsx
function(){
   tmp <- tempfile(fileext = ".xlsx")
  data.frame(a=1:10,b=11:20)
write.xlsx(erg,tmp,withFilter=T)
  
  readBin(tmp, "raw", n=file.info(tmp)$size)
}
})
```

The serializer contentType returned named files without extension.

```r
 library(plumber)
library(openxlsx)
#* @serializer contentType list(type="application/vnd.openxmlformats-officedocument.spreadsheetml.sheet")

#* @get /xlsx
function(){
   tmp <- tempfile(fileext = ".xlsx")
  data.frame(a=1:10,b=11:20)
write.xlsx(erg,tmp,withFilter=T)
  
  readBin(tmp, "raw", n=file.info(tmp)$size)
}
```


ycphs:
corrected documentation

tests/testthat/test-serializer.R
```diff
@@ -114,6 +114,59 @@ test_that("nullSerializer serializes properly", {
...
+  expect_equal(val1$header$`Content-Type`,"text/plain")
+  expect_equal(val1$header$`Content-Disposition`, "attachment; filename=\"Test.csv\"")
+  expect_equal(val1$body,readBin(tmp, "raw", n=file.info(tmp)$size))
+})
+
+test_that("content type disposition error", {
+
+  expect_error(serializer_content_type_disposition(
+                                            disposition='attachment; filename="Test.csv"')(
+                                              readBin(tmp, "raw", n=file.info(tmp)$size),list(),PlumberResponse$new(),stop))
```
schloerke: there is no tmp file to read here

tests/testthat/test-serializer.R
```diff
@@ -114,6 +114,59 @@ test_that("nullSerializer serializes properly", {
...
+  expect_error(serializer_content_type_disposition(
+    type="text/plain")(
+      readBin(tmp, "raw", n=file.info(tmp)$size),list(),PlumberResponse$new(),stop))
+
+  # print(str(val1))
+  # expect_error()
```
schloerke: please remove these two lines

tests/testthat/test-serializer.R
```diff
@@ -114,6 +114,59 @@ test_that("nullSerializer serializes properly", {
...
+  expect_equal(val2$body,readBin(tmp, "raw", n=file.info(tmp)$size))
+})
+
+test_that("content type disposition error", {
+
+  expect_error(serializer_content_type()(
+      readBin(tmp, "raw", n=file.info(tmp)$size),list(),PlumberResponse$new(),stop))
+
+  # print(str(val1))
+  # expect_error()
```
schloerke: please remove these two lines

schloerke:
@ycphs Thank you for the PR and tests!

With content disposition being tied to content type, how do you feel about making `content_type` serializer better vs making a new method?

```r
# untested
serializer_content_type <- function(type, disposition = NULL, filename = NULL) {
  if (missing(type)) {
    stop("You must provide the custom content type to the serializer_content_type_disposition")
  }

  # https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Disposition
  if (!is.null(disposition)) {
    disposition <- match.arg(disposition, c("inline", "attachment"))
    if (disposition == "attachment" && !is.null(filename)) {
      filename <- as.character(filename)
      if (grepl("\"", filename, fixed = TRUE)) {
        stop("quotes can not be used in the value of `filename`")
      }
      disposition <- paste0(disposition, "; filename=\"", filename, "\"")
    }
  }

  function(val, req, res, errorHandler) {
    tryCatch({
      res$setHeader("Content-Type", type)
      if (!is.null(disposition)) {
        res$setHeader("Content-Disposition", disposition)
      }
      res$body <- val

      return(res$toResponse())
    }, error=function(e) {
      errorHandler(req, res, e)
    })
  }
}
```

ycphs:
@schloerke: Yes for me that's also good. 

Do you change it yourself or should I change it?

tests/testthat/test-serializer.R
ycphs: Changes Applied

tests/testthat/test-serializer.R
ycphs: Changes Applied

ycphs:
@schloerke I updated the method and tests.

ycphs:
change according of appveyor.yml to: [Provide more advice re: env vars #147](https://github.com/krlmlr/r-appveyor/pull/147)

cpsievert:
@schloerke 

* consider generalizing this so that other exported serializers can take advantage

* consider adding a `withHeaders` serializer

Issue #461, opened 2019-07-17
Incorrect cookie encoding


antoine-sachet:
Cookies should be encoded using `httpuv::encodeURIComponent` instead of `httpuv::encodeURI` and decoded accordingly.

Reference: [MDN](https://developer.mozilla.org/en-US/docs/Web/API/document/cookie)
> The cookie value string can use encodeURIComponent() to ensure that the string does not contain any commas, semicolons, or whitespace (which are disallowed in cookie values).

Issue #460, opened 2019-07-17
Incorrect decoding of URI components


antoine-sachet:
Parameters should be decoded using `httpuv::decodeURIComponent` instead of `httpuv::decodeURI`. 

Consider the string: `"This is über nice, 100% approved! +1"` which url-encodes to `"This%20is%20%C3%BCber%20nice%2C%20100%25%20approved%21%20%2B1"`.

This is decoded by plumber as `"This is über nice%2C 100% approved! %2B1"`. 

Indeed commas and plus signs are forbidden in URLs, but they should be decoded in a URI component.

To reproduce:

```R
library(plumber)

#* @get /echo
#* @post /echo
function(msg=""){
  list(msg = msg)
}
```

```bash
curl -Gv localhost:8413/echo --data-urlencode msg="This is über nice, 100% approved! +1"
# GET /echo?msg=This%20is%20%C3%BCber%20nice%2C%20100%25%20approved%21%20%2B1
# GET result: {"msg":["This is über nice%2C 100% approved! %2B1"]}

curl localhost:8413/echo --data-urlencode msg="This is über nice, 100% approved! +1"
# POST result: {"msg":["This is über nice%2C 100% approved! %2B1"]}
```

I am assuming this was an oversight but do let me know if there was a good reason to use `decodeURI`! 

As it is now, one is forced to re-decode the string, which I thought was fine but has edge cases. I just had the following issue in live code: `100%, yes` decoded as `100%%2C yes` by plumber and erroneously re-decoded as `100%2C yes`.

PR #462, opened 2019-07-17
Fix encoding/decoding of query parameters and cookies

antoine-sachet:
Plumber should use `decodeURIComponent` instead of `decodeURI` for query parameters, otherwise some characters are not correctly decoded in strings (#460 ).

Plumber should also use `encode/decodeURIComponent` when setting and reading cookies (#461). Otherwise, invalid characters such as `;` can appear in a cookie which breaks the cookie string. This is best practice as per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Document/cookie) .

Note that the latter impacted an unrelated cookie test in `test-sessions.R` which was parsing cookies manually without URL-decoding. Using `parseCookies` solved the issue, however I am not familiar with this part of the codebase so please do double-check!

PR task list:
- [X] Update NEWS
- [X] Add tests


R/cookie-parser.R
```diff
@@ -4,7 +4,7 @@ cookieFilter <- function(req){
   forward()
 }
 
-#' @importFrom httpuv decodeURI
+#' @importFrom httpuv decodeURIComponent
```
schloerke: I like the `httpuv::` namespacing below.  Can you remove the `@importFrom httpuv` as they are not necessary now?

PR #480, opened 2019-08-27
Do not require plumber.R when entrypoint.R is provided

### LINKED ISSUE #392: plumber.R shouldn't be a requirement if there's a valid entrypoint.R file
When a directory is plumbed (`plumber::plumb(dir = "/path/to/dir")`), Plumber looks for an `entrypoint.R` file that returns a runnable Plumber router and a `plumber.R` file that contains the logic for the Plumber router.

However, if the directory looks like this:
```
├── entrypoint.R
└── foo.R
```

Where `foo.R` contains the definition of the Plumber API and `entrypoint.R` "plumbs" `foo.R`, Plumber will complain when `plumber::plumb(dir = '.')` is run because a `plumber.R` file isn't found:
```r
 #> Error in plumber::plumb(dir = ".") :
 #>  No plumber.R file found in the specified directory: .
```
I think instead Plumber should run if a valid `entrypoint.R` file is found, regardless of whether the accompanying API definition is named `plumber.R`. If no `entrypoint.R` file is present, then I think it remains valid to enforce the `plumber.R` naming convention.

schloerke:
Fixes #471 
Fixes #392

Followup to  #471


R/plumber.R
```diff
@@ -4,69 +4,80 @@ NULL
...
-    # Both were explicitly set. Error
-    stop("You must set either the file or the directory parameter, not both")
+plumb <- function(file = NULL, dir = ".") {
+
+  if (!is.null(file) && !identical(dir, ".")) {
+    # both were explicitly set.
+    # assume it is a file in that dir and continue like normal
+    file <- file.path(
+      # removing trailing slash in dir
+      sub("/$", "", dir),
```
cpsievert, line 23: Doesn't seem necessary

R/plumber.R
```diff
@@ -4,69 +4,80 @@ NULL
...
+    )
+  }
 
-  } else if (missing(file)){
-    if (identical(dir, "")){
+  if (is.null(file)) {
+    if (identical(dir, "")) {
       # dir and file are both empty. Error
       stop("You must specify either a file or directory parameter")
     }
```
cpsievert, line 32: Seems like this would be rendered unnecessary with a `dir <- normalizePat(dir, mustWork = TRUE)`

schloerke: Going to keep the error as is.  The message is more useful than seeing `path:'' can not be found`.

R/plumber.R
```diff
@@ -4,69 +4,79 @@ NULL
...
-    # Both were explicitly set. Error
-    stop("You must set either the file or the directory parameter, not both")
+plumb <- function(file = NULL, dir = ".") {
+
+  if (!is.null(file) && !identical(dir, ".")) {
+    # both were explicitly set.
+    # assume it is a file in that dir and continue like normal
+    file <- file.path(
+      # removing trailing slash in dir
+      sub("/$", "", dir),
```
cpsievert, line 23:
```suggestion
      normalize_dir_path(dir),
```

R/plumber.R
```diff
@@ -4,69 +4,79 @@ NULL
...
+        entrypoint <- entrypoint[1]
+        warning("Found multiple files named 'entrypoint.R'. Using: '", entrypoint, "'")
+      }
 
-  entrypoint <- list.files(dir, "^entrypoint\\.r$", ignore.case = TRUE)
-  if (dirMode && length(entrypoint) > 0){
-    # Dir was specified and we found an entrypoint.R
+      # set working directory to dir before sourcing
+      old <- setwd(dir)
+      on.exit(setwd(old))
```
cpsievert, line 46:
```suggestion
      on.exit(setwd(old), add = TRUE)
```

R/plumber.R
```diff
@@ -4,69 +4,79 @@ NULL
...
-
-  if (!missing(file) && !missing(dir)){
-    # Both were explicitly set. Error
-    stop("You must set either the file or the directory parameter, not both")
+plumb <- function(file = NULL, dir = ".") {
+
+  if (!is.null(file) && !identical(dir, ".")) {
+    # both were explicitly set.
+    # assume it is a file in that dir and continue like normal
+    file <- file.path(
```
schloerke, line 21: Fix the test for this

PR #501, opened 2019-10-22
Ensures that mounted routers have a leading slash

### LINKED ISSUE #476: Mounted routers should require a slash for route location
On MacOS 10.14.16 with R 3.6.1 and Plumber 0.4.6, mounted routers are inaccessible. Consider the following example:
``` r
library(plumber)

root <- plumber$new()

# Mount first
first <- plumber$new()
first$handle("GET", "/echo", function(msg = "") {
  list(msg = paste0("The message is '", msg, "'"))
})
root$mount("first", first)

# Mount second
second <- plumber$new()
second$handle("GET", "/add", function(x, y) {
  as.numeric(x) + as.numeric(y)
})
root$mount("second", second)

root
#> # Plumber router with 0 endpoints, 4 filters, and 2 sub-routers.
#> # Call run() on this object to start the API.
#> ├──[queryString]
#> ├──[postBody]
#> ├──[cookieParser]
#> ├──[sharedSecret]
#> ├──/first
#> │  │ # Plumber router with 1 endpoint, 4 filters, and 0 sub-routers.
#> │  ├──[queryString]
#> │  ├──[postBody]
#> │  ├──[cookieParser]
#> │  ├──[sharedSecret]
#> │  └──/echo (GET)
#> ├──/second
#> │  │ # Plumber router with 1 endpoint, 4 filters, and 0 sub-routers.
#> │  ├──[queryString]
#> │  ├──[postBody]
#> │  ├──[cookieParser]
#> │  ├──[sharedSecret]
#> │  └──/add (GET)
```
This seems to indicate that `echo` would be available at `/first/echo` and `add` would be available at `/second/add`. However, when trying to access either of those endpoints, a 404 error is returned:

```bash
curl -X GET 'http://localhost:5762/first/echo?msg=hi'
{"error":["404 - Resource Not Found"]}
```

```bash
curl -X GET 'http://localhost:5762/second/add?x=1&y=9'
{"error":["404 - Resource Not Found"]}
```

atheriel:
This should fix #476. Since Plumber already ensures that there is a trailing slash, this just does the same for a leading slash as well, rather than issue an error (which was the other suggested solution).

The examples from that thread now work correctly.

PR task list:
- [x] Update NEWS
- [x] Add tests

### DIFF
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -322,7 +322,13 @@ plumber <- R6Class(
       httpuv::runServer(host, port, self)
     },
     mount = function(path, router){
-      path <- sub("([^/])$", "\\1/", path)
+      # Ensure that the path has both a leading and trailing slash.
+      if (!startsWith(path, "/")) {
+        path <- paste0("/", path)
+      }
+      if (!endsWith(path, "/")) {
+        path <- paste0(path, "/")
+      }
 
       private$mnts[[path]] <- router
     },
--- a/tests/testthat/test-plumber.R
+++ b/tests/testthat/test-plumber.R
@@ -150,9 +150,17 @@ test_that("mounts can be read correctly", {
 
   stat <- PlumberStatic$new(".")
   pr$mount("/static", stat)
+  pr$mount("missing-slashes", stat)
+  pr$mount("/both-slashes/", stat)
+  pr$mount("trailing-slash/", stat)
+  pr$mount("/extra-slash//", stat)
 
-  expect_length(pr$routes, 3)
+  expect_length(pr$routes, 7)
   expect_s3_class(pr$mounts[["/static/"]], "plumberstatic")
+  expect_s3_class(pr$mounts[["/missing-slashes/"]], "plumberstatic")
+  expect_s3_class(pr$mounts[["/both-slashes/"]], "plumberstatic")
+  expect_s3_class(pr$mounts[["/trailing-slash/"]], "plumberstatic")
+  expect_s3_class(pr$mounts[["/extra-slash//"]], "plumberstatic")
   expect_s3_class(pr$mounts[["/mysubpath/"]], "plumber")
 })
 
--- a/NEWS.md
+++ b/NEWS.md
@@ -54,6 +54,8 @@ plumber 0.5.0
 
 ### Bug fixes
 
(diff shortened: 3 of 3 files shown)


R/plumber.R
```diff
@@ -322,7 +322,8 @@ plumber <- R6Class(
       httpuv::runServer(host, port, self)
     },
     mount = function(path, router){
-      path <- sub("([^/])$", "\\1/", path)
+      # Ensure that the path has both a leading and trailing slash.
+      path <- sub("^/?(.*)([^/])$", "/\\1\\2/", path)
```
wch: This won't catch the case where `path` has a trailing slash but no leading slash.

```R
sub("^/?(.*)([^/])$", "/\\1\\2/", "abc/")
#> [1] "abc/"
```

In general, I get a little nervous when there's conditional logic in a regexp, so I would try to avoid it if possible, especially if there are multiple conditions you're testing for. I'd suggest using two separate `if` statements, with something like this: 

```R
if (substr(path, 1, 1) != "/") {
  path <- paste0("/", path)
}
if (substr(path, nchar(path), nchar(path)) != "/") {
  path <- paste0(path, "/")
}
```

It's very verbose, but at least it's clear what it's doing.

tests/testthat/test-plumber.R
```diff
@@ -150,9 +150,11 @@ test_that("mounts can be read correctly", {
 
   stat <- PlumberStatic$new(".")
   pr$mount("/static", stat)
+  pr$mount("missing-slash", stat)
```
wch: It would be good to have two more cases, with trailing slashes.

R/plumber.R
atheriel: I prefer `substr` or even `startsWith()` as well, I just stuck to a regex to minimize changes.

> This won't catch the case where path has a trailing slash but no leading slash.

Ah. I thought I had tested that, but clearly not.

tests/testthat/test-plumber.R
atheriel: Will do.

R/plumber.R
schloerke: +1 to `startsWith`

`?base::startsWith`
```
‘startsWith()’ is equivalent to but much faster than

     
     substring(x, 1, nchar(prefix)) == prefix
```

There's also an `endsWith`. TIL

atheriel:
Should be good to go now. I added some additional tests and moved to using `startsWith()` and `endsWith()` instead of a regex. I've listed @wch as a co-author on the commit, since he basically composed the new solution.

Issue #506, opened 2020-01-10
Annotation for replacing errorHandler?


jonkeane:
First, let me start off by saying thank you for an awesome flexible package for deploying APIs using R. 

I see in [the docs](https://www.rplumber.io/docs/rendering-and-output.html#error-handling) that one can set a custom error handler using something like (where`custom_error_handler` is a function that handles errors as they come up, similar to the [`defaultErrorHandler`](https://github.com/rstudio/plumber/blob/e22e4b052618829662baac321fce37eaa12c391d/R/default-handlers.R#L8-L30) used by Plumber) :

```
r <- plumb("plumber_api.R")
r$setErrorHandler(custom_error_handler)
r$run()
```
This works great and I was able to get my custom error handling working well. However, I didn't see (in the docs or source) a way to use an annotation like one can for creating serializers or filters for defining a custom error handler for an app.

I've forked the repo, and created [a working proof of concept ](https://github.com/rstudio/plumber/compare/master...jonkeane:errorHandler) for adding this which was based off of how filters and serializers were implemented. I've also tested this locally and publishing an API to RStudio Connect that includes and uses custom error handling (which is the use case I was looking for the errorhandling annotation for).

I'm happy to submit this as a PR if this approach is reasonable (after I add bit more checking for e.g. that only one errorhandler is provided).

schloerke:
Closing via https://github.com/rstudio/plumber/pull/507#issuecomment-675693897

PR #507, opened 2020-01-21
Add Error handler from within `plumb`

schloerke:
From https://github.com/rstudio/plumber/issues/506 via @jonkeane

### DIFF
--- a/R/parse-block.R
+++ b/R/parse-block.R
@@ -16,6 +16,7 @@ parseBlock <- function(lineNum, file){
   paths <- NULL
   preempt <- NULL
   filter <- NULL
+  errorhandler <- NULL
   image <- NULL
   imageAttr <- NULL
   serializer <- NULL
@@ -58,6 +59,22 @@ parseBlock <- function(lineNum, file){
       filter <- f
     }
 
+    errorMat <- stringi::stri_match(line, regex="^#['\\*]\\s*@errorhandler(\\s+(.*)$)?")
+    if (!is.na(errorMat[1,1])){
+      e <- stri_trim_both(errorMat[1,3])
+
+      # if (is.na(f) || f == ""){
+      #   stopOnLine(lineNum, line, "No @errorMat name specified.")
+      # }
+
+      if (!is.null(errorhandler)){
+        # Must have already assigned.
+        stopOnLine(lineNum, line, "Multiple @errorhandlers specified for one function.")
+      }
+
+      errorhandler <- e
+    }
+
     preemptMat <- stringi::stri_match(line, regex="^#['\\*]\\s*@preempt(\\s+(.*)\\s*$)?")
     if (!is.na(preemptMat[1,1])){
       p <- stri_trim_both(preemptMat[1,3])
@@ -90,6 +107,7 @@ parseBlock <- function(lineNum, file){
 
     serMat <- stringi::stri_match(line, regex="^#['\\*]\\s*@serializer(\\s+([^\\s]+)\\s*(.*)\\s*$)?")
     if (!is.na(serMat[1,1])){
+
       s <- stri_trim_both(serMat[1,3])
       if (is.na(s) || s == ""){
         stopOnLine(lineNum, line, "No @serializer specified")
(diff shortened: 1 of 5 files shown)


R/plumber.R
```diff
@@ -811,6 +811,10 @@ plumber <- R6Class(
     globalSettings = list(info=list()), # Global settings for this API. Primarily used for Swagger docs.
 
     errorHandler = NULL,
+    setErrorHandlerInternal = function(errorhandler){
```
schloerke: I would delete this method as there is already a public `setErrorHandler` method. (This can be reached using `self$setErrorHandler` in the `evaluateBlock` call above.)

R/plumber.R
```diff
@@ -220,7 +220,7 @@ plumber <- R6Class(
           srcref <- attr(e, "srcref")[[1]][c(1,3)]
 
           evaluateBlock(srcref, private$lines, e, private$envir, private$addEndpointInternal,
-                        private$addFilterInternal, self$mount)
+                        private$addFilterInternal, private$setErrorHandlerInternal, self$mount)
```
schloerke: Use `self$setErrorHandler`

R/plumber.R
jonkeane: Thanks, for this pointer that's exactly what I was missing when trying to access `setErrorHandler` earlier.

R/plumber.R
jonkeane: Done

jonkeane:
Ok, I've addressed the comments and clarified the error message. Sorry for the rebase I realized I had the wrong email address associated previously.

schloerke:
This PR is good.

However, I am hoping to go in a different direction completely. With an `#' @plumber` tag mixed with a pipeable interface. (both have not been created yet)

I'm going to add it to the next release milestone for tracking, but it should be handled by a different PR.

meztez:
Done in #568 , can be closed I think.

schloerke:
#568 did solve this. Thinking more in it... doesn't mean this can't also exist.

Notes for later...

Since this function is a one off function, like a filter, it should be treated like a filter. (Should not be able to add routes and an error handler at the same time.)

The `parseBlock` content should be changed to not capture to the right of `@errorhandler`. The content in the test is not used (and should not be used).  Within `parseBlock`, `errorhandler <- TRUE` if the regex is matched.

R/parse-block.R
```diff
@@ -236,7 +250,7 @@ parseBlock <- function(lineNum, file){
 #' Evaluate and activate a "block" of code found in a plumber API file.
 #' @include images.R
 #' @noRd
-evaluateBlock <- function(srcref, file, expr, envir, addEndpoint, addFilter, mount) {
+evaluateBlock <- function(srcref, file, expr, envir, addEndpoint, addFilter, setErrorHandler, mount) {
```
meztez, line 253: no need anymore since #568.

R/parse-block.R
```diff
@@ -273,6 +287,9 @@ evaluateBlock <- function(srcref, file, expr, envir, addEndpoint, addFilter, mou
     filter <- PlumberFilter$new(block$filter, expr, envir, block$serializer, srcref)
     addFilter(filter)
 
+  } else if (!is.null(block$errorhandler)){
+    setErrorHandler(eval(expr))
```
meztez, line 291: Since the plumberRouter is now passed to `evaluateBlock`, use pr$setErrorHandler directly

R/plumber.R
```diff
@@ -220,7 +220,7 @@ plumber <- R6Class(
           srcref <- attr(e, "srcref")[[1]][c(1,3)]
 
           evaluateBlock(srcref, private$lines, e, private$envir, private$addEndpointInternal,
-                        private$addFilterInternal, self$mount)
+                        private$addFilterInternal, self$setErrorHandler, self$mount)
```
meztez, line 223: No need since #568.

cpsievert:
Thanks for the PR @jonkeane.

Since we can already set a custom error handler via `plumber::pr_set_error()` (or the `$setErrorHandler()` method), we're going to pass on this short-hand plumb method since, if we do it here, then we'll need to do it for a bunch of other methods (which doesn't seem worthwhile doing). Here's an example:

```r
handler <- function(req, res, err) {
  rlang::abort(err)
}

#* @plumber 
function(pr) {
  pr %>% pr_set_error(handler)
}
```

Issue #517, opened 2020-03-27
postserialize hook doesn't respect image serializers


blairj09:
### System details

Output of `sessioninfo::session_info()`:

``` r
sessioninfo::session_info()
─ Session info ────────────────────────────────────────────────────────────────────────────────────────────────────
 setting  value                       
 version  R version 3.6.3 (2020-02-29)
 os       macOS Catalina 10.15.3      
 system   x86_64, darwin15.6.0        
 ui       RStudio                     
 language (EN)                        
 collate  en_US.UTF-8                 
 ctype    en_US.UTF-8                 
 tz       America/Denver              
 date     2020-03-27                  

─ Packages ────────────────────────────────────────────────────────────────────────────────────────────────────────
 ! package     * version     date       lib source                           
 P assertthat    0.2.1       2019-03-21 [?] CRAN (R 3.6.0)                   
 P cli           2.0.2       2020-02-28 [?] CRAN (R 3.6.2)                   
 P crayon        1.3.4       2017-09-16 [?] CRAN (R 3.6.0)                   
 P fansi         0.4.1       2020-01-08 [?] CRAN (R 3.6.0)                   
 P fs            1.3.2       2020-03-05 [?] CRAN (R 3.6.0)                   
 P glue        * 1.3.2       2020-03-12 [?] CRAN (R 3.6.3)                   
   httpuv        1.5.2.9000  2020-03-27 [1] Github (rstudio/httpuv@622c76a)  
 P jsonlite      1.6.1       2020-02-02 [?] CRAN (R 3.6.0)                   
 P later         1.0.0       2019-10-04 [?] CRAN (R 3.6.0)                   
 P logger      * 0.1         2019-01-02 [?] CRAN (R 3.6.0)                   
 P magrittr      1.5         2014-11-22 [?] CRAN (R 3.6.0)                   
   plumber     * 0.4.7.9000  2020-03-27 [1] Github (rstudio/plumber@54fc7f7) 
   promises      1.1.0.9000  2020-03-27 [1] Github (rstudio/promises@627dfc6)
 P R6            2.4.1       2019-11-12 [?] CRAN (R 3.6.0)                   
   Rcpp          1.0.4.5     2020-03-27 [1] Github (RcppCore/Rcpp@7416859)   
   renv          0.9.3       2020-02-10 [1] CRAN (R 3.6.2)                   
 P rstudioapi    0.11        2020-02-07 [?] CRAN (R 3.6.0)                   
[...]

PR #518, opened 2020-03-29
Image serializer

meztez:
Suggestion to fix issue #517. 

Modified images handling to serialize during serialization step instead of postexec step

Comments welcome.

 PR task list:
 
 * [x]  Update NEWS
 * [ ]  Add tests
 * [ ]  Update documentation with `devtools::document()`


schloerke:
Seems odd to set the `res$serializer` in a `post` key val.  

Note: Look into `pre`/`post` and what they mean

meztez:
I agree, image handling seems to be its own thing. Maybe it could be refactored to match the rest of the components?

meztez:
it is preexec and postexec, I was mistaken.

R/images.R
```diff
@@ -2,32 +2,31 @@
...
-      res
+      unlink(data$file)
+      img
     }
   )
 }
 
 render_jpeg <- function(args){
-  render_image(jpeg, "image/jpeg", args)
+  render_image(jpeg, args)
```
cpsievert, line 27:
```suggestion
  render_image(grDevices::jpeg, args)
```

R/images.R
```diff
@@ -2,32 +2,31 @@
...
 }
 
 render_png <- function(args){
-  render_image(png, "image/png", args)
+  render_image(png, args)
```
cpsievert, line 31:
```suggestion
  render_image(grDevices::png, args)
```

R/images.R
```diff
@@ -2,32 +2,31 @@
...
+    postexec = function(value, req, res, data){
       dev.off()
 
       con <- file(data$file, "rb")
       img <- readBin(con, "raw", file.info(data$file)$size)
       close(con)
-      res$body <- img
-      res$setHeader("Content-type", contentType)
-      res
+      unlink(data$file)
```
cpsievert, line 20: Consider doing this in an on.exit() before reading (in case it fails)

Issue #421, opened 2019-05-01
`removeNAOrNulls` breaks on no swagger spec


schloerke:
From #417 

```r
pr <- plumber$new()
pr$handle("GET", "/:path/here", function(){})

pr$run(
  port = 1234,
  swagger = function(pr_, spec, ...) {
    spec$info$title <- Sys.time()
    spec
  }
)
```

Visiting http://127.0.0.1:1234/openapi.json causes an error

```
<simpleError in if (any(toRemove)) {    x[toRemove] <- NULL}: missing value where TRUE/FALSE needed>
```

meztez:
Maybe in isNA?

``` r
> plumber:::isNa
function (x) 
{
    if (is.list(x) || length(x)==0) {
        return(FALSE)
    }
    is.na(x)
}
<bytecode: 0x000001ab119c6b68>
<environment: namespace:plumber>
```

schloerke:
@meztez 

A PR would be great!

Two things come to mind. 

1. Yes, we should check for length.  I believe if the length is 0, it should be removed.
2. The default url is coming in as `character(0)`.  The url is a required field.  I believe this should default to `http://127.0.0.1:PORT` so that it isn't removed.  I don't know why that isn't happening already.

Thank you for the help!

schloerke:
Weird.  The url is being supplied by the referrer.  This should not be missing.  Maybe a header name changed.  If not, we should set a sensible default.

https://github.com/trestletech/plumber/blob/6e3417e34d8a6da260fbd8d86d61043eae0c546e/R/plumber.R#L268-L276

meztez:
I'll look into it and submit something sensible.

meztez:
Notes : 

So this path http://127.0.0.1:8004/__swagger__/ has a HTTP_REFERER and this one http://127.0.0.1:8004/openapi.json has not (it is NULL). Might be something in httpuv, down the rabbit hole I go.

schloerke:
Hmmmmm before we go down that route, it might be a logical flaw.

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referer

A referrer is an optional header.  Maybe the url should be be fixed to be `http://127.0.0.1:PORT` as that is the URL location of the server. (Or set with an option.  I can add the option later if you get the default url location.)

schloerke:
> A Referer header is not sent by browsers if:
> 
> * The referring resource is a local "file" or "data" URI.
[...]

PR #528, opened 2020-04-18
Swagger server url / OpenAPI server url

meztez:
For #421 

Fallback to `req$HTTP_REFERER` when available
OpenAPI server url property is left empty if options are not set to avoid leaking info on `openapi.json` endpoint.

New plumber options to document somewhere.

- `plumber.apiURL` (has precedence over the others)
- `plumber.apiScheme`
- `plumber.apiPort`
- `plumber.apiPath`

Trailing `/` is removed from the OpenAPI server url property per [docs](https://swagger.io/docs/specification/api-host-and-base-path/).

Console message about Swagger UI uses options if they are set.
Modified `urlHost` function and calls to add `scheme` and `path` with default `http` and `character(1)`

Best practice should be to either set `plumber.apiURL`
```r
options("plumber.apiURL" = "https://el.gato/bueno/v1")
```
or use individual options.
```r
options("plumber.apiScheme" = "https")
options("plumber.apiHost" = "el.gato")
options("plumber.apiPort" = 80)
options("plumber.apiPath" = "/bueno/v1")
```
In this case `port` should be provided since it will default to `httuv` serving port when not set.

 PR task list:
 
 * [x]  Update NEWS
 * [x]  Add tests
 * [ ]  Update documentation with `devtools::document()`

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -54,6 +54,9 @@ plumber 0.5.0
 
 ### Bug fixes
 
+* Fix bug in setting openapi.json url when HTTP_REFERER is null (@meztez,
+[#421](https://github.com/rstudio/plumber/issues/421))
+
 * Fix possible bugs due to mounted routers without leading slashes (@atheriel, [#476](https://github.com/rstudio/plumber/issues/476) [#501](https://github.com/rstudio/plumber/pull/501)).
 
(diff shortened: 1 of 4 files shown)


meztez:
See #536, better handling there.

Issue #522, opened 2020-04-08
Document R6 methods using new roxygen2


schloerke:
See https://roxygen2.r-lib.org/articles/rd.html#r6

The base plumber object has a lot of autogenerated and missing documentation. This causes a few warnings.

`devtools::document()`

meztez:
I'll help, I have a fews days to put on the plumber

schloerke:
Fixed in #530

PR #530, opened 2020-04-19
update doc R6, fix docs check

meztez:
per #522 

updated documentation on R6 class
fix doc link for sodium random -> helpers
added custom swagger spec examples

comments welcome

mental check list:
remove deprecated plumber methods? or mark for deletion
should we add more documentation to initialize of PlumberEndpoint and PlumberStep?
should they even be exported? Maybe for historical reason.

 PR task list:
 
 * [x]  Update NEWS
 * [ ]  Add tests
 * [x]  Update documentation with `devtools::document()`


R/plumber-static.R
```diff
@@ -75,6 +79,12 @@ PlumberStatic <- R6Class(
       filter <- PlumberFilter$new(paste("static-asset", direc, sep="|"), expr, private$envir)
       private$addFilterInternal(filter)
     },
+    #' @description Print reprensation of plumberstatic router.
```
schloerke, line 82:
```suggestion
    #' @description Print representation of plumberstatic router.
```

R/plumber-static.R
```diff
@@ -75,6 +79,12 @@ PlumberStatic <- R6Class(
...
+    #' @param prefix a character string. Prefix to append to representation.
```
schloerke, line 83:
```suggestion
    #' @param prefix a character string. Prefix to append to representation.
```

R/plumber-static.R
```diff
@@ -75,6 +79,12 @@ PlumberStatic <- R6Class(
...
+    #' @param topLevel a logical value. When method executed on top level
+    #' router, set to `TRUE`.
+    #' @param ... additional arguments for recursive calls
+    #' @return A terminal friendly represention of a plumberstatic router.
```
schloerke, line 87:
```suggestion
    #' @return A terminal friendly represention of a `PlumberStatic()` router.
```

R/plumber-step.R
```diff
@@ -20,12 +20,23 @@ resetForward <- function() {
...
+#' of a request by a plumber router.
 PlumberStep <- R6Class(
   "PlumberStep",
   inherit=hookable,
   public = list(
+    #' @field lines lines from step block
     lines = NA,
+    #' @field serializer step serializer function
     serializer = NULL,
+    #' @description Create a new `PlumberStep` object
```
schloerke, line 34: Ah, just remembered.  Can you change all  `` `func` `` to `[func()]`.  This will allow for pkgdown to link to the proper help pages for free.

```suggestion
    #' @description Create a new [PlumberStep()] object
```

(I forgot in my earlier changes above)

schloerke:
> should we add more documentation to initialize of PlumberEndpoint and PlumberStep?
> should they even be exported? Maybe for historical reason.

We need to keep them for historical reasons. They can be minimally documented. 

Can we add a `@keywords internal` so that we don't have to do a pkgdown link, but documentation is still produced?

meztez:
I'll let you go through this and I'll circle around later.

R/plumber.R
```diff
@@ -332,12 +398,84 @@ plumber <- R6Class(
...
+    #' processing this same request.\cr
+    #' \cr
+    #' One feature when defining hooks in Plumber routers is the ability to modify
+    #' the returned value. The convention for such hooks is: any function that accepts
+    #' a parameter named `value` is expected to return the new value. This could
+    #' be an unmodified version of the value that was passed in, or it could be a
+    #' mutated value. But in either case, if your hook accepts a parameter
+    #' named `value`, whatever your hook returns will be used as the new value
+    #' for the response.\cr
+    #' \cr
```
schloerke, line 425:
```suggestion
    #' processing this same request.
    #'
    #' One feature when defining hooks in Plumber routers is the ability to modify
    #' the returned value. The convention for such hooks is: any function that accepts
    #' a parameter named `value` is expected to return the new value. This could
    #' be an unmodified version of the value that was passed in, or it could be a
    #' mutated value. But in either case, if your hook accepts a parameter
    #' named `value`, whatever your hook returns will be used as the new value
    #' for the response.
    #'
```

inst/examples/15-swagger-spec/plumber_custom_swagger_spec.R
```diff
@@ -0,0 +1,49 @@
...
+  list("name" = multipart$somefile$name,
+       "tmp_name" = multipart$somefile$datapath,
+       "size" = file.size(multipart$somefile$datapath))
+}
+
+pr$handle("POST", "/upload", handler, serializer = serializer_json())
+
+pr$run(swagger = swagger)
+
+#In case you have have problems, it is wise to insert a browser() in your swagger function
```
schloerke, line 49:
```suggestion
#In case you have have problems, insert a `browser()` in your swagger function
```

R/session-cookie.R
```diff
@@ -146,7 +146,7 @@ sessionCookie <- function(
 
 #' Random cookie key generator
 #'
-#' Uses a cryptographically secure pseudorandom number generator from \code{sodium::\link[sodium]{random}} to generate a 64 digit hexadecimal string.  \href{https://github.com/jeroen/sodium}{'sodium'} wraps around \href{https://download.libsodium.org/doc/}{'libsodium'}.
+#' Uses a cryptographically secure pseudorandom number generator from \code{sodium::\link[sodium]{helpers}} to generate a 64 digit hexadecimal string.  \href{https://github.com/jeroen/sodium}{'sodium'} wraps around \href{https://download.libsodium.org/doc/}{'libsodium'}.
```
schloerke, line 149: I don't know if it'll need the `()`. `[sodium::helpers()]`

```suggestion
#' Uses a cryptographically secure pseudorandom number generator from [sodium::helpers] to generate a 64 digit hexadecimal string.  \href{https://github.com/jeroen/sodium}{'sodium'} wraps around \href{https://download.libsodium.org/doc/}{'libsodium'}.
```

Issue #187, opened 2017-10-24
Swagger parameters listed in reverse order


alexzaytsev-newsroomly:
My method has a bunch of parameters like so:
```
#* @post /dosomething
do_something <- function(par1, dataframe, somebool = TRUE, someint = 0)
{}
```

However, when i launch the plumber server using the sample code provided, swagger shows 'no parameters' for my action? 

What should i add to list them.

dselivanov:
@trestletech same issue. Is there any instruction how to point swagger to show parameters? do we miss something?

jeffkeller87:
The documentation for decorating code to generate Swagger documentation needs a lot of work, but I've discovered a few decorations so far. The one you are looking for is `@param`.

```
#' @param par1 description text
```

> Note: you can use `#'` instead of `#*` if you like.

Strangely, parameters documented this way appear in reverse order in the rendered Swagger document. So to get them to appear in the way you are likely familiar with, your code would look something like this:

```
#' @post /dosomething
#' @param someint description
#' @param somebool description
#' @param dataframe description
#' @param par1 description
do_something <- function(par1, dataframe, somebool = TRUE, someint = 0)
{}
```

Some others you might find useful:

#' @apiTitle Title text
#' @apiDescription Description text

trestletech:
Sorry I'm so late to the party, and thanks for the tip @jeffkeller87 .

PRs welcomed on the docs to start fleshing this stuff out. 

I just renamed the ticket and I'll keep it open to note the bug pointed out above that parameters are in reverse order. That seems like something we should fix.

meztez:
A simple fix for that would be to do `rev(params)` there

https://github.com/rstudio/plumber/blob/b7ff05ac6bd21db3341a9a91fb90c4c581bf56b9/R/parse-block.R#L229

mskyttner:
Using plumber 1.0, it seems "multiline" summary descriptions in endpoint annotations still get reversed in the generated docs?

In the screenshot the endpoint summary description is broken into two lines (upper part of picture), which get reversed in the docs (green field, lower part of picture).

Does everything in the annotation need to go into one long line? 

![image](https://user-images.githubusercontent.com/1715840/101932000-73676f80-3bda-11eb-9732-4c8685dbace5.png)

This document http://spec.openapis.org/oas/v3.0.3#fixed-fields-6 has two fields described:

summary: An optional, string summary, intended to apply to all operations in this path.
[...]

Issue #523, opened 2020-04-10
parameter default values for Swagger


bhogan-mitre:
The documentation nicely describes default values for the handling functions that are behind the endpoints, but those are not carried through to the Swagger interface.  A user could submit the form without a parameter value and the default would be applied behind the scenes, _but they would not know that from looking at the interface_. 

It would be useful to carry through the default values into Swagger.  I believe that data would need to be provided in the `schema` [here](https://github.com/rstudio/plumber/blob/master/R/swagger.R#L145-L153):
```
    paramList <- list(
      name = p,
      description = endpointParams[[p]]$desc,
      `in` = location,
      required = endpointParams[[p]]$required,
      schema = list(
        type = type
      )
    )
```

Here's an example per the [Swagger documentation](https://swagger.io/docs/specification/describing-parameters/)
```
      parameters:
        - in: query
          name: offset
          schema:
            type: integer
            minimum: 0
            default: 0
          required: false
          description: The number of items to skip before starting to collect the result set.
```

Thanks for considering.

Issue #516, opened 2020-03-27
Support arrays as query parameters


mattyb:
Standard libraries construct array query parameters as below:
```
http://hostname/endpoint?colors[]=blue&colors[]=green
```

If plumber sees this for an endpoint with a `colors` parameter, it should construct a list and pass it to the function.

meztez:
What you propose would not be supported by OpenApi standard.
See https://swagger.io/docs/specification/serialization/ and https://stackoverflow.com/questions/11490326/is-array-syntax-using-square-brackets-in-url-query-strings-valid

But something could be done by defining plumber param as 
```
#* @param number:numeric[] a numeric array
```

Since it is supported in OpenApi, I'll look into how that could be done.

Issue #512, opened 2020-03-17
Send data frame to post request in body 


mansi-aggarwal-2504:
I want a data frame to a @post request body, and do something (say, time series analysis) and return a data frame
Or maybe send JSON and inside the post function, convert it to a data frame and do something and return a data frame
I am new to plumber and am having trouble understanding this

shrektan:
Converting to JSON is feasible but you have to handle these things by yourself - zero-row table, the Date attributes, etc.

So if you only want to pass the `data.frame`, a safer way is to use the `rds` serializer. 

### plumber

```r
#* @post /api
#* @serializer rds
function() {
  ...
}
```

### users

```r
out <- httr::POST(
  url,
  encode = "raw",
  body = body,
  httr::content_type("application/octet-stream"),
  ...
)
# you may need to check httr::status_code() == 200L 
# or if is.raw(httr::content(out)) is TRUE, first
base::unserialize(httr::content(out))
```

mansi-aggarwal-2504:
Thank you for your comment; but I have a query.
This is what I am doing:
```
#'@post /testdf
#'@serializer rds
function(req){
  data <- (req$postBody)
  return(ncol(data))
}
```
User:
```
data <- read.csv("XYZ.csv")

result <- POST(url = 'http://127.0.0.1:8000/testdf', body = data, encode = "raw"
                   httr::content_type("application/octet-stream"))
```
But it says this: 
`Error in body_raw(body) : is.raw(body) is not TRUE`

Can you tell me what I am doing wrong here

shrektan:
Sorry, I didn't notice that you were asking "POSTING the data.frame". I misunderstood for "REQUSTING the data.frame". There're workarounds for this. I remembered https://github.com/ozean12/protopretzel uses this. 

However, I don't have time to write an example for you. But if you really need it, I can make an example later.

mansi-aggarwal-2504:
I didn't find what I am looking for here at https://github.com/ozean12/protopretzel 
Thanks for the help, though. If you could provide and example whenever you can, that'd be great. Thanks!

shrektan:
@mansi-aggarwal-2504 It's there actually. Here the example. Put the three files in the same folder. Run `main.R` to launch a server. Run `client.R` to test.

### plumber.R

```r
#* @post /api
#* @serializer rds
function(req) {
  req$robj
}
```

### main.R

```r
library(plumber)
x <- plumb("plumber.R")
x$filter("robj", function(req) {
[...]

PR #532, opened 2020-04-27
Swaggerfile support for array parameters, object, files

meztez:
Fixes https://github.com/rstudio/plumber/issues/187

Add swagger support for array parameters per https://swagger.io/docs/specification/serialization/ using square brackets around typed parameters.

Add swagger support for openapi `object` type to be listed under `requestBody` using plumber types `list`, `df` and `data.frame`.

Add swagger/plumber support for files upload, multipart content-type

Add support for detection of endpoint plumberExpression parameters.

Add support for custom body parsers, like serializers, more raw.

Around 30% overheads reduction for POST request with json body.

Syntax support for parameters would become :
- **in: query**:
  -  name
  -  name:type
  -  name:type*
  -  _new_ name:[type]*
- **in: path** 
  -  <name\>
  -  \<name:type\>
  -  _new_ \<name:[type]\>

So something like below works with a matching openapi file to go with it.
```r
library(plumber)

#* Sum numbers
#* @get /names3/<number:[integer]>/path
function(number) {
  sum(c(number, 5L))
}

#* Do something
#* @param iris2:df* Use something like jsonlite::toJSON(list(iris2=iris), auto_unbox = TRUE)
#* @png
#* @post /smoore
function(iris2) {
  iris2$Species <- as.factor(iris2$Species)
  plot(iris)
}

#* Do some more
#* @param owners:[chr]
#* @param dogId:[int]
#* @get /whosagoodboy
function(owners, dogId) {
  #query string is still parsed as character
  dogId <- as.integer(dogId)
  list(sample(owners, 1), sample(dogId , 1))
}

#* Will that work? Test detection, examples and object type
#* @post /risky
function(owners = list(name = c("Bob", "Larry"),
                       cars = c("Corvette", "Tesla")),
         dogId = 1L:2L,
         moredata = women) {
  list(owners_equal = all.equal(owners, list(name = c("Bob", "Larry"), cars = c("Corvette", "Tesla"))),
       dogId_equal = all.equal(as.integer(dogId), 1L:2L),
       moredata_equal = all.equal(moredata, women))
}

#* What about a file and some garbage?
#* @param filename:file Test upload file
#* @param filename2:file Test upload file2
#* @post /upload
function(filename, filename2, fakenews = list(a = list(b = 2, c =3))) {
  list(filename, lapply(file.info(filename), as.character),
       filename2, lapply(file.info(filename2), as.character),
       fakenews)
}
```

Notes:
 `object` parameters would not be supported in path parameters. 
Default upload location set to getOption("plumber.upload.dir", "."). Use multipart parse from mime package. (copied with very slight change to buffer size, filename decoding and file args response) 
openapi field `example` for `object` feeded from plumberExpression arguments.

Fixes:
- Swagger definition absent when defining a parameter in path that is not on a @param line.
- Prevent example from being removed by swagger remove null
- Swagger parameters now show in the order they were defined

Breaking:
req$postBody is raw since it can contains file and other breaking havoc on string elements.

Related:
#516, #512, #469, #509, #417, #523, #502, #463, #320

At this point I would need help to do more testing across different platforms.

Thank you, comments are welcome.

PR task list:
- [x] Update NEWS
- [x] Add tests
See other PR for documentation, will update there if this one pass.


meztez:
Performance testing on a puny old laptop from the ages with library `loadtest`, simple enough.
TLDR : it is faster.

```r
library(plumber)
tmpf <- tempfile()
writeLines('
#* Echo back the input
#* @param msg:character The message to echo
#* @post /test8
function(msg = list(part1 = "a", part1 = "b", part1 = "c")) {
  do.call(paste, msg)
}
', tmpf)
pr <- plumb(tmpf)
pr$run(port = 8477, swagger = FALSE)
```

[plumber_0.4.6.pdf](https://github.com/rstudio/plumber/files/4584399/plumber_0.4.6.pdf)
[plumber_0.4.7.9000.pdf](https://github.com/rstudio/plumber/files/4584400/plumber_0.4.7.9000.pdf)
[plumber_PR532.pdf](https://github.com/rstudio/plumber/files/4584401/plumber_PR532.pdf)

R/parse-block.R
```diff
@@ -226,7 +218,7 @@ parseBlock <- function(lineNum, file){
     imageAttr = imageAttr,
     serializer = serializer,
     assets = assets,
-    params = params,
+    params = rev(params),
```
schloerke, line 221: why rev?

schloerke: https://github.com/rstudio/plumber/issues/187

R/parse-block.R
```diff
@@ -166,39 +167,30 @@ parseBlock <- function(lineNum, file){
...
-      reqd <- FALSE
-      if (!is.na(nameType[1,4])){
-        reqd <- nameType[1,4] == "*"
+      if (is.na(name)){
+        stopOnLine(lineNum, line, "No parameter specified.")
       }
-      params[[name]] <- list(desc=paramMat[1,5], type=type, required=reqd)
+      type <- stri_replace_all(paramMat[1,4], "$1", regex = "^\\[([^\\]]*)\\]$")
+      type <- plumberToSwaggerType(type)
+      serialization <- stri_detect_regex(paramMat[1,4], "^\\[[^\\]]*\\]$")
```
schloerke, line 185: `serialization` -> `isArray`

R/post-body.R
```diff
@@ -1,34 +1,47 @@
 postBodyFilter <- function(req){
   handled <- req$.internal$postBodyHandled
-  if (is.null(handled) || handled != TRUE){
-    body <- paste0(req$rook.input$read_lines(), collapse = "\n")
-    charset <- getCharacterSet(req$HTTP_CONTENT_TYPE)
-    args <- parseBody(body, charset)
-    req$postBody <- body
+  if (is.null(handled) || handled != TRUE) {
+    body <- req$rook.input$read()
```
schloerke, line 4: Add comment about this returning raw bytes

R/post-body.R
```diff
@@ -1,34 +1,47 @@
...
+    type <- req$HTTP_CONTENT_TYPE
+    args <- parseBody(body, type)
+    req$postBody <- body #this is a raw body contrary to what plumber did before
```
schloerke, line 7: Need to maintain the original behavior of `postBody`

Could get around this by allowing for a global option

```r
if (getOption("plumber.postBody", TRUE)) {
	req$postBody <- paste0(req$rook.input$read_lines(), collapse = "\n")
}
req$rawPostBody <- body # req$rook.input$read() # this object
```

R/post-parsers.R
```diff
@@ -0,0 +1,183 @@
...
+    .globals$parsers <- list()
+  }
+  if (!is.null(.globals$parsers$f[[name]])) {
+    stop("Already have a parser by the name of ", name)
+  }
+  if (is.null(pattern)) {
+    pattern <- paste0("application/", name)
+  }
+  .globals$parsers$f[[name]] <- parser
+  .globals$parsers$p[[name]] <- pattern
```
schloerke, line 60: rename `f` and `p` to _full names_

R/post-parsers.R
```diff
@@ -0,0 +1,183 @@
...
+}
+
+
+
+
+#' OCTET
+#' @rdname parsers
+#' @param ... Raw values and headers are passed there.
+#' @export
+parser_octet <- function(...) {
```
schloerke, line 154: return raw by default. Allow for a option / parameter to be set to return the file name (of an uncleared file).

meztez: Not sure about the option name, any suggestion welcome

R/swagger.R
```diff
@@ -4,15 +4,34 @@
...
+        location = location,
+        arraySupport = arraySupport,
+        realType = realType
       )
 
+    if (arraySupport == TRUE) {
+      swaggerTypeInfo[[swaggerType]] <<- modifyList(
+        swaggerTypeInfo[[swaggerType]],
+        list(regexArray = paste0("(?:(?:", regex, "),?)+"),
+             converterArray = function(x) {converter(stri_split_fixed(x, ",")[[1]])})
```
schloerke, line 32: I don't know if this will handle `c("A,a", "B")`. 

I don't know of a clean solution. Will take some digging, maybe in a second PR later.

meztez: In the path you mean?

meztez: `/test/A%2Ca,B/testy`

Something like that

schloerke: If the string value has a comma, but is an item array... can we determine that?

meztez: Which item are arrays is defined when the router is created. Array converter will only be used on predefined array path parameters.

If you want to use a coma in a value that is part of an array path parameter, you would have to URLEncode the coma first (%2C).

meztez: > Additionally, the allowReserved keyword specifies whether the reserved characters :/?#[]@!$&'()*+,;= in parameter values are allowed to be sent as they are, or should be percent-encoded. By default, allowReserved is false, and reserved characters are percent-encoded. For example, / is encoded as %2F (or %2f), so that the parameter value quotes/h2g2.txt will be sent as quotes%2Fh2g2.txt
https://swagger.io/docs/specification/serialization/

It is the same if you want a `/` in your string, you have to encode it.

schloerke: Should we be url decoding the values as well, then?

schloerke: Ah. For now, we will not `allowReserved`.

Thank you!

schloerke:
Note: Do we need to safe guard against special characters, such as `,`?

No. 

> Additionally, the allowReserved keyword specifies whether the reserved characters :/?#[]@!$&'()*+,;= in parameter values are allowed to be sent as they are, or should be percent-encoded. By default, allowReserved is false, and reserved characters are percent-encoded. For example, / is encoded as %2F (or %2f), so that the parameter value quotes/h2g2.txt will be sent as quotes%2Fh2g2.txt
https://swagger.io/docs/specification/serialization/

It is the same if you want a `/` in your string, you have to encode it.

See: https://github.com/rstudio/plumber/pull/532#discussion_r439584727

R/swagger.R
meztez: Probably? It would be in `extractPathParams`

schloerke, line 32:
```suggestion
             # Q: Do we need to safe guard against special characters, such as `,`?
             # https://github.com/rstudio/plumber/pull/532#discussion_r439584727
             # A: https://swagger.io/docs/specification/serialization/
             # > Additionally, the allowReserved keyword specifies whether the reserved 
             # > characters :/?#[]@!$&'()*+,;= in parameter values are allowed to be sent as they are, 
             # > or should be percent-encoded. By default, allowReserved is false, and reserved characters 
             # > are percent-encoded. For example, / is encoded as %2F (or %2f), so that the parameter 
             # > value quotes/h2g2.txt will be sent as quotes%2Fh2g2.txt
             converterArray = function(x) {converter(stri_split_fixed(x, ",")[[1]])})
```

schloerke: For now, I think we should not. 

Until we add an option to the route to declare that params should be decoded.

meztez: What tickles me abit is that query parameters are decoded but not class converted, path parameters are not decoded but are class converted. Seems like an behavior to let the user handle the conversion when the type is declared in the plumber file.

R/swagger.R
```diff
@@ -191,5 +346,58 @@ removeNaOrNulls <- function(x) {
...
+  }
+
   ret
 }
+
+#' For openapi definition
+#' @noRd
+priorizeProperty <- function(...) {
+  l <- list(...)
+  isnullordefault <- sapply(l, function(x) {isNaOrNull(x) || isTRUE(attributes(x)$default)})
```
schloerke, line 361: `sapply` is not a _consistent_ return shape function. Since we are looking for a vector, we can use `vapply`.

```suggestion
  isnullordefault <- vapply(l, function(x) {isNaOrNull(x) || isTRUE(attributes(x)$default)}, logical(1))
```

R/swagger.R
```diff
@@ -191,5 +346,58 @@ removeNaOrNulls <- function(x) {
...
+  l[[which.min(isnullordefault)]]
```
schloerke, line 362: I've been tripped up many times. `which.min(logical_vect)` does not return the first value. It returns the smallest valid location, which is `FALSE`, a falsey value.

```r
❯❯ sample(21:25) %>% print() %>% which.min()
[1] 25 23 22 24 21
[1] 5

❯❯ sample(c(T,T,F,F,F)) %>% print() %>% which.min()
[1]  TRUE FALSE FALSE  TRUE FALSE
[1] 2
# Not expected!

❯❯ sample(c(T,T,F,F,F)) %>% print() %>% which() %>% min()
[1] FALSE FALSE FALSE  TRUE  TRUE
[1] 4
# Expected!
```

```suggestion
  l[[min(which(isnullordefault))]]
```

schloerke: Are we guaranteed that there is a valid value?  I don't want `which()` to return a `numeric(0)`

R/swagger.R
```diff
@@ -191,5 +346,58 @@ removeNaOrNulls <- function(x) {
...
+  testJSONserializable
+}
+
+#' Extract metadata on args of plumberExpression
+#' @noRd
+getArgsMetadata <- function(plumberExpression){
+  #return same format as getTypedParams or params?
+  args <- formals(eval(plumberExpression))
+  lapply(args[!names(args) %in% c("...", "res", "req")], function(arg) {
+    required <- identical(arg, formals(function(x){})$x)
```
schloerke, line 383:
```suggestion
  args <- formals(eval(plumberExpression))
  required_arg <- formals(function(x){})$x
  lapply(args[!names(args) %in% c("...", "res", "req")], function(arg) {
    required <- identical(arg, required_arg)
```

R/swagger.R
meztez: > ```r
> sample(c(T,T,F,F,F)) %>% print() %>% which() %>% min()
> ```
As I understand it FALSE is 0 and TRUE is 1, which.min will return first FALSE, then TRUE.
For this particular function, it needs to return like your sample 2

meztez: To make it clear, we are looking for the first value in the list where the value from the expression above is FALSE.

`min(which())` is going to return the first position of a TRUE value, which is not what this function is looking for.

I'll attempt a simpler prioritizer.

meztez: > 
> 
> Are we guaranteed that there is a valid value? I don't want `which()` to return a `numeric(0)`

It can if the function is called without parameters. Will fix.

R/parse-block.R
```diff
@@ -166,39 +167,30 @@ parseBlock <- function(lineNum, file){
...
-      if (!is.na(nameType[1,4])){
-        reqd <- nameType[1,4] == "*"
+      if (is.na(name)){
+        stopOnLine(lineNum, line, "No parameter specified.")
       }
-      params[[name]] <- list(desc=paramMat[1,5], type=type, required=reqd)
+      type <- stri_replace_all(paramMat[1,4], "$1", regex = "^\\[([^\\]]*)\\]$")
+      type <- plumberToSwaggerType(type)
+      isArray <- stri_detect_regex(paramMat[1,4], "^\\[[^\\]]*\\]$")
+      isArray <- isArray & supportsArray(type)
```
cpsievert, line 186:
```suggestion
      isArray <- isArray && supportsArray(type)
```

R/swagger.R
```diff
@@ -191,5 +346,63 @@ removeNaOrNulls <- function(x) {
...
+             testJSONserializable <<- FALSE}
+  )
+  testJSONserializable
+}
+
+#' Extract metadata on args of plumberExpression
+#' @noRd
+getArgsMetadata <- function(plumberExpression){
+  #return same format as getTypedParams or params?
+  args <- formals(eval(plumberExpression))
```
cpsievert, line 385: `plumberExpression` should already be a function at this point, perhaps just add a check for it instead of `eval()` ing

R/swagger.R
```diff
@@ -191,5 +346,63 @@ removeNaOrNulls <- function(x) {
...
+  required_arg <- formals(function(x){})$x
+  lapply(args[!names(args) %in% c("...", "res", "req")], function(arg) {
+    required <- identical(arg, required_arg)
```
cpsievert, line 388: Consider doing `rlang::is_missing(arg)`

cpsievert:
```
f <- function(a, b = 1){}
rlang::is_missing(formals(f)$a)
#> [1] TRUE
```

R/swagger.R
```diff
@@ -191,5 +346,63 @@ removeNaOrNulls <- function(x) {
...
+    if (is.call(arg) || is.name(arg)) {
+      arg <- tryCatch(
+        eval(arg),
```
cpsievert, line 391: This likely needs to be evaluate in the environment of the function

R/swagger.R
```diff
@@ -191,5 +346,63 @@ removeNaOrNulls <- function(x) {
...
+        error = function(cond) {NA})
+    }
+    if (!is.logical(arg) && !is.numeric(arg) && !is.character(arg)
+        && !(is.list(arg) && isJSONserializable(arg))) {
+      arg <- NA
+    }
```
cpsievert, line 397: Add a comment about why this `if` statement is here and maybe throw a warning/message in the `else` case

R/query-string.R
```diff
@@ -11,116 +11,131 @@ queryStringFilter <- function(req){
...
       )
     )
   }
-  if (length(types) > 0) {
-    types[is.na(types)] <- defaultSwaggerType
+
+  plumberTypes <- stri_replace_all(match[,3], "$1", regex = "^\\[([^\\]]*)\\]$")
+  if (length(funcParams) > 0) {
+    # Override with detection of function args if type not found in map
+    idx <- !(types %in% names(plumberToSwaggerTypeMap))
```
cpsievert, line 79:
```suggestion
    idx <- !(plumberTypes %in% names(plumberToSwaggerTypeMap))
```

R/swagger.R
meztez: Thanks a lot, I suspected there was an obivous to check that.

R/plumber-step.R
```diff
@@ -190,6 +191,8 @@ PlumberEndpoint <- R6Class(
       }
       private$envir <- envir
 
+      private$regex <- createPathRegex(path, self$getFuncParams())
```
cpsievert, line 194: @schloerke to look into removing this function completely and replace with known parsed structure from `parseBlock()`

R/swagger.R
meztez: Hopefully not as it is supposed to capture arguments value at the call level before the function is executed.

Say:
```r
avalue <- c(1,2)
function(a = avalue) {
 something
}
```

It should eval `a` to `c(1,2)`

R/post-body.R
```diff
@@ -1,34 +1,52 @@
...
-    charset <- getCharacterSet(req$HTTP_CONTENT_TYPE)
-    args <- parseBody(body, charset)
-    req$postBody <- body
+  if (is.null(handled) || handled != TRUE) {
+    # This will return raw bytes
+    body <- req$rook.input$read()
+    type <- req$HTTP_CONTENT_TYPE
+    args <- parseBody(body, type)
     req$args <- c(req$args, args)
+    req$rawPostBody <- body
```
cpsievert, line 9:
```suggestion
    req$postBodyRaw <- body
```

R/post-body.R
```diff
@@ -1,34 +1,52 @@
...
+    if (getOption("plumber.postBody", TRUE)) {
```
cpsievert, line 10: Use `isTRUE()`

R/post-body.R
```diff
@@ -1,34 +1,52 @@
...
-  # Is it JSON data?
-  if (stri_startswith_fixed(body, "{")) {
-    ret <- safeFromJSON(body)
+parserPicker <- function(content_type, first_byte, filename = NULL) {
+  #fast default to json when first byte is 7b (ascii {)
+  if (first_byte == as.raw(123L))
+    return(.globals$parsers$func[["json"]])
+  if (is.null(content_type)) {
+    return(.globals$parsers$func[["query"]])
+  }
```
cpsievert, line 37: Consistency nit-pick, either use `{` or don't use it (prefer to use it)

R/post-body.R
meztez: Added. Note on performance, it adds about a microsecond of execution time. I don't think it is a problem.

R/post-parsers.R
```diff
@@ -0,0 +1,185 @@
...
+#' }
+#' ```
+#'
+#' It should return a named list if you want values to map to
+#' plumber endpoint function args.
+#'
+#' @examples
+#' parser_json <- function(...) {
+#'   function(value, content_type = "application/json", ...) {
+#'     charset <- getCharacterSet(content_type)
```
cpsievert, line 43: Consider exporting `getCharacterSet`

R/post-parsers.R
```diff
@@ -0,0 +1,185 @@
...
+}
+
+
+
+#' JSON
+#' @rdname parsers
+#' @param ... Raw values and headers are passed there.
+#' @export
+parser_json <- function(...) {
+  function(value, content_type = NULL, ...) {
```
cpsievert, line 72: Could have a more informative argument name than `value` (maybe `postBodyRaw`)?

R/post-parsers.R
```diff
@@ -0,0 +1,185 @@
...
+
+
+#" RDS
+#' @rdname parsers
+#' @export
+parser_rds <- function(...) {
+  function(value, filename, ...) {
+    tmp <- tempfile("plumb", fileext = paste0("_", basename(filename)))
+    writeBin(value, tmp)
+    on.exit(file.remove(tmp))
```
cpsievert, line 120:
```suggestion
    on.exit(file.remove(tmp), add = TRUE)
    writeBin(value, tmp)
```

R/post-parsers.R
```diff
@@ -0,0 +1,185 @@
...
+    list(readRDS(tmp))
```
cpsievert, line 121: Why the extra `list()`?

R/post-parsers.R
```diff
@@ -0,0 +1,185 @@
...
+    boundary <- stri_match_first_regex(content_type, "boundary=([^; ]{2,})", case_insensitive = TRUE)[,2]
+    toparse <- parse_multipart(value, boundary)
+    # content-type detection
+    for (i in seq_len(length(toparse))) {
+      if (!is.null(toparse[[i]]$filename)) {
+        ext <- tools::file_ext(toparse[[i]]$filename)
+        toparse[[i]]$content_type <- getContentType(ext)
+      }
+    }
+    lapply(toparse, parseRaw)
```
cpsievert, line 145:
```suggestion
    lapply(toparse, function(x) {
      if (!is.null(x$filename)) {
        x$content_type <- getContentType(tools::file_ext(x$filename))
      }
      parseRaw(x)
    })
```

R/post-parsers.R
meztez: So that plumber can match plumberExpression arguments eventually.

R/post-parsers.R
```diff
@@ -0,0 +1,185 @@
...
+
+
+#' OCTET
+#' @rdname parsers
+#' @param ... Raw values and headers are passed there.
+#' @export
+parser_octet <- function(...) {
+  function(value, filename, ...) {
+    if (!missing(filename) && getOption("plumber.saveFileToDisk", FALSE)) {
+      if (interactive()) {
```
cpsievert, line 158: Seems weird to have different behavior here if interactive (why not always use a tempfile?)

R/post-parsers.R
```diff
@@ -0,0 +1,185 @@
...
+
+
+
+#' OCTET
+#' @rdname parsers
+#' @param ... Raw values and headers are passed there.
+#' @export
+parser_octet <- function(...) {
+  function(value, filename, ...) {
+    if (!missing(filename) && getOption("plumber.saveFileToDisk", FALSE)) {
```
cpsievert, line 157:
```suggestion
  function(value, filename = NULL, ...) {
    if (!is.null(filename) && getOption("plumber.saveFileToDisk", FALSE)) {
```

R/post-parsers.R
meztez: Mostly through internal use, people did not like having to fetch in their temp folder when they tested their API locally. I don't like it either. I'm very ok with having a consistent behavior.

R/post-parsers.R
cpsievert: Could you provide a bit more context (maybe w/ an example)?

R/post-parsers.R
```diff
@@ -0,0 +1,184 @@
...
+      } else {
+        tmp <- tempfile("plumb", fileext = paste0("_", basename(filename)))
+        writeBin(value, tmp)
+        ret <- tmp
+        attr(ret, "filename") <- filename
+      }
+      return(ret)
+    } else {
+      return(value)
+    }
```
cpsievert, line 171:
```suggestion
    attr(value, "filename") <- filename
    value
```

R/query-string.R
```diff
@@ -11,116 +11,131 @@ queryStringFilter <- function(req){
...
   }
-  if (stri_startswith_fixed(qs, "?")) {
-    qs <- substr(qs, 2, nchar(qs))
-  }
 
-  parts <- strsplit(qs, "&", fixed = TRUE)[[1]]
-  kv <- strsplit(parts, "=", fixed = TRUE)
-  kv <- kv[vapply(kv, length, numeric(1)) == 2] # Ignore incompletes
+  qs <- stri_replace_first_regex(qs, "^[?]", "")
+  #qs <- stri_replace_all_fixed(qs, "+", " ")
```
cpsievert, line 20:
```suggestion
```

R/post-parsers.R
meztez: reusing the argument name from `webutils::parse_multipart`. Otherwise it breaks `do.call(parser(), toparse)`. Or should I rename ouputs from `webutils::parse_multipart`.

Let me know and I'll change them.

R/query-string.R
```diff
@@ -11,116 +11,131 @@ queryStringFilter <- function(req){
...
+
+  args <- stri_split_fixed(qs, "&", omit_empty = TRUE)[[1L]]
+  kv <- lapply(args, function(x) {
+    # returns utf8 strings
+    httpuv::decodeURIComponent(stri_split_fixed(x, "=", omit_empty = TRUE)[[1]])
+  })
+  kv <- kv[lengths(kv) == 2] # Ignore incompletes
```
cpsievert, line 27:
```suggestion
  kv <- kv[vapply(kv, length, numeric(1)) == 2] # Ignore incompletes
```

cpsievert: lengths was added in R 3.2.0 ...we currently support 3.0 and above

R/query-string.R
```diff
@@ -11,116 +11,131 @@ queryStringFilter <- function(req){
...
 
   if (length(kv) == 0) {
     # return a blank list of args if there is nothing to parse
     return(list())
   }
 
-  keys <- httpuv::decodeURIComponent(vapply(kv, "[[", character(1), 1)) # returns utf8 strings
-  if (any(Encoding(keys) != "unknown")) {
+  k <- vapply(kv, "[", character(1), 1)
```
cpsievert, line 34:
```suggestion
  k <- vapply(kv, `[`, character(1), 1)
```

R/query-string.R
```diff
@@ -11,116 +11,131 @@ queryStringFilter <- function(req){
...
+  kenc <- unique(Encoding(k))
```
cpsievert, line 35: How come we are no longer decoding the URI?

R/query-string.R
meztez: I left it there in case we wanted to consider `+` query as a space character.

R/query-string.R
meztez: It is done on line 25 above,

R/query-string.R
```diff
@@ -11,116 +11,130 @@ queryStringFilter <- function(req){
...
-
-  ret <- as.list(vals)
-  names(ret) <- keys
-
+  v <- lapply(kv, "[", 2)
   # If duplicates, combine
-  combine_elements <- function(name){
-    unname(unlist(ret[names(ret)==name]))
-  }
+  v <- sapply(unique(k), function(x) do.call(c, v[x == k]), simplify = FALSE)
```
cpsievert, line 46:
```suggestion
  v <- lapply(unique(k), function(x) do.call(c, v[k %in% x])) 
  names(v) <- k
```

R/query-string.R
```diff
@@ -11,116 +11,130 @@ queryStringFilter <- function(req){
...
     )
   }
 
-  vals <- vapply(kv, "[[", character(1), 2)
-  vals <- httpuv::decodeURIComponent(vals) # returns utf8 strings
-
-  ret <- as.list(vals)
-  names(ret) <- keys
-
+  v <- lapply(kv, "[", 2)
```
cpsievert, line 44: Just realized a lot of the the implementation above could be replaced by `webutils::parse_query()`

meztez: Yeah, the only part missing is handling parameters with the same name:
https://github.com/jeroen/webutils/issues/4

R/plumber-step.R
cpsievert: Be aware, this is also relevant to swagger changes

R/swagger.R
meztez: Reverted missing does not seems to be storable in `required_arg`, leaving it as it is.

R/swagger.R
meztez: Reverted, rlang cause import problem with crayon plus another deb

R/swagger.R
```diff
@@ -124,54 +200,125 @@ extractResponses <- function(resps){
...
-    location <- "query"
     if (p %in% pathParams$name) {
       location <- "path"
+      required <- TRUE
+      style <- "simple"
+      explode <- FALSE
+      type <- priorizeProperty(defaultSwaggerType,
+                               pathParams[pathParams$name == p,]$type,
+                               endpointParams[[p]]$type,
+                               funcParams[[p]]$type)
```
cpsievert, line 234: Could store this info at the start of the loop and then refer in multiple places

R/swagger.R
meztez: added envir

R/swagger.R
```diff
@@ -191,5 +346,69 @@ removeNaOrNulls <- function(x) {
...
+
+#' Extract metadata on args of plumberExpression
+#' @noRd
+getArgsMetadata <- function(plumberExpression){
+  #return same format as getTypedParams or params?
+  if (!is.function(plumberExpression)) plumberExpression <- eval(plumberExpression)
+  args <- formals(plumberExpression)
+  lapply(args[!names(args) %in% c("...", "res", "req")], function(arg) {
+    required <- identical(arg, formals(function(x){})$x)
+    if (is.call(arg) || is.name(arg)) {
```
cpsievert, line 389: Maybe we should always be always be `eval()`ing

tests/testthat/test-path-subst.R
```diff
@@ -78,3 +110,43 @@ test_that("integration of path parsing works", {
                list(error = "404 - Resource Not Found"))
   expect_equal(r$route(make_req("GET", "/car/sold/true"), PlumberResponse$new()), TRUE)
 })
+
+test_that("multiple variations in path works nicely with function args detection", {
+
+  # Check when detection is not provided
+  pathDef <- "/<var0:str>/<var1:chr*>/<var2:[int]>/<var3>/<var4:*>/<var5:[*>/<var6:[]*>/<var7:[]>/<var8*>"
+  regex <- suppressWarnings(createPathRegex(pathDef))
```
cpsievert, line 118: Instead of `suppressWarnings()` here and below, please use `expect_warning()` with a proper `regexp`

R/swagger.R
meztez: This is done one time when plumber is first ran. Any reason beside performance? This has to be done for each parameter individually at the moment. PriorizeProperty was not made to handle all of them at once.

cpsievert: Just performance, not a blocking issue

R/swagger.R
meztez: Could you detail your reasoning, I'm confused since your stated otherwise above. Anything we need to aware of?

cpsievert: Mainly just to make sure we're not missing a case of some un-evaluated object (I can't think of one at the moment, though).

It also makes it a bit more clear what the intention is (that is, give the evaluated argument value)

R/query-string.R
meztez: This does not produce the same results, I'll try to fix it.

schloerke:
I'm off for the weekend. I'll look into it on Monday. Enjoy your weekend!

R/post-parsers.R
meztez: Well for the life of me I can't remember why I needed to add that and it seems to work fine without it. Removing it for now.

R/post-parsers.R
meztez: Removed interactive differenciation, added attribute.

R/plumber-step.R
schloerke: This is the parseBlock code path.  The only time the path is parsed is in `createPathRegex()`.

Resolving comment

JosiahParry:
Super excited about multipart file upload! Is this documented? I cannot find it in `rplumber.io`.

meztez:
No documented yet on rplumber.io. I believe the goal is to use pkgdown to build plumber documenter. There is another branch in development for that.

I'm super pump too.

ThHarbig:
Tried it out, works nicely! Is it possible to do this with an arbitrary number of files? I'm not talking about having multiple file parameters but one multi-file parameter. Thanks!

ThHarbig:
I tried it, works great! Is there a way to upload an arbitrary number of files at the same time?

meztez:
It would probably work via curl, I would mostly have to adapt the swagger spec generation. Please open a different issue so I can reference it in a PR.

chrischung28:
This is a great addition! Thank you very much. Just wondering if it would be possible to get an example of the file upload feature? I'm having some trouble getting it to work with the example above when i try to upload some pdfs.

meztez:
In the implementation that made it to master, you get a vector of raw bytes to play with. Here is a working example with the current master.

```R
#* @apiTitle Upload test

#* @param pdf:file
#* @post /upload
function(pdf) {
  f <- tempfile()
  on.exit(unlink(f), add = TRUE)
  writeBin(as.raw(pdf), f, useBytes = TRUE)
  finfo <- file.info(f)
  finfo$mode <- as.character(finfo$mode)
  finfo$real_filename <- attr(pdf, "filename")
  finfo
}
```

chrischung28:
@meztez Thank you very much! This works great.

s-fleck:
@meztez sorry to bug you, but could you provide an example how to use this with `httr::POST()`? It is not clear to me where the `"filename"` attributes comes from. Is it possible with this approach to have an endpoint where I can upload with `curl -F` **and** `httr::POST()` without jumping through too many hoops **and have the filename available in plumber**?

meztez:
yeah, the filename comes from the multipart form-data.

```R
plumber::options_plumber(port = 8004)

#* @param a_file:file
#* @post /upload
function(a_file) {
  # Plumber a named list. The names correspond
  # to the files names. The content of the list
  # will depend on the content-type set
  # for each files.
  return(name_of_file = list(names(a_file), file_content_length = nchar(a_file[[1]])))
}
```

with httr::POST
```R
f <- tempfile("stempy")
writeLines(stringi::stri_rand_strings(1, 100), f)
res <- httr::POST("http://127.0.0.1:8004/upload", 
                  body = list(a_file = httr::upload_file(f)),
                  encode = "multipart")
httr::content(res)
```

Should be similar with curl, files is specific to my system

```
curl -F a_file=@bobettebo.txt http://127.0.0.1:8004/upload
```

s-fleck:
thanks! i didn't know about `httr::upload_file()`, i was always sending raw binary data

PR #536, opened 2020-05-17
Work in progress, combined PR

meztez:
I combined all the work done into one PR, it was getting hard to track all work across different branches.

This is still a work in progress. 

### What this branch contains:
* [x] document options with roxygen like shiny
* [x] ignore req, res, data in function args detection
* [x] parser yaml / serializer
* [x] redoc
* [x] mount openapi.yaml
* [x] Distinguish between 404 and 405 http error
* [x] Distinguish between openapi and swagger in objects names
* [x] Set body max size with getOption('plumber.maxRequestSize', 0)
* [x] rename some files to better match the purpose of their content
* [x] Add yaml dep in case we want to suppoer fromOpenAPI yaml format
* [x] Modular body parsing, more like serializers
* [x] R6 Documentation
* [x] toOpenAPI from a file or plumber router
* [x] Support array parameters, files, object as plumber type
* [x] Fix OpenAPI api url when referer from same domain (empty)
* [x] Fix tests on non UTF-8 platform / Windows
* [x] Add libsodium-dev dep
* [x] Switch to xenial-cran35, trusty-cran35
* [x] Fix doc / imports to avoid :: on moderatly used functions
* [x] Perf improvement getCharacterSet
* [x] Fix isNaOrNULL when x is c(NULL, value)
* [x] Add expression args detection to openAPI specs
* [x] Fix image serializer execution position
* [x] Switch to remotes instead of devtools
* [x] Remove non-test output from test outputs

### Still to be done:
* [ ] http error generalization ??
* [ ] fromOpenAPI
* [ ] modular plumb methods for plumbing files / custom tags
* [ ] parser text/csv / serializer
* [ ] Proof read docs once development stabilizes

### Breaking:
swagger/swaggerCallback in run method of PlumberRouter  :
 - Now there is a `ui` and `callback` param
 - `ui` can be TRUE, FALSE or character vector
 - It supports "Swagger" and "Redoc" (meztez/redoc, CRAN pending)
 - `callback` has the same function as before (RStudio)
 - PlumberRouter gains customSpec field for Custom OpenAPI (expect named list)
 - Removed deprecated code for major version release 5.0.0
 - Removed requestBody from req env

Why I broke stuff : I felt it was needed for 5.0.0 to make a clear split between the UIs (Swagger, Redoc) and OpenAPI specifications. You don't need to run a router to get the specs with your custom definitions in this PR. I also think the run method had too much clutter.
Custom definitions are applied using `utils::modifyList` and overwrite default plumber spec. This is a bit different than prior approach.

### Open questions to maintainers:

Body parsing
* Which dep? mimes, webutils, or write multipart parser?
* When picking a body parser, what should happen when picker finds multiple matches? Currently uses the first indexed one
* Normally nothing should read rook.input before postBodyFilter, rewind to be safe if change are made elsewhere?
* Is it memory wise to store the entire body in req$requestBody when it contains whole files

Decide what to do with files sent to API
* Should plumber let the user delete them?
* Handling maxfilesize limit, with option ok?
* Should it even write file sent to disk?
* What should be the default max request body size (0 Unlimited?)

Misc
* Get rid of Legacy methods in Plumber R6 object?
* in R6 Class PlumberStep, should getPathParams, getTypedParams and getFunctParams return all in the format? Maybe a name change in methods here could help distinguish between getTypedParams and getPathParams?
* Bump to R 4.0.0 already in droplet?


R/openapi-types.R
```diff
@@ -0,0 +1,120 @@
...
+    invisible(TRUE)
+  }
+
+  addDataTypeInfo(
+    "boolean",
+    c("bool", "boolean", "logical"),
+    "[01tfTF]|true|false|TRUE|FALSE",
+    as.logical,
+    "(?:(?:[01tfTF]|true|false|TRUE|FALSE),?)+",
+    function(x) {as.logical(stri_split_fixed(x, ",")[[1]])},
```
schloerke, line 40: Do not define the regex twice. 
Do not define this function split function.

Move the generation of these two objects into addDataTypeInfo and store them as different args. `regexArray`, `converterArray`

R/ui.R
```diff
@@ -0,0 +1,134 @@
...
+      changeHostLocation = TRUE
+    )
+  )
+
+  # Mount openAPI spec paths openapi.json and openapi.yaml
+  mountOpenAPI(pr, api_url)
+
+  # Mount UIs
+  if (isTRUE(ui)) ui <- getOption("plumber.ui", "Swagger")
+  for (interface in ui) {
```
schloerke, line 22: Let's remove the for-loop.

`ui` should resolve to a single value

R/ui.R
```diff
@@ -0,0 +1,134 @@
...
+  for (path in c("/__redoc__/index.html", "/__redoc__/")) {
+    pr$handle(
+      "GET", path, redoc_index,
+      serializer = serializer_html()
+    )
+  }
+  pr$mount("/__redoc__", PlumberStatic$new(redoc::redoc_path()))
+
+  return(redocUrl)
+}
```
schloerke, line 131: Let's move this to the `redoc` package.

R/ui.R
```diff
@@ -0,0 +1,134 @@
...
+    )
+  }
+  pr$mount("/__swagger__", PlumberStatic$new(swagger::swagger_path()))
+
+  return(swaggerUrl)
+}
+
+#' @include globals.R
+.globals$interfaces[["swagger"]] <- mountSwagger
+.globals$interfaces[["Swagger"]] <- mountSwagger
```
schloerke, line 106: Will need to move this to the swagger package (can happen after the merge)

R/plumber-step.R
```diff
@@ -120,26 +136,60 @@ PlumberEndpoint <- R6Class(
...
+    #' @description ability to serve request
+    #' @param req a request object
+    #' @param res a response object
+    #' @return a logical. `TRUE` when endpoint can serve request.
+    canServe = function(req, res){
+      if (!is.na(stri_match_first_regex(req$PATH_INFO, private$regex$regex)[1,1])) {
+        if (req$REQUEST_METHOD %in% self$verbs) {
+          return(TRUE)
+        } else {
+          res$status = 405L
```
schloerke, line 169: Add comment about what this is doing / why we are altering the `res` object

R/plumber-step.R
```diff
@@ -120,26 +136,60 @@ PlumberEndpoint <- R6Class(
...
+          res$setHeader("Allow", paste(self$verbs, collapse = ", "))
```
schloerke, line 170: This logic have to be done after all routes have failed to serve.

Ex:
```
/home POST
/home GET
```

This would end up producing an `Allow` header on a GET post.

schloerke:
@meztez Do you mind remerging the `master` branch? (Sometime before this Monday)
Thank you!

meztez:
I was thinking about splitting this one in smaller chunk since you merge #532 if it is ok with you. yes before monday.

schloerke:
Even better! 🙌

R/openapi-types.R
```diff
@@ -0,0 +1,132 @@
+# calculate all OpenAPI type information at once and use created information throughout package
+dataTypesInfo <- list()
+dataTypesMap <- list()
```
schloerke, line 3: May we use a more specific name? Maybe `apiTypesInfo`? (`sub("data", "api", .)`)

It's a lot of name changes, but it helps when reading to supply context

R/plumber.R
```diff
@@ -246,59 +243,39 @@ plumber <- R6Class(
...
+    #' UI will be served on path `/__{ui}__/index.html` and `/__{ui}__/`.
     run = function(
       host = '127.0.0.1',
       port = getOption('plumber.port'),
-      swagger = interactive(),
+      ui = interactive(),
       debug = interactive(),
-      swaggerCallback = getOption('plumber.swagger.url', NULL)
+      callback = getOption('plumber.ui.callback', getOption('plumber.swagger.url', NULL)),
+      ...
```
schloerke, line 275: Can we change the order?  And set `ui` to TRUE.

```r
..., 
ui = TRUE,
debug = interactive(),
callback = ....
```

R/openapi-types.R
meztez: I was in the process a decoupling the mountui part and de renaming, reorganizing part.

Also wondering what would be a good name for dataType.

It refers to OpenAPI data types. If you think apiTypes would work, I'm down for it.

Don't worry about replacing stuff. Naming things is the hardest part.

R/plumber.R
```diff
@@ -890,29 +802,30 @@ plumber <- R6Class(
...
-
       ret
     },
-    #' @description Retrieve openAPI file
-    openAPIFile = function() {
-      self$swaggerFile()
-    },
+    #' @field customSpec A list of custom spec to overlay over openAPI spec
+    #' generated from a plumber router.
+    customSpec = NULL,
```
schloerke, line 818: Rather than using a static object, I'd like to supply a function directly.

```r
#### in the private defs
    apiHandler = force,
```

```r
    #' @description Add a function to customize what is returned in `$apiSpec()`. 
    #' @param api_fun This function should accept an OpenAPI Specification autogenerated by `plumber`.  It should return a list object in valid OpenAPI Specification format. This will not be validated.
    setApiHandler = function(api_fun = force) {
      # allow null values
      if (!is.null(api_fun)) {
        stopifnot(is.function(api_fun))
      }
      private$apiHandler <- api_fun
    }
```

R/plumber.R
```diff
@@ -890,29 +802,30 @@ plumber <- R6Class(
...
-
-      swaggerPaths <- private$swaggerFileWalkMountsInternal(self)
-
-      # Extend the previously parsed settings with the endpoints
-      def <- utils::modifyList(private$globalSettings, list(paths = swaggerPaths))
-
-      # Lay those over the default globals so we ensure that the required fields
-      # (like API version) are satisfied.
+    #' @description Retrieve OpenAPI specification
+    openAPISpec = function() {
```
schloerke, line 806: I think it's a little shorter and less confusing with the caps

```suggestion
    apiSpec = function() {
```

In the documentation, we can always refer to it as the `OpenAPI Specification function`

R/plumber.R
```diff
@@ -890,29 +802,30 @@ plumber <- R6Class(
...
-    swaggerFile = function() { #FIXME: test
-
-      swaggerPaths <- private$swaggerFileWalkMountsInternal(self)
-
-      # Extend the previously parsed settings with the endpoints
-      def <- utils::modifyList(private$globalSettings, list(paths = swaggerPaths))
-
-      # Lay those over the default globals so we ensure that the required fields
-      # (like API version) are satisfied.
+    #' @description Retrieve OpenAPI specification
```
schloerke, line 805: [They seem to like](https://spec.openapis.org/oas/v3.0.3) `Specification` to be capitalized

> The OpenAPI Specification (OAS) defines ....


```suggestion
    #' @description Retrieve OpenAPI Specification
```

R/plumber.R
```diff
@@ -890,29 +802,30 @@ plumber <- R6Class(
...
-      self$swaggerFile()
-    },
+    #' @field customSpec A list of custom spec to overlay over openAPI spec
+    #' generated from a plumber router.
+    customSpec = NULL,
 
     ### Legacy/Deprecated
+    #' @description Retrieve OpenAPI specification (deprecated and will be removed in a coming release)
+    openAPIFile = function() {
+      self$openAPISpec()
```
schloerke, line 823:
```suggestion
      warning("`$swaggerFile()` has been deprecated in v1.0.0 and will be removed in a coming release. Please use `$apiSpec()`.")
      self$openAPISpec()
```

R/plumber.R
```diff
@@ -890,29 +802,30 @@ plumber <- R6Class(
...
+    },
+    #' @description Retrieve OpenAPI specification (deprecated and will be removed in a coming release)
+    swaggerFile = function() {
```
schloerke, line 826:
```suggestion
    warning("`$swaggerFile()` has been deprecated in v1.0.0 and will be removed in a coming release. Please use `$apiSpec()`.")
    swaggerFile = function() {
```

R/plumber.R
```diff
@@ -1071,7 +985,7 @@ plumber <- R6Class(
       private$ends[[preempt]] <- c(private$ends[[preempt]], ep)
     },
 
-    swaggerFileWalkMountsInternal = function(router, parentPath = "") {
+    routerSpecificationInternal = function(router, parentPath = "") {
```
schloerke, line 988: Thank you for renaming these variables! 😃

R/openapi-types.R
schloerke: > Also wondering what would be a good name for dataType.

`apiType` helps a lot. It lets me know that it's prolly a string value describing the OpenAPI Spec type.

`dataType` feels like an output of `typeof()` or something unrelated to OpenAPI

> I was in the process a decoupling the mountui part and de renaming, reorganizing part.

Great! Feel free to make those changes there. 😄

schloerke: I am hoping to refer to all things `OpenAPI` by using `api` in the variable name.

(Things with `plumber` are `pr` or `router`.)

meztez:
By adding #562, this one can die in peace. Goodbye shemera PR.

Issue #417, opened 2019-04-27
Swagger Support


NavinCJ:
Does the current annotation support swagger specs fully?  For example I wanted to have a POST API with name value in form-data whose value is of type 'File'.  If annotations do not support, is it possible for me to prepare the swagger json/yml separately and pass it as argument when I launch the router?

schloerke:
> Current annotation support full swagger spec?

The GitHub version currently allows for users to add their own spec information.  plumber does not currently validate this spec, so please return a valid swagger spec object.    

To get the latest: `remotes::install_github("trestletech/plumber")`

"Add you own spec" example: (changes title to the current time) 
https://github.com/trestletech/plumber/blob/796e8fe9cdad22b76b63f8f7bd48a5605a243817/scripts/manual_testing.R#L14-L20

If elements are required to be an array, I suggest using `list()`, rather than a vector.

schloerke:
* I do not see `plumber` ever supporting a completely full spec through `plumber` tags alone.  
* `plumber` is open to adding new tags that are simple / concise.  

Please see #390 as an example of an error caused a vector of size 1 when it should be treated as a list.

lorenzwalthert:
Pretty interesting. Two comments on that: 

When I tried to run the above code with the devel version of swagger, I get an error: 

```r
library(plumber)
pr <- plumber$new()
pr$handle("GET", "/:path/here", function(){})

pr$run(
  port = 1234,
  swagger = function(pr_, spec, ...) {
    spec$info$title <- Sys.time()
    spec
  }
)
Running plumber API at http://127.0.0.1:1234
Running Swagger UI  at http://127.0.0.1:1234/__swagger__/
<simpleError: 'swagger_spec' is not an exported object from 'namespace:swagger'>
<simpleError: 'swagger_spec' is not an exported object from 'namespace:swagger'>
```
When I use the devel version of rstudio/swagger, then I don't get an error. 

So maybe specify a minimal version requirement on swagger? 

But when I want to click on the json link right below the API title (http://127.0.0.1:1234/openapi.json in my case), I get this error: 
```
<simpleError in if (any(toRemove)) {    x[toRemove] <- NULL}: missing value where TRUE/FALSE needed>
```

This also happens when I don't specify the `swagger` argument, so maybe it is unrelated to this thread.

> I do not see plumber ever supporting a completely full spec through plumber tags alone.

Makes sense to me. What about a mixed approach? With `purrr::list_modify()`, we one could:
[...]

PR #550, opened 2020-06-16
Swaggerfile support for array parameters, object, files. Followup to #532

### LINKED ISSUE #187: Swagger parameters listed in reverse order

### LINKED ISSUE #320: Support definition of custom parsers
Much like defining customer serializers it would be helpful to be able to define custom parsers for `req$postBody`.

### LINKED ISSUE #463: Error in do.call(fi$exec, req$args): variable names are limited to 10000 bytes
### System details

Output of `sessioninfo::session_info()`:
```
Session info -------------------------------------------------------------------
 setting  value                       
 version  R version 3.6.0 (2019-04-26)
 os       Windows 10 x64              
 system   x86_64, mingw32             
 ui       RStudio                     
 language (EN)                        
 collate  Polish_Poland.1250          
 ctype    Polish_Poland.1250          
 tz       Europe/Warsaw               
 date     2019-07-19                  

- Packages -----------------------------------------------------------------------
 ! package     * version  date       lib source                              
   assertthat    0.2.1    2019-03-21 [1] CRAN (R 3.6.0)                      
   base64enc   * 0.1-3    2015-07-28 [1] CRAN (R 3.6.0)                      
   cli           1.1.0    2019-03-19 [1] CRAN (R 3.6.0)                      
   crayon        1.3.4    2017-09-16 [1] CRAN (R 3.6.0)                      
   fftw          1.0-5    2019-03-04 [1] CRAN (R 3.6.0)                      
   httpuv        1.5.1    2019-04-05 [1] CRAN (R 3.6.0)                      
   jsonlite      1.6      2018-12-07 [1] CRAN (R 3.6.0)                      
   later         0.8.0    2019-02-11 [1] CRAN (R 3.6.0)                      
   magrittr      1.5      2014-11-22 [1] CRAN (R 3.6.0)                      
   MASS          7.3-51.4 2019-03-31 [2] CRAN (R 3.6.0)                      
 V plumber       0.4.6    2019-07-19 [1] Github (trestletech/plumber@5a5223f)
   promises      1.0.1    2018-04-13 [1] CRAN (R 3.6.0)                      
   R6            2.4.0    2019-02-14 [1] CRAN (R 3.6.0)                      
   Rcpp          1.0.1    2019-03-17 [1] CRAN (R 3.6.0)                      
   remotes       2.1.0    2019-06-24 [1] CRAN (R 3.6.1)                      
   seewave       2.1.3    2019-03-30 [1] CRAN (R 3.6.1)                      
   sessioninfo   1.1.1    2018-11-05 [1] CRAN (R 3.6.1)                      
   signal        0.7-6    2015-07-30 [1] CRAN (R 3.6.1)                      
   stringi       1.4.3    2019-03-12 [1] CRAN (R 3.6.0)                      
   tuneR         1.3.3    2018-07-08 [1] CRAN (R 3.6.1)                      
   uuid        * 0.1-2    2015-07-28 [1] CRAN (R 3.6.0)                      
   withr         2.1.2    2018-03-15 [1] CRAN (R 3.6.1)                      
```
### Problem description
When I'm trying to send a large text body in POST request (~60kB), sometimes I get an error: `Error in do.call(fi$exec, req$args): variable names are limited to 10000 bytes`.  I found that the problem exists only when text body ends with two characters: `==` (I send big files as base64 text body). The Exception is thrown before entering my controller's code, so i thinkg this is an issue with Plumber. 

### How to reproduce
* Create simple controller:
```
#* @post /test
function (req, res) {
  print("TEST")
  res
}
```
* Send large text as body to that endpoint - example file is 
[here](https://github.com/trestletech/plumber/files/3411199/test.txt). To generate the error, simply add `==` at the end of text in that file.

Regards,
Radek

### LINKED ISSUE #469: Specify API Request Body JSON Schema Options and Documentation?
Hi @schloerke,

Opening this new issue, hope this is the correct place.

I'm trying to recreate something similar to the example in @blairj09 talk at RStudio::conf2019 earlier this year, where the expected request Schema is returned in the Swagger interface, his example is here:

https://github.com/sol-eng/plumber-model/blob/master/R/model-api/entrypoint.R
 
Based on this example, I would like to specify the expected JSON format for incoming requests. The above `entrypoint.R` file generates a section in the Swagger Interface under the **Example Value | Schema** that looks like this:
```
{
  "cyl": 0,
  "hp": 0
}
```

However I would like the example/scheme to actually run and need to specify an array of data, i.e. multiple rows in the array, which can be easily converted to a `data.frame` downstream. The Schema must look like this, with `[` and `]`:
```
[
  {
    "cyl": 0,
    "hp": 0
  }
]
```

... or even better (with actual values for 2 rows):

```
[
  {
    "cyl": 25.8,
    "hp": 33.6
  },
  {
    "cyl": 23.6,
    "hp": 22.9
  }
]
```

Is there a way in the `entrypoint.R` file to specify the square brackets? I have noticed that it is possible to add actual values to the Swagger interface example by specifying an `example` entry rather than a `properties` entry in the `schema` section of the call. Though I am still unable to figure out how to add `[` `]` to the example.

Some examples and/or documentation about what the options are in specifying the JSON scheme via the `entrypoint.R` file would be very helpful (it seems others also think so). If I've missed something obvious please let me know. Thank you!

### LINKED ISSUE #502: Sending raw/serialized data
Hi,

How can I send serialized data to the server?

I've seen the example to receive serialized data back (https://shrektan.com/post/2018/11/14/use-base-serializer-in-the-plumber-api/ provided by @shrektan) but I can't get the other way working.

Below the code. r_obj_serializer is the registered serializer from the above link. I receive the '20' back, but req is not what I sent in the body.

```
#* @post /api
#* @serializer r_obj_serializer
function(req) { 
  print(req$postBody)
  return(20)
}
```

```
post <- httr::POST(url = "http://localhost:8210/api",
                   body = base::serialize(list(myarg=1),NULL,FALSE),
                   encode='raw',
                   httr::content_type("application/octet-stream"))
```

Here the serializer.

```
plumber::addSerializer("r_obj_serializer", function() {
  function(val, req, res, errorHandler) {
    tryCatch({
      res$setHeader("Content-Type", "application/octet-stream")
      res$body <- base::serialize(val, NULL, ascii = FALSE)
      print('hi')
      return(res$toResponse())
    }, error = function(e) {
      errorHandler(req, res, e)
    })
  }
})
```

Context: I want to be able to send large serialized data objects to docker worker nodes. Using the json serialization takes too long.

Christian

### LINKED ISSUE #509: OpenAPI value should report if a variable is required
See https://github.com/rstudio/plumber/issues/417#issuecomment-584359202

cc @jeffkeller87

### LINKED ISSUE #512: Send data frame to post request in body

### LINKED ISSUE #516: Support arrays as query parameters

### LINKED ISSUE #523: parameter default values for Swagger

schloerke:
Fixes #187
Fixes #532

Fixes #516 
Fixes #512
Fixes #469
Fixes #509
Related #417
Fixes #523
Fixes #502
Fixes #463
Fixes #320
 

Incorporates all of #532
* Moved swagger info and parsers to be processed in `.onLoad()`
* Added a new "combine query params with same name" logic for larger nums of params
  * Added a test
* Added test for the root cause for #463
* Removed "save to disk" when files are posted
  * Q: Should this be kept? A: no


R/plumber.R
```diff
@@ -818,7 +819,24 @@ plumber <- R6Class(
     #' @param req request object
     #' @details required for httpuv interface
     onHeaders = function(req) {
-      NULL
+      maxSize <- private$maxSize
+      if (maxSize <= 0)
```
cpsievert, line 823:
```suggestion
      if (isTRUE(maxSize <= 0))
```

R/plumber.R
```diff
@@ -818,7 +819,24 @@ plumber <- R6Class(
...
+        return(NULL)
+
+      reqSize <- 0
+      if (length(req$CONTENT_LENGTH) > 0)
+        reqSize <- as.numeric(req$CONTENT_LENGTH)
+      else if (length(req$HTTP_TRANSFER_ENCODING) > 0)
+        reqSize <- Inf
+
+      if (reqSize > maxSize) {
```
cpsievert, line 832:
```suggestion
      if (isTRUE(reqSize > maxSize)) {
```

R/plumber.R
```diff
@@ -818,7 +819,24 @@ plumber <- R6Class(
...
     onHeaders = function(req) {
-      NULL
+      maxSize <- private$maxSize
+      if (maxSize <= 0)
+        return(NULL)
+
+      reqSize <- 0
+      if (length(req$CONTENT_LENGTH) > 0)
+        reqSize <- as.numeric(req$CONTENT_LENGTH)
+      else if (length(req$HTTP_TRANSFER_ENCODING) > 0)
```
cpsievert, line 829: I'm not sure I see why this case is here? Appears as though this header is no longer in HTTP/2, so should we really bother here? https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Transfer-Encoding

cpsievert: If we do go this route, can we document the fact that it should be `option('plumber.maxRequestSize' = Inf)` to accept requests in this case?

R/plumber.R
```diff
@@ -818,7 +819,24 @@ plumber <- R6Class(
...
+        reqSize <- Inf
```
schloerke, line 830:
```suggestion
      # https://github.com/rstudio/shiny/blob/a022a2b4/R/middleware.R#L298-L301
      if (length(req$CONTENT_LENGTH) > 0)
        reqSize <- as.numeric(req$CONTENT_LENGTH)
      else if (length(req$HTTP_TRANSFER_ENCODING) > 0)
        reqSize <- Inf
```

R/plumber.R
schloerke: Currently, shiny does not even document this

https://github.com/rstudio/shiny/blob/64e09315fc5f38a31e1bf26179296a97be30b0be/R/shiny-options.R#L73-L74

schloerke: Closing this convo here. Continue in #552

PR #554, opened 2020-06-19
Add 405 http code handling

### LINKED ISSUE #310: Expose `pr` object to filters
Much like `req` and `res` can be passed to filters and endpoints, it would be nice if the plumbed object (`pr`) was also able to be passed and accessed in filters.

One example use case for this is exposing the Swagger definition based on some filter logic. For instance, an authentication filter could check for auth and, if it's not there, return `pr$swaggerFile()` so that the Swagger UI can still be generated.

### LINKED ISSUE #493: Requests to valid endpoints with an incorrect request method should return HTTP 405 instead of HTTP 404
### Example application

Suppose we have the following API:

```R
srv <- plumber::plumber$new()
srv$handle("POST", "/submit", function(req, res) {
  # TODO: Actual work.
  res$headers[["Content-Type"]] <- "text/plain"
  res$body <- "Submitted"
  res
})
srv$run(host = "0.0.0.0")
```

If you send a `GET` request to this API (instead of the `POST` request it is expecting) you see the following (e.g. with `curl`):

```
$ curl -sv -X GET localhost:7312/submit
> GET /submit HTTP/1.1
> Host: localhost:7312
> User-Agent: curl/7.58.0
> Accept: */*
>
< HTTP/1.1 404 Not Found
< Date: Mon, 30 Sep 2019 22:15:12 GMT
< Content-Type: application/json
< Date: Mon, 30 Sep 2019 10:15:12 PM GMT
< Content-Length: 38
<
{"error":["404 - Resource Not Found"]}
```

There is a well-known and well-supported HTTP status code for the specific issue of "this URI is valid but the method is not", namely `405 Method Not Allowed`. It would be nice if Plumber could support returning this method instead of falling back to 404.

From a cursory look at how routing is implemented, this should be quite possible. It might require that users add a new error handler, though.

meztez:
Fixes #493 
Fixes #310

PR task list:
- [x] Update NEWS
- [x] Add tests


NEWS.md
```diff
@@ -42,6 +42,8 @@ plumber 0.5.0
 
 ### New features
 
+* Support HTTP 405 Code. (@meztez)
```
schloerke, line 45:
```suggestion
* Support HTTP 405 Code. (@meztez, #554)
```

R/plumber-step.R
```diff
@@ -163,7 +163,15 @@ PlumberEndpoint <- R6Class(
...
-      req$REQUEST_METHOD %in% self$verbs && !is.na(stri_match_first_regex(req$PATH_INFO, private$regex$regex)[1,1])
+      if (!is.na(stri_match_first_regex(req$PATH_INFO, private$regex$regex)[1,1])) {
+        if (req$REQUEST_METHOD %in% self$verbs) {
+          return(TRUE)
+        } else {
+          req$methodNotAllowed <- TRUE
+          req$verbsAllowed <- c(req$verbsAllowed, self$verbs)
+        }
+      }
+      return(FALSE)
```
schloerke, line 174: I believe the original logic needs to be kept and the new logic needs to be moved inside `default404Handler()`.  There can be a situation where you add handler to `GET /meztez` and add another handler of `POST /meztez`. Following this setup, asking for a `POST /meztez` would return `TRUE` w/ `req$methodNotAllowed == TRUE` as the `GET` route would be processed first.

Inside `default404Handler, the `pr` object should be accessible with `req$pr`.  I have not thought through the `private$regex$regex`

R/plumber.R
```diff
@@ -694,6 +694,10 @@ plumber <- R6Class(
             }
           }
         }
+        if (isTRUE(req$methodNotAllowed)) {
+          res$status = 405L
+          res$setHeader("Allow", paste(req$verbsAllowed, collapse = ", "))
+        }
```
schloerke, line 700: also move to `default404Handler()`

R/plumber-step.R
meztez: methodNotAllowed isTRUE (plumber.R line 697) is not going to be hit if a valid handle is found before. Thus no 405.

But I agree it could be cleaner.

schloerke: > methodNotAllowed isTRUE (plumber.R line 697) is not going to be hit if a valid handle is found before. Thus no 405.

True. Feels odd setting that variable when it's not completely true.

> But I agree it could be cleaner.

Yes. There are too many other things inside the router handler that use this function for me to feel safe. 

---------------

As always, thank you for the help!

meztez: I'll provide an alternative where we use a separate validPath  function in plumberStep (similar to canServe but do not check verbs).

meztez: I'm moving back on my words. I would have to replicate `handle` loop with `filt` done in `route` and everything to achieve a similar result inside the `default404handler`.

I'll keep it as it is for the moment, maybe I'll think of something.

R/default-handlers.R
```diff
@@ -1,5 +1,10 @@
 #' @include plumber.R
 default404Handler <- function(req, res){
+  if (isTRUE(req$methodNotAllowed)) {
+    res$status = 405L
+    res$setHeader("Allow", paste(req$verbsAllowed, collapse = ", "))
```
schloerke, line 5: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Allow

R/default-handlers.R
```diff
@@ -1,5 +1,10 @@
 #' @include plumber.R
 default404Handler <- function(req, res){
+  if (isTRUE(req$methodNotAllowed)) {
+    res$status = 405L
```
schloerke, line 4:
```suggestion
    res$status <- 405L
```

schloerke:
This is much better. Thank you. 

It still feels weird overloading the `$canServe()` method by adding to the `req` object. Don't know if we need to just recursively reinspect at `default404Handler()` time.  Will talk with `@cpsievert` later.

R/plumber-step.R
meztez: I don't have a better idea on how to handle it. No rush to merge this one so let it simmer, maybe a better solution will present itself.

schloerke:
Ok. I believe I have a solution that we can use inside `default404handler()`. It just needs access to the `pr` object which we can get from `req$pr`. 

----------------------

```r
# this function works under the assumption that the regular route was not found.  
# Here we are only trying to find if there are other routes that can be 
allowed_verbs <- function(pr, path_to_find) {

  verbs_allowed <- c()

  # look at all possible endpoints
  for (endpoint_group in pr$endpoints) {
    for (endpoint in endpoint_group) {
      if (endpoint$matchesPath(path_to_find)) {
        verbs_allowed <- c(verbs_allowed, endpoint$verbs)
      }
    }
  }

  # look at all possible mounts
  for (i in seq_along(pr$mounts)) {
    mount <- pr$mounts[[i]]
    mount_path <- sub("/$", "", names(pr$mounts)[i]) # trim trailing `/`

    # if the front of the urls don't match, move on to next mount
    if (!identical(
      substr(path_to_find, 0, nchar(mount_path)),
      mount_path
    )) {
      next
    }

    # remove path and recurse
    mount_path_to_find <- substr(path_to_find, nchar(mount_path) + 1, nchar(path_to_find))
    m_verbs <- allowed_verbs(mount, mount_path_to_find)

    # add any verbs found
    if (length(m_verbs) > 0) {
      verbs_allowed <- c(verbs_allowed, m_verbs)
    }
  }

  # return verbs
  return(verbs_allowed)
}

pr <- plumber$new()
sub <- plumber$new()
sub$handle("GET", "/test", force)
pr$mount("/barret", sub)


allowed_verbs(pr, path_to_find = "/barret/test")        # "GET"
allowed_verbs(pr, path_to_find = "/subroute/not_found") # NULL
allowed_verbs(pr, path_to_find = "/barret/")            # NULL
allowed_verbs(pr, path_to_find = "/barret/wrong")       # NULL


####################


is_405 <- function(pr, path_to_find, verb_to_find) {
  verbs_allowed <- allowed_verbs(pr, path_to_find)
  
  # nothing found, not 405
  if (length(verbs_allowed) == 0) return(FALSE)

  !(verb_to_find %in% verbs_allowed)
}

is_405(pr, path_to_find = "/barret/test", "GET")  # FALSE
is_405(pr, path_to_find = "/barret/test", "POST") # TRUE
is_405(pr, path_to_find = "/subroute/not_found")  # FALSE
```


---------------

...woot

schloerke:
(done with updates in the above comment)

meztez:
pr is not a value of req, did you expect it to be or should I add it?

schloerke:
Sorry. It is currently not implemented.

See https://github.com/rstudio/plumber/issues/310

Want to add it to this PR?

schloerke:
The `pr$req` could prolly be added here: https://github.com/rstudio/plumber/blob/038e0bc0b2fd7a2e4728ee35e6469bf6dc4e2e4b/R/plumber.R#L584

meztez:
I added it here : https://github.com/rstudio/plumber/blob/038e0bc0b2fd7a2e4728ee35e6469bf6dc4e2e4b/R/plumber.R#L814, above route

meztez:
Im finishing up, just need to fix test because we use a mock req with make_req that does not have a router

NEWS.md
```diff
@@ -48,6 +48,8 @@ plumber 1.0.0
 
 ### Minor new features and improvements
 
+* Support HTTP 405 Code. (@meztez, #554)
```
schloerke, line 51:
```suggestion
* Support HTTP 405 Code. (@meztez, #554)

* Attached the router to the incoming request object at `req$pr`. (@meztez, #554)
```

schloerke:
@meztez Is this PR ready for review / merging?

schloerke:
@meztez Is this PR ready for review / merging? 
(resending again as it was close to the other PR. Sorry for the spam).

meztez:
@schloerke  yes

R/default-handlers.R
```diff
@@ -29,3 +35,56 @@ defaultErrorHandler <- function(){
...
+    m_verbs <- allowed_verbs(mount, mount_path_to_find)
+
+    # add any verbs found
+    if (length(m_verbs) > 0) {
+      verbs_allowed <- c(verbs_allowed, m_verbs)
+    }
+  }
+
+  # return verbs
+  return(verbs_allowed)
```
cpsievert, line 79:
```suggestion
  sort(unique(verbs_allowed))
```

R/default-handlers.R
```diff
@@ -29,3 +35,56 @@ defaultErrorHandler <- function(){
...
+  # return verbs
+  sort(unique(verbs_allowed))
+}
+
+#' @noRd
+is_405 <- function(pr, path_to_find, verb_to_find) {
+  verbs_allowed <- allowed_verbs(pr, path_to_find)
+
+  # nothing found, not 405
+  if (length(verbs_allowed) == 0) return(FALSE)
```
cpsievert, line 87: Pretty sure this can be dropped because of how `%in%` works

```suggestion
```

tests/testthat/test-default-handlers.R
```diff
@@ -8,6 +8,22 @@ test_that("404 handler sets 404", {
...
+test_that("405 handling is ok, get the right verbs", {
+  pr <- plumber$new()
+  sub <- plumber$new()
+  sub$handle("GET", "/test", force)
+  pr$mount("/barret", sub)
+
+  expect_equal(allowed_verbs(pr, path_to_find = "/barret/test"), "GET")
+  expect_null(allowed_verbs(pr, path_to_find = "/subroute/not_found"), NULL)
+  expect_null(allowed_verbs(pr, path_to_find = "/barret/"), NULL)
+  expect_null(allowed_verbs(pr, path_to_find = "/barret/wrong"), NULL)
```
cpsievert, line 20:
```suggestion
  expect_null(allowed_verbs(pr, path_to_find = "/subroute/not_found"))
  expect_null(allowed_verbs(pr, path_to_find = "/barret/"))
  expect_null(allowed_verbs(pr, path_to_find = "/barret/wrong"))
```

R/default-handlers.R
```diff
@@ -29,3 +35,56 @@ defaultErrorHandler <- function(){
...
+  }
+
+  # look at all possible mounts
+  for (i in seq_along(pr$mounts)) {
+    mount <- pr$mounts[[i]]
+    mount_path <- sub("/$", "", names(pr$mounts)[i]) # trim trailing `/`
+
+    # if the front of the urls don't match, move on to next mount
+    if (!identical(
+      substr(path_to_find, 0, nchar(mount_path)),
```
cpsievert, line 62:
```suggestion
      substr(path_to_find, 1, nchar(mount_path)),
```

tests/testthat/test-default-handlers.R
```diff
@@ -8,6 +8,22 @@ test_that("404 handler sets 404", {
   expect_match(val$error, "Not Found")
 })
 
+test_that("405 handling is ok, get the right verbs", {
+  pr <- plumber$new()
+  sub <- plumber$new()
+  sub$handle("GET", "/test", force)
+  pr$mount("/barret", sub)
```
cpsievert, line 15: Let's add another test with `pr$mount("/", sub)` (or similar)

PR #560, opened 2020-06-23
Rename Swagger to OpenAPI Specification

meztez:
… OpenAPI

- NA Update NEWS, no user facing changes I think
- NA Add tests


schloerke:
Small comment: Can we move the content of `./R/parsers-body.R` underneath the content of `./R/parse-body.R`? This should all be grouped anyways.

meztez:
> 
> 
> Small comment: Can we move the content of `./R/parsers-body.R` underneath the content of `./R/parse-body.R`? This should all be grouped anyways. (Then delete `./R/parsers-body.R`)

Should we do the same for serializers? There is bunch of them under different files.

schloerke:
@meztez Let's leave them for now.  I'm going to create as `serializer_with_headers(...)` function as a base case for `serializer_content_type()`. This will allow for https://github.com/rstudio/plumber/pull/455 to have a leg to stand on and allow all other serializers to leverage the benefits / remove the boilerplate.

(PR coming in a few days.)

R/plumber.R
```diff
@@ -902,7 +902,13 @@ plumber <- R6Class(
     },
     #' @description Retrieve openAPI file
     openAPIFile = function() {
-      self$swaggerFile()
+      warning("`$openAPIFile()` has been deprecated in v1.0.0 and will be removed in a coming release. Please use `$apiSpec()`.")
```
cpsievert, line 905: Consider using lifecycle (if you anticipate wanting to avoid repeated warnings) and/or `.Deprecated`

R/plumber.R
```diff
@@ -902,7 +902,13 @@ plumber <- R6Class(
...
+      self$apiSpec()
+    },
+    #' @description Retrieve openAPI file
+    swaggerFile = function() {
+      warning("`$swaggerFile()` has been deprecated in v1.0.0 and will be removed in a coming release. Please use `$apiSpec()`.")
```
cpsievert, line 910: Same as above

R/plumber.R
cpsievert: If so, would be good in another PR (this is non-blocking)

PR #562, opened 2020-06-23
Interface mount

meztez:
Decouple ui mounting from the run method and generalize interface

PR task list:
- [x] Update NEWS
- [x] Add tests


R/ui.R
```diff
@@ -0,0 +1,132 @@
...
+  name = "swagger",
+  index = function() {
+    swagger::swagger_spec(
+      api_path = 'window.location.origin + window.location.pathname.replace(/\\(__swagger__\\\\/|__swagger__\\\\/index.html\\)$/, "") + "openapi.json"',
+      version = "3"
+    )
+  },
+  static = swagger::swagger_path
+)
+
```
meztez, line 118: swaggerInterface should go in swagger package.

R/ui.R
```diff
@@ -0,0 +1,132 @@
...
+#   package = "redoc",
+#   name = "redoc",
+#   index = function(redoc_options = structure(list(), names = character())) {
+#     redoc::redoc_spec(
+#       spec_url = "\' + window.location.origin + window.location.pathname.replace(/\\(__redoc__\\\\/|__redoc__\\\\/index.html\\)$/, '') + 'openapi.json' + \'",
+#       redoc_options = redoc_options
+#     )
+#   },
+#   static = redoc::redoc_path
+# )
```
meztez, line 129: redocInterface in redoc package

R/ui.R
```diff
@@ -0,0 +1,132 @@
...
+
+mountInterfaces(swaggerInterface)
```
meztez, line 131: This is where I'm not sure how to handle it. `mountInterfaces` is from plumber. Where should this particular part so that interface is available when needed.

Currently interfaces are added to globals on package load but they could also be added only when user use the $run method with a ui.

NEWS.md
```diff
@@ -3,6 +3,8 @@ plumber (version)
 
 ### New features
 
+* Add support for `Redoc` OpenAPI UI. (@meztez)
+* Generalize interface integration. (@meztez)
```
schloerke, line 7:
```suggestion
* Add support for `Redoc` OpenAPI UI. (@meztez, #562)
* Generalize interface integration. (@meztez, #562)
```

NEWS.md
```diff
@@ -11,7 +13,11 @@ plumber (version)
 * Fix `plumb()` function when `plumb()`ing a directory so that `plumber.R` is
   not a requirement if a valid `entrypoint.R` file is found. 
   (@blairj09, [#471](https://github.com/trestletech/plumber/pull/471)).
-  
+* Fix OpenAPI url error when using an empty `HTTP_REFERER`. (@meztez)
+
+### Breaking changes
+* `$run()` arguments `swagger` and `swaggerCallback` replaced by `ui` and `callback`. (@meztez)
+* Moved user provided OpenAPI Specification function from `$run()` to `$setApiHandler`. (@meztez)
```
schloerke, line 20:
```suggestion
* `$run()` arguments `swagger` and `swaggerCallback` replaced by `ui` and `callback`. `ui` is now set to `TRUE` by default. (@meztez, #562)
* Moved user provided OpenAPI Specification function from `$run()` to `$setApiHandler()`. (@meztez, #562)
```

scripts/manual_testing.R
```diff
@@ -10,13 +10,22 @@ library(plumber)
...
   pr <- plumber$new()
   pr$handle("GET", "/:path/here", function(){})
-
+  pr$setApiHandler(function(spec) {
+    spec$info$title <- Sys.time()
+    spec
+  })
+  pr$run(
+    port = 1234,
+    ui = "swagger"
```
schloerke, line 19: Let's follow pattern of how `api` is set outside of the run method. 

Trying to keep the run method the bare minimum.

Ex: 
```r
pr$setUI("redoc")
pr$setUI(FALSE)
pr$setUI("redoc", scrollYOffset = 250, disableSearch = TRUE)
```

Something like this to maintain the args and ui value:
```r
setUI = function(ui, ...) {
  private$ui_info = list(
    ui = ui,
    args = list(...)
  )
}
```

R/ui.R
schloerke: There could be a public method in each package. Ex: `swagger::plumber_mount_interface()` 

This method could also be run in `.onLoad()`?

```r

plumber_mount_interface <- function() {
  plumber::mountInterface(....)
}

.onLoad <- function(...) {
  if (require("plumber", quietly = TRUE)) {
    plumber_mount_interface()
  }

}
```


> Currently interfaces are added to globals on package load

As long as they say `library(MYINTERFACEPKG)` in the plumb file, then it'll get picked up.

This setup will also help when APIs are deployed so packrat finds pkg `MYINTERFACEPKG`. If it's only a variable, packrat won't bundle it when deployed.

R/ui.R
```diff
@@ -0,0 +1,132 @@
...
+
+}
+
+#' Mount Interface UI
+#' @noRd
+mountInterfaces <- function(interface) {
+  mountInterface <- function(pr, api_url, ...) {
+    if (!requireNamespace(interface$package, quietly = TRUE)) {
+      stop(interface$package, " must be installed for the ", interface$name," UI to be displayed")
+    }
```
schloerke, line 86:
```suggestion
mountInterface <- function(interface) {
  if (!requireNamespace(interface$package, quietly = TRUE)) {
    stop(interface$package, " must be installed for the ", interface$name," UI to be displayed")
  }
  interfaceFunc <- function(pr, api_url, ...) {
```

schloerke: Let's move the pkg check higher up in the processing

R/ui.R
```diff
@@ -0,0 +1,132 @@
...
+    for (path in paste0(interfacePath, c("index.html", ""))) {
+      pr$handle(
+        "GET", path, interface_index,
+        serializer = serializer_html()
+      )
+    }
+    pr$mount(interfacePath, PlumberStatic$new(interface$static()))
+    return(interfaceUrl)
+  }
+  .globals$interfaces[[interface$name]] <- mountInterface
```
schloerke, line 104:
```suggestion
  .globals$interfaces[[interface$name]] <- interfaceFunc
```

R/ui.R
```diff
@@ -0,0 +1,132 @@
...
+    interface_index <- function() {
+      html_content
+    }
+    for (path in paste0(interfacePath, c("index.html", ""))) {
+      pr$handle(
+        "GET", path, interface_index,
+        serializer = serializer_html()
+      )
+    }
+    pr$mount(interfacePath, PlumberStatic$new(interface$static()))
```
schloerke, line 101: Also changing `interface$static` to an object over a method.

```suggestion
    if (!is.null(interface$static)) {
      pr$mount(interfacePath, PlumberStatic$new(interface$static))
    }
```

R/ui.R
```diff
@@ -0,0 +1,132 @@
...
+swaggerInterface <- list(
+  package = "swagger",
+  name = "swagger",
+  index = function() {
+    swagger::swagger_spec(
+      api_path = 'window.location.origin + window.location.pathname.replace(/\\(__swagger__\\\\/|__swagger__\\\\/index.html\\)$/, "") + "openapi.json"',
+      version = "3"
+    )
+  },
+  static = swagger::swagger_path
```
schloerke, line 116:
```suggestion
  static = swagger::swagger_path()
```

schloerke, line 116: I understand the function now... Let's pass in the plumber object if possible

```suggestion
  index = function(pr) {
    swagger::swagger_spec(
      api_path = 'window.location.origin + window.location.pathname.replace(/\\(__swagger__\\\\/|__swagger__\\\\/index.html\\)$/, "") + "openapi.json"',
      version = "3"
    )
  },
  static = function(pr) {
    swagger::swagger_path()
  }
```

schloerke: Removing in favor of passing in `pr`

R/ui.R
schloerke: Added checks for the interface object

R/ui.R
schloerke: removing

R/ui.R
```diff
@@ -0,0 +1,132 @@
+#' @include globals.R
+
+# Mount OpenAPI and UI
+#' @noRd
+mountUI <- function(pr, host, port, ui, callback, ...) {
```
schloerke, line 5: Pass in args as a list

```suggestion
mountUI <- function(pr, host, port, ui, callback, args = list()) {
```

R/ui.R
```diff
@@ -0,0 +1,132 @@
...
+  mountOpenAPI(pr, api_url)
+
+  # Mount UIs
+  if (isTRUE(ui)) {
+    ui <- getOption("plumber.ui", "swagger")
+  }
+  interface <- ui[1]
+  ui_mount <- .globals$interfaces[[interface]]
+  if (!is.null(ui_mount)) {
+    ui_url <- ui_mount(pr, api_url, ...)
```
schloerke, line 29:
```suggestion
    ui_url <- ui_mount(pr = pr, api_url = api_url, args = args)
```

R/ui.R
```diff
@@ -0,0 +1,132 @@
...
+  pr$handle("GET", "/openapi.json", openapi_fun, serializer = serializer_unboxed_json())
+
+  return(invisible())
+
+}
+
+#' Mount Interface UI
+#' @noRd
+mountInterfaces <- function(interface) {
+  mountInterface <- function(pr, api_url, ...) {
```
schloerke, line 83:
```suggestion
  mountInterface <- function(pr, api_url, args) {
```

R/ui.R
```diff
@@ -0,0 +1,132 @@
...
+      stop(interface$package, " must be installed for the ", interface$name," UI to be displayed")
+    }
+
+    interfacePath <- paste0("/__", tolower(interface$name), "__/")
+    interfaceUrl <- paste0(api_url, interfacePath)
+
+    html_content <- interface$index(...)
+    interface_index <- function() {
+      html_content
+    }
```
schloerke, line 94: Call it every time. Might have different displays given particular routers / routes.

```suggestion
    args$pr <- pr
    interface_index <- function() {
      do.call(interface$index, args)
    }
```

R/ui.R
schloerke: This will require an update in `redoc`

R/ui.R
schloerke: Yes!

schloerke:
Lots of changes. Trying to future-safeguard external "contracts" with other packages as we won't be able to update them when we want.

R/ui.R
meztez: Why? Both swagger_path redoc_path do not accept arguments.

schloerke: I don't know if we should pass in the OAS `version`, `pr` (which is prolly not used), or nothing.

It is possible that down the line, my version of plumber has a different OAS version than someone else's version. However, both of us would be able to use the same version of `swagger` pkg. `swagger` would then retrieve the proper OAS ui matching the version.

meztez: What I think
- `version` should come from plumber in the args list, I'll remove it from the interface.
- `pr` should not be passed as it would require `swagger` to be aware of the current `pr` in plumber

schloerke: Works for me!

scripts/manual_testing.R
meztez: While I was at it, I think I prefer having different parameter for enabling ui and selecting it. See `setUI` signature.

R/ui.R
meztez: Since this will live in swagger and redoc package, change will be made there if they are needed.

R/ui.R
meztez: This is going in redoc and swagger package. It should be non blocking here. I'll pr them some time soon.

meztez:
After doing some test, I added functionalities to help unmount UI on.exit of the run method.

It allows for quick prototyping.

ui Enabled
pr$run()
stop
change pr
pr$run()
stop
change pr
setUI(enabled = TRUE, "redoc", redoc_options = list(...))
pr$run()

Without it, UI endpoints are added in sequence in an unnamed list `c(private$ends[[preempt]], ep)`, the first UI you mount will always be picked up when you ask for it even if you updated `pr` since the changes are further down the endpoints list.

R/ui.R
meztez: Pr done in swagger, redoc updated, works from 2.0.0.30

R/ui.R
```diff
@@ -0,0 +1,165 @@
...
+  pr$removeHandle("GET", "/openapi.json")
+  return(NULL)
+
+}
+
+#' Mount Interface UI
+#' @param interface An interface (list) that plumber can use to mount
+#' a UI.
+#' @export
+mountInterface <- function(interface) {
```
schloerke, line 103: Could we switch to snake case, please? (Same with functions above. Thank you!)

schloerke: And variables below too

R/plumber.R
```diff
@@ -875,6 +812,29 @@ plumber <- R6Class(
...
+      if (!is.null(api_fun)) {
+        stopifnot(is.function(api_fun))
+      }
+      private$apiHandler <- api_fun
+    },
+    #' @description Set UI to use for API
+    #' @param enabled a logical value.
+    #' @param ui a character value. Default to `plumber.ui` option value.
+    #' @param ... Other params to be passed to ui functions.
+    setUI = function(enabled = TRUE, ui = getOption("plumber.ui", "swagger"), ...) {
```
schloerke, line 828: The `enabled` argument seems redundant. 
* If `enabled=FALSE`, `ui` is not used.
* If `enabled=TRUE`, `ui` is required.

I don't mind having the `enabled` value in the `private$ui_info`.  

```r
if (length(ui) == 1 && is.character(ui)) {
  enabled <- TRUE
} else {
  enabled <- FALSE
  ui <- "__not_enabled__"
}
```

R/ui.R
```diff
@@ -0,0 +1,165 @@
+#' @include globals.R
+
+# Mount OpenAPI and UI
+#' @noRd
+mountUI <- function(pr, host, port, ui_info, callback) {
+
```
schloerke, line 6:
```suggestion
# return early if not enabled
if (!isTRUE(ui_info$enabled)) {
  return()
}
```

R/ui.R
```diff
@@ -0,0 +1,165 @@
...
+  # Mount UIs
+  ui_mount <- .globals$interfaces$mount[[ui_info$ui]]
+  if (!is.null(ui_mount)) {
+    ui_url <- do.call(ui_mount, c(list(pr, api_url), ui_info$args))
+    message("Running ", ui_info$ui, " UI at ", ui_url, sep = "")
+  } else {
+    message("Unknown user interface \"", ui_info$ui,"\". ",
+            ". Maybe try library(", ui_info$ui,").")
+    return(NULL)
```
schloerke, line 30: The return value is not used. For readability, do not return anything.

```suggestion
    return()
```

R/ui.R
```diff
@@ -0,0 +1,165 @@
...
+  }
+
+  # Use callback when defined
+  if (is.function(callback)) {
+    callback(ui_url)
+  }
+
+  return(NULL)
```
schloerke, line 38:
```suggestion
  invisible()
```

R/ui.R
meztez: across package it uses return(NULL) which is the similar to return() as I understand it. Should we use a consistent return across package?

schloerke: Correct. They have the same returned value.

> Should we use a consistent return?

I'm all for it. (Not a big priority)

If it's for the purposes of early returning, let's use `return()`.

If the value means something, `return(NULL)`.

meztez:
Could it be possible to have a custom mounting point for the ui instead of having hardcoded path?

schloerke:
@meztez Yes. This seems reasonable!

meztez:
> 
> 
> @meztez Yes. This seems reasonable!

ok, I'll rename a couple thing then. I think I prefer ui instead of interface and add_ui seems like a better name then mount_interface, it follows the addParsers and addSerializers philosophie.

I have figured out, the only missing piece is how to replace an arbitrary string in javascript for the spec part

```js
window.location.pathname.replace(/\\(__swagger__\\\\/|__swagger__\\\\/index.html\\)$/, "")
```

schloerke:
> I have figured out, the only missing piece is how to replace an arbitrary string in javascript for the spec part

There is a `version` parameter within `swagger::swagger_spec()`.  So `swagger::swagger_spec(path, version = version)`

IDK if plumber should supply the full version, or just the major version. Part of me says the full version.

schloerke:
(Sorry for the `options_plumber()` conflicts. Decided to remove the `_PACKAGE` docs as they were only really documenting the `options_plumber()`

meztez:
> 
> 
> > I have figured out, the only missing piece is how to replace an arbitrary string in javascript for the spec part
> 
> There is a `version` parameter within `swagger::swagger_spec()`. So `swagger::swagger_spec(path, version = version)`
Yeah that`s what I have, but I have to build the regex for replace and make sure no special character can mess it up. Im not as familiar with javascript regex so it is gonna take me a bit more time.
> 
> IDK if plumber should supply the full version, or just the major version. Part of me says the full version.
Plumber would not supply anything as this is going to be provided by the swagger package, but you could define version in setUI. It would be relative to the version parameter definition in swagger.

schloerke:
> > IDK if plumber should supply the full version, or just the major version. Part of me says the full version.
> Plumber would not supply anything as this is going to be provided by the swagger package

You're right. The version is the swaggerUI version, not the OAS version.

meztez:
`path` parameter is character vector of possibly length > 1

```r
index = function(path, version = "3", ...) {
    api_path = replace_function(path) {
      # strip any string found in path from the end of window.location.pathname
      # maybe keep trailing / or add it before openapi.json
      # /\\(__swagger__\\\\/|__swagger__\\\\/index.html\\)$/
      'window.location.origin + window.location.pathname.replace(, "") + "/openapi.json"'
    }
    swagger::swagger_spec(
      api_path = api_path(path,
      version = version
    )
  }
```

meztez:
I will do a different PR for the custom OpenAPI path.

This one has enough stuff already.

R/plumber-options.R
```diff
@@ -1,60 +1,91 @@
-
 #' How to use Plumber options
```
schloerke, line 1:
```suggestion
```

R/plumber-options.R
```diff
@@ -1,60 +1,91 @@
...
-#'
-#' @section Options:
+#' @title Plumber options
```
schloerke, line 2:
```suggestion
#' Plumber options
```

R/plumber-options.R
```diff
@@ -1,60 +1,91 @@
...
+#' @details
```
schloerke, line 3:
```suggestion
#'
```

R/ui.R
```diff
@@ -0,0 +1,186 @@
...
+      api_path = 'window.location.origin + window.location.pathname.replace(/\\(__swagger__\\\\/|__swagger__\\\\/index.html\\)$/, "") + "openapi.json"',
+      version = version
+    )
+  },
+  static = function(version = "3", ...) {
+    swagger::swagger_path(version)
+  }
+)
+
+add_ui(swagger_ui)
```
schloerke, line 186: Can this be moved into the `.onLoad()` similar to how we did the parsers and serializers

R/plumber.R
```diff
@@ -893,6 +830,33 @@ plumber <- R6Class(
...
+      }
+      private$apiHandler <- api_fun
+    },
+    #' @description Set UI to use for API
+    #' @param ui a character value or a logical value. Default to `plumber.ui` option value.
+    #' @param ... Other params to be passed to ui functions.
+    setUI = function(ui = getOption("plumber.ui", TRUE), ...) {
+      stopifnot(isTRUE(length(ui) == 1L))
+      stopifnot(is.logical(ui) || is.character(ui))
+      if (isTRUE(ui) || is.character(ui)) {
```
schloerke, line 848: Must set a value to a character string. For now, let's default to `"swagger"`

```suggestion
      if (isTRUE(ui)) {
        ui <- "swagger"
      }
      if (is.character(ui)) {
```

R/ui.R
meztez: Yes, this will have to end up in swager package at some point.

meztez:
I've done the same for meztez/rapidoc that I did for meztez/redoc, changed the signature to look like @schloerke suggested.

Seems like rapidoc has more customizations and no for profit attachment to it.
https://github.com/mrin9/RapiDoc/issues/141

```r
library(redoc)
pr$setUI(ui = "redoc")
pr$run(port = 1234)
pr$setUI(ui = "redoc", scrollYOffset = 250, disableSearch = TRUE)
pr$run(port = 1235)
library(rapidoc)
pr$setUI(ui = "rapidoc", heading_text = "Rumbepl", bg_color = "#457872")
pr$run(port = 1235)
```

schloerke:
@meztez Is this PR ready for review / merging?

meztez:
> 
> 
> @meztez Is this PR ready for review / merging?

Yes, I am done with this one.

R/plumber.R
```diff
@@ -254,44 +255,23 @@ plumber <- R6Class(
...
-    #' pr$run(swagger = swagger)
-    #' }
+    #' `callback` When set, it will be called with a character string corresponding
+    #' to the UI url. It allows RStudio to open UI when plumber router
+    #' run method is executed using default `plumber.ui.url` option.
     run = function(
       host = '127.0.0.1',
       port = getOption('plumber.port'),
-      swagger = interactive(),
       debug = interactive(),
```
cpsievert, line 269: Add back `swagger` argument (so as not to break existing to code), then throw a deprecation warning. If it's somewhat straight-forward/sensible, also try translating the value to the new approach

R/plumber.R
```diff
@@ -254,44 +255,23 @@ plumber <- R6Class(
...
-      swaggerCallback = getOption('plumber.swagger.url', NULL)
+      callback = getOption('plumber.ui.callback', getOption('plumber.swagger.url', NULL))
     ) {
```
cpsievert, line 271: Similar thing here (don't completely remove `swaggerCallback`)

R/plumber-options.R
```diff
@@ -1,60 +1,90 @@
...
+  apiPort              = getOption("plumber.apiPort"),
+  apiPath              = getOption("plumber.apiPath"),
+  apiURL               = getOption("plumber.apiURL"),
   debug                = getOption("plumber.debug"),
   maxRequestSize       = getOption("plumber.maxRequestSize"),
   postBody             = getOption("plumber.postBody"),
   port                 = getOption("plumber.port"),
   sharedSecret         = getOption("plumber.sharedSecret"),
-  swagger.url          = getOption("plumber.swagger.url")
+  swagger.url          = getOption("plumber.swagger.url"),
```
cpsievert, line 71: Since this is a new function, maybe consider removing this

cpsievert: Tests could black-list this parameter as well

R/plumber-step.R
```diff
@@ -222,6 +222,13 @@ PlumberEndpoint <- R6Class(
     #' @description retrieve endpoint expression parameters
     getFuncParams = function() {
       getArgsMetadata(private$func)
+    },
+    #' @description retrieve endpoint defined parameters
+    getEndpointParams = function() {
+      if (is.na(self$params) || isTRUE(length(self$params) == 0L)) {
+        return(list())
+      }
```
cpsievert, line 230:
```suggestion
      if (length(self$params) == 0) {
        return(list())
      }
      if (isTRUE(is.na(self$params))) {
        return(list())
      }
```

R/plumber.R
```diff
@@ -402,6 +320,18 @@ plumber <- R6Class(
...
+    #' @description Unmount a plumber router
+    #' @param path a character string. Where to mount router.
+    unmount = function(path){
+      # Ensure that the path has both a leading and trailing slash.
+      if (!startsWith(path, "/")) {
+        path <- paste0("/", path)
+      }
+      if (!endsWith(path, "/")) {
+        path <- paste0(path, "/")
+      }
```
cpsievert, line 332:
```suggestion
      if (!grepl("^/", path)) {
        path <- paste0("/", path)
      }
      if (!grepl("/$", path)) {
        path <- paste0(path, "/")
      }
```

meztez: Should we replace across package? They are currently being used in master.

R/plumber.R
```diff
@@ -893,6 +830,36 @@ plumber <- R6Class(
...
       private$errorHandler <- fun
     },
+    #' @description Add a function to customize what is returned in `$apiSpec()`.
+    #' @param api_fun This function should accept an OpenAPI Specification autogenerated by `plumber`.  It should return a list object in valid OpenAPI Specification format. This will not be validated.
+    setApiHandler = function(api_fun = force) {
+      # allow null values
+      if (!is.null(api_fun)) {
+        stopifnot(is.function(api_fun))
+      }
+      private$apiHandler <- api_fun
```
cpsievert, line 840:
```suggestion
    setApiHandler = function(api_fun = NULL) {
      if (is.null(api_fun)) {
        api_fun <- identity
      }
      if (!is.function(api_fun)) {
        stop("`api_fun` must be a function")
      }
      private$apiHandler <- api_fun
```

R/plumber.R
```diff
@@ -893,6 +830,36 @@ plumber <- R6Class(
...
+    },
+    #' @description Set UI to use for API
+    #' @param ui a character value or a logical value. Default to `plumber.ui` option value.
+    #' @param ... Other params to be passed to ui functions.
+    setUI = function(ui = getOption("plumber.ui", TRUE), ...) {
+      stopifnot(isTRUE(length(ui) == 1L))
```
cpsievert, line 846:
```suggestion
      stopifnot(length(ui) == 1)
```

R/plumber.R
```diff
@@ -223,6 +223,7 @@ plumber <- R6Class(
       private$errorHandler <- defaultErrorHandler()
       private$notFoundHandler <- default404Handler
       private$maxSize <- getOption('plumber.maxRequestSize', 0) #0 Unlimited
+      private$ui_info <- list(enabled = TRUE, ui = getOption("plumber.ui", TRUE), args = list())
```
cpsievert, line 226:
```suggestion
      self$setUI()
```

R/plumber.R
```diff
@@ -913,7 +880,11 @@ plumber <- R6Class(
       # (like API version) are satisfied.
       ret <- utils::modifyList(defaultGlobals, def)
 
-      # remove NA or NULL values, which swagger doesn't like
+      if (!is.null(private$apiHandler)) {
+        ret <- private$apiHandler(ret)
+      }
```
cpsievert, line 887:
```suggestion
      ret <- private$apiHandler(ret)
```

R/plumber.R
```diff
@@ -1061,6 +1032,9 @@ plumber <- R6Class(
     notFoundHandler = NULL,
     maxSize = NULL, # Max request size in bytes
 
+    apiHandler = NULL,
```
cpsievert, line 1037:
```suggestion
    apiHandler = identity,
```

R/plumber.R
```diff
@@ -495,6 +425,13 @@ plumber <- R6Class(
       }
       private$addEndpointInternal(endpoint, preempt)
     },
+    #' @description Remove endpoints
+    #' @param methods a character string. http method.
+    #' @param path a character string. Api endpoints
+    #' @param preempt a preempt function.
+    removeHandle = function(methods, path, preempt){
```
cpsievert, line 432:
```suggestion
    removeHandle = function(methods, path, preempt = NULL){
```

R/plumber.R
```diff
@@ -1087,6 +1061,22 @@ plumber <- R6Class(
 
       private$ends[[preempt]] <- c(private$ends[[preempt]], ep)
     },
+    removeEndpointInternal = function(methods, path, preempt){
+      noPreempt <- missing(preempt) || is.null(preempt)
```
cpsievert, line 1067:
```suggestion
      noPreempt <- is.null(preempt)
```

R/plumber.R
```diff
@@ -1087,6 +1061,22 @@ plumber <- R6Class(
...
+      }
+      i <- 0L; toRemove <- integer()
+      for (ep in private$ends[[preempt]]) {
+        i <- i + 1L
+        if (isTRUE(all(ep$verbs %in% methods)) &&
+            isTRUE(ep$path == path)) {
+          toRemove <- c(toRemove, i)
+        }
+      }
+      private$ends[[preempt]][toRemove] <- NULL
```
cpsievert, line 1078: Consider replacing this for loop with a `vapply`

R/plumber.R
```diff
@@ -1156,13 +1146,6 @@ urlHost <- function(host, port, changeHostLocation = FALSE) {
   if (grepl(":[^/]", host)) {
     host <- paste0("[", host, "]")
   }
```
cpsievert, line 1148: Add a check for `nzchar(scheme)`

R/ui.R
```diff
@@ -0,0 +1,209 @@
...
+  if (!isTRUE(ui_info$enabled)) {
+    return()
+  }
+
+  # Unount openAPI spec paths openapi.json
+  unmount_openapi(pr)
+
+  # Mount UIs
+  ui_unmount <- .globals$UIs$unmount[[ui_info$ui]]
+  if (!is.null(ui_unmount)) {
```
cpsievert, line 75:
```suggestion
  if (length(ui_unmount) && is.function(ui_unmount)) {
```

R/ui.R
```diff
@@ -0,0 +1,209 @@
...
+  if (is_ui_available(ui_info$ui)) {
+    ui_mount <- .globals$UIs$mount[[ui_info$ui]]
+    ui_url <- do.call(ui_mount, c(list(pr, api_url), ui_info$args))
+    message("Running ", ui_info$ui, " UI at ", ui_url, sep = "")
+  } else {
+    return()
+  }
+
+  # Use callback when defined
+  if (is.function(callback)) {
```
cpsievert, line 42:
```suggestion
  if (length(callback) && is.function(callback)) {
```

R/ui.R
```diff
@@ -0,0 +1,209 @@
...
+      path   = getOption("plumber.apiPath", ""),
+      changeHostLocation = TRUE
+    )
+  )
+
+  # Mount openAPI spec paths openapi.json
+  mount_openapi(pr, api_url)
+
+  # Mount UIs
+  if (isTRUE(length(.globals$UIs$mount)==0L)) {
```
cpsievert, line 28:
```suggestion
  if (length(.globals$UIs$mount) == 0) {
```

R/plumber.R
meztez: setting it to something else than NULL makes it locked : cannot change value of locked binding for 'apiHandler', will change in init method

R/plumber.R
```diff
@@ -258,48 +260,47 @@ plumber <- R6Class(
...
+    #' to the UI url. It allows RStudio to open UI when plumber router
+    #' run method is executed using default `plumber.ui.url` option.
     run = function(
       host = '127.0.0.1',
       port = getOption('plumber.port'),
-      swagger = interactive(),
+      swagger,
       debug = interactive(),
-      swaggerCallback = getOption('plumber.swagger.url', NULL)
+      swaggerCallback,
```
cpsievert, line 278:
```suggestion
      swaggerCallback, 
      ...,
```

R/plumber.R
```diff
@@ -258,48 +260,47 @@ plumber <- R6Class(
...
+
       if (isTRUE(private$disable_run)) {
         stop("Plumber router `$run()` method should not be called while `plumb()`ing a file")
       }
 
-      port <- findPort(port)
+      if (!missing(swagger)) {
+        warning("`swagger` parameter has been deprecated in v1.0.0 and will be removed in a coming release. Please use `$setUI()` and `$setApiHandler()`.")
+        if (is.function(swagger)) {
+          self$setUI = TRUE
```
cpsievert, line 289:
```suggestion
          self$setUI(TRUE)
```

R/plumber-options.R
```diff
@@ -1,60 +1,84 @@
...
+#' \item{`plumber.apiHost`}{Host used to build UI url and server url for
+#' OpenAPI Specification. Defaults to `host` defined by `run` method, or an empty string
+#' when used outside a running router}
+#' \item{`plumber.apiPort`}{Port used to build UI url and server url for
+#' OpenAPI Specification. Defaults to `port` defined by `run` method, or an empty string
+#' when used outside a running router}
+#' \item{`plumber.apiPath`}{Path used to build UI url and server url for
+#' OpenAPI Specification. Defaults to an empty string}
+#' \item{`plumber.apiURL`}{UI url and server urls for OpenAPI Specification respecting
+#' pattern `scheme://host:port/path`. Defaults to the combination of the above options}
```
cpsievert, line 20: Put `plumber.apiURL` first and make it more clear that if you specify this option, the rest of the other `api*` options are ignored

R/plumber.R
```diff
@@ -258,48 +260,47 @@ plumber <- R6Class(
...
-    #' pr <- plumber$new
-    #' swagger <- function(pr_, spec) {
-    #'   spec$servers[[1]]$description <- "MyCustomAPISpec"
-    #'   spec
-    #' }
-    #' pr$run(swagger = swagger)
-    #' }
+    #' `callback` When set, it will be called with a character string corresponding
+    #' to the UI url. It allows RStudio to open UI when plumber router
+    #' run method is executed using default `plumber.ui.url` option.
```
cpsievert, line 272:
```suggestion
    #' run method is executed using default `plumber.ui.callback` option.
```

R/plumber.R
meztez: Nothing is using the triple dots, are you keeping them for historical compatibility?

inst/examples/15-openapi-spec/entrypoint.R
```diff
@@ -19,34 +19,6 @@ handler <- function(num) { sum(as.integer(num)) }
 
 pr$handle("GET", "/sum", handler, serializer = serializer_json())
 
-# pr$run(swagger = swagger) # TODO-barret make function for this
+pr$setApiHandler(api_fun = openapi_func)
```
cpsievert, line 22: I'd like to see the naming of this method share a "basename" with `$apiSpec()`...perhaps this should be named `$apiSpecTransform()`/`$apiSpecMutate()`?

R/plumber.R
```diff
@@ -1166,13 +1178,10 @@ urlHost <- function(host, port, changeHostLocation = FALSE) {
...
     host <- paste0("[", host, "]")
   }
-  # if no match against a protocol
-  if (!grepl("://", host)) {
-    # add http protocol
-    # RStudio IDE does NOT like empty protocols like "127.0.0.1:1234/route"
-    # Works if supplying "http://127.0.0.1:1234/route"
-    host <- paste0("http://", host)
+
+  if (!nzchar(scheme)) {
```
cpsievert, line 1182:
```suggestion
  if (is.null(scheme) || !nzchar(scheme)) {
```

inst/examples/15-openapi-spec/entrypoint.R
cpsievert: @meztez We still prefer `$apiSpecTransform()` over `$apiSpecHandler()` since `$handler()` means something different.

NEWS.md
```diff
@@ -18,27 +18,31 @@ plumber 1.0.0
...
-
 ### Breaking changes
 
-* Plumber's swagger definition is now defined using
-  [OpenAPI 3](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.2.md),
+* Plumber's OpenAPI Specification is now defined using
+  [OpenAPI 3](https://github.com/OAI/OpenAPI-Specification/blob/master/versions/3.0.3.md),
   upgrading from Swagger Specification. (#365)
 
+* Plumber router `$run()` method arguments `swagger` removed and `swaggerCallback` replaced by `callback`. User interface is now enabled by default and managed through Plumber router `$setUI()` method and option `plumber.ui`. (@meztez, #562)
```
cpsievert, line 27: `removed` -> `deprecated`

NEWS.md
```diff
@@ -18,27 +18,31 @@ plumber 1.0.0
...
 * The source files used in plumber **must use** the UTF-8 encoding if they contain
   non-ASCII characters (@shrektan, #312,
   #328).
 
 * R repository modified to `focal-cran40` using Ubuntu 20.04 LTS for digital ocean provisioning (@meztez, #529)
 
-
 ### New features
 
+* Generalize user interface integration. Plumber can now use other OpenAPI compatible user interfaces like `RapiDoc` and `Redoc`. (@meztez, #562)
```
cpsievert, line 37: Provide a link to the Github repos

NEWS.md
```diff
@@ -86,6 +90,8 @@ plumber 1.0.0
 
 ### Bug fixes
 
+* Fix OpenAPI server url error when `HTTP_REFERER` is NULL. (@meztez, #562)
```
cpsievert, line 93: This doesn't seem worth mentioning (has anyone filed an issue about it?)

```suggestion
```

meztez: Removing. #421.

R/plumber.R
```diff
@@ -258,48 +260,49 @@ plumber <- R6Class(
...
     run = function(
       host = '127.0.0.1',
       port = getOption('plumber.port'),
-      swagger = interactive(),
+      swagger,
       debug = interactive(),
-      swaggerCallback = getOption('plumber.swagger.url', NULL)
+      swaggerCallback,
+      ...,
+      callback = getOption('plumber.ui.callback', getOption('plumber.swagger.url', NULL))
```
schloerke, line 281: Received more feedback from the Connect team.  I'd like to remove this argument and make it a R6 router method.   

The default return value for the R6 method should still be `getOption('plumber.swagger.url')`.

I'd also like to do this for `debug`. Where the default value is `interactive()`. 

This makes it so the `$run()` method has NOTHING in the parameters except for `host` and `port`.

meztez: public :

setUICallback or add a callback parameter to setUI?

setDebug

private : 
callback
debug

muschellij2:
It looks like https://github.com/rstudio/plumber/blob/dcc178e95757a7d3b0d1fcee22b6db169a4c36c0/R/plumber.R#L871 was changed to `set_ui` but still uses `setUI` in https://github.com/rstudio/plumber/blob/13f813400062c9a287af4602172a4232b4c6a545/inst/server/plumber.service#L5

schloerke:
@muschellij2 Fixed in 69e5011. 🙌

PR #568, opened 2020-06-24
Add router modifier tag support (`#' @plumber`)

meztez:
Add support for a @plumber tag that use the expr to modify the plumber router programmatically.
```r
#* @plumber
function(pr){
  pr$handle("GET", "/avatartare", function(raw) {raw}, serializer = serializer_json)
}
```

PR task list:
- [X] Update NEWS
- [X] Add tests
- [X] Update documentation with `devtools::document()`

PR #574, opened 2020-06-28
Prevent `$run()` method while `plumb()`ing file

### LINKED ISSUE #573: Calling pr$run() inside a @plumber tag function
```R
#* @plumber
function(pr) {
  pr$run
}

#* Echo back the input
#* @get /echo
function(msg = "") {
}
```

It preemptively call run method.

Option A : We let the user deal with the consequences
Option B : We run all @plumber code last (non-viable still cause issue, can't be sure of the order the run is called in expression)
Option C : Run method does nothing when called inside a plumber tag

I'll PR for C

meztez:
## Pull Request

Fixes #573 

Introduced by #568

PR task list:
- [NA] Update NEWS
- [X] Add tests

### DIFF
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -293,7 +293,7 @@ plumber <- R6Class(
       swaggerCallback = getOption('plumber.swagger.url', NULL)
     ) {
       if (isTRUE(private$disable_run)) {
-        stop("Plumber router run method called from a router modifier expression (@plumber)")
+        stop("Plumber router `$run()` method should not be called while `plumb()`ing a file")
       }
 
       port <- findPort(port)
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -233,6 +233,8 @@ plumber <- R6Class(
       if (!is.null(file)){
         private$lines <- readUTF8(file)
         private$parsed <- parseUTF8(file)
+        private$disable <- TRUE
+        on.exit(private$disable <- FALSE, add = TRUE)
 
         for (i in 1:length(private$parsed)){
           e <- private$parsed[i]
@@ -288,6 +290,8 @@ plumber <- R6Class(
       debug = interactive(),
       swaggerCallback = getOption('plumber.swagger.url', NULL)
     ) {
+      if (isTRUE(private$disable)) {return()}
+
       port <- findPort(port)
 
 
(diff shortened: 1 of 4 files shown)


R/plumber.R
```diff
@@ -288,6 +290,8 @@ plumber <- R6Class(
       debug = interactive(),
       swaggerCallback = getOption('plumber.swagger.url', NULL)
     ) {
+      if (isTRUE(private$disable)) {return()}
```
schloerke, line 293: @blairj09  

If trying to run a plumber router while plumbing a file, would you expect it to 
* Skip over the run method
* Present a message / warning
* Throw an error
?

My thought is to throw. This will strongly discourage this _bad_ coding pattern.

blairj09: Agreed

meztez: Let's throw then

schloerke: Can we also change the flag name to something a little longer? `$disable_run` ?

Thank you, @meztez !

R/plumber.R
```diff
@@ -288,6 +290,10 @@ plumber <- R6Class(
       debug = interactive(),
       swaggerCallback = getOption('plumber.swagger.url', NULL)
     ) {
+      if (isTRUE(private$disable_run)) {
+        stop("Plumber router run method called from a router modifier expression (@plumber)")
```
schloerke, line 294:
```suggestion
        stop("Plumber router `$run()` method should not be called while `plumb()`ing a file")
```

tests/testthat/test-router-modifier.R
```diff
@@ -1,6 +1,9 @@
 context("Router modifier (@plumber tag)")
 
-test_that("router modifier tag works correctly", {
+test_that("router modifier works, run does nothing", {
+  expect_error(
+    plumber$new(test_path("files/router-modifier-run.R")),
+    "Plumber router run method called")
```
schloerke, line 6:
```suggestion
    "method should not be called while")
```

R/plumber.R
```diff
@@ -233,6 +233,8 @@ plumber <- R6Class(
       if (!is.null(file)){
         private$lines <- readUTF8(file)
         private$parsed <- parseUTF8(file)
+        private$disable_run <- TRUE
+        on.exit(private$disable_run <- FALSE, add = TRUE)
```
schloerke, line 237:
```suggestion
        on.exit({
          private$disable_run <- FALSE
        }, add = TRUE)
```

PR #578, opened 2020-06-30
Upload an arbitrary number of files

### LINKED ISSUE #575: Support arbitrary number of files upload
Modify OpenAPI Specification generation.
Enable isArray on file plumber type
Modify parse_multi to combine args sharing the same name

Note:
Swagger UI does not support this yet
Rapidoc UI does
Redoc does not have a try feature

### LINKED ISSUE #582: Can't writeBin without removing filename attribute when using a file parameter
writeBin does not like the raw vector to have attributes (in this case filename)

```R
#* @apiTitle Upload test

#* @param pdf:file
#* @post /upload
function(pdf) {
  f <- tempfile()
  on.exit(unlink(f), add = TRUE)
  writeBin(pdf, tempfile())
  class(pdf)
}

#* @param pdf:file
#* @post /upload2
function(pdf) {
  f <- tempfile()
  on.exit(unlink(f), add = TRUE)
  writeBin(as.raw(pdf), tempfile())
  class(pdf)
}

#* @param pdf:file
#* @post /upload3
function(pdf) {
  f <- tempfile()
  on.exit(unlink(f), add = TRUE)
  attr(pdf, "filename") <- NULL
  writeBin(pdf, tempfile())
  class(pdf)
}

```

upload : Error in writeBin(pdf, tempfile()): can only write vector objects
upload2 and upload3 : ok

Should we fix or document use case?

If fix:
  a) remove attributes
  b) return a named list of bytes list(filename = bytesVector); comes with testing what filename constitute a valid R name.
  c) other option that comes down to where does filename should live during execution

meztez:
Fixes #575 
Fixes #582  

Issue #575, should probably have been there in #550.

Swagger UI does not support it, Rapidoc does, it will work from a CURL command

PR task list:
- [NA] Update NEWS
- [X] Add tests


R/parse-body.R
```diff
@@ -191,12 +191,22 @@ parser_multi <- function() {
...
     })
+    # combine together args that share the same name
+    keys <- unique(names(args))
+    for (k in keys) {
+      m <- which(names(args)==k)
+      if (isTRUE(length(m) > 1)) {
+        args[[m[1]]] <- unname(args[m])
+        args[m[-1]] <- NULL
+      }
+    }
```
schloerke, line 208: Can this use the query string "same name" combiner?  Let's make that function a helper method that can be used here.

meztez: Not in this form it is at the moment. It will have to be reworked. Also I,m worried about large copies of object in memory.

schloerke: > Not in the form it is at the moment. It will have to be reworked. Also I,m worried about large copies of object in memory.

At least in r >= 3.5 (I believe), this should not be a big concern. 

See examples of `pryr::object_size()`

```r
pryr::object_size(ggplot2::diamonds)
#> 3.46 MB
pryr::object_size(list(ggplot2::diamonds, ggplot2::diamonds, ggplot2::diamonds))
#> 3.46 MB

```


> I think it would have to be two different code path as unlisting a list of raw bytes just gives you a large raw vector.


Fair. I wouldn't mind adding a `call_unlist = TRUE` parameter to the helper method.  

The part I'm concerned about is how keys are removed from the args and how they are collected.  The query string method has been tested and I'd like to leverage it if we can.

--------------

Something like...

```r
combine_keys <- function(obj, call_unlist = TRUE) {
  keys <- names(obj)
  vals <- unname(obj)
  unique_keys <- unique(keys)

  if (length(unique_keys) == length(keys)) {
    return(obj)
  }


  cleanup_item <- 
    if (isTRUE(call_unlist)) {
      function(x) { 
        unname(unlist(x))
      }
    } else {
      unname
    }

  vals <-
    if (length(unique_keys) > 150) {
      lapply(split(vals, keys), function(items) {
        cleanup_item(items)
      })
    } else {
      # n < 150
      lapply(unique_keys, function(key) {
        cleanup_item(vals[keys == key])
      })
    }
  names(vals) <- unique_keys
  vals
}
```

```r
test_val <- replicate(5, replicate(10, charToRaw(sample(letters, 1))), simplify = FALSE) %>% setNames(c("a", "b", "a", "b", "a")) %>% print()
#> $a
#>  [1] 6f 66 6b 6d 73 79 63 67 66 6f
#> 
#> $b
#>  [1] 6a 62 63 6d 62 76 71 78 64 78
#> 
#> $a
#>  [1] 61 62 65 70 6b 67 7a 62 63 73
#> 
#> $b
#>  [1] 62 74 74 6b 64 6a 68 6b 75 74
#> 
#> $a
#>  [1] 64 67 6a 6d 75 72 76 77 6a 67
#> 
#> 
combine_keys(test_val, TRUE)
#> $a
#>  [1] 6f 66 6b 6d 73 79 63 67 66 6f 61 62 65 70 6b 67 7a 62 63 73 64 67 6a 6d 75 72 76 77 6a 67
#> 
#> $b
#>  [1] 6a 62 63 6d 62 76 71 78 64 78 62 74 74 6b 64 6a 68 6b 75 74
#> 
#> 
combine_keys(test_val, FALSE)
#> $a
#> $a[[1]]
#>  [1] 6f 66 6b 6d 73 79 63 67 66 6f
#> 
#> $a[[2]]
#>  [1] 61 62 65 70 6b 67 7a 62 63 73
#> 
#> $a[[3]]
#>  [1] 64 67 6a 6d 75 72 76 77 6a 67
#> 
#> 
#> $b
#> $b[[1]]
#>  [1] 6a 62 63 6d 62 76 71 78 64 78
#> 
#> $b[[2]]
#>  [1] 62 74 74 6b 64 6a 68 6b 75 74
```

meztez: Yeah I don't mind doing that somewhere else if it handle memory, I'll fiddle with it. and append this one.

meztez: What do you think about two separate code path? keeping the lapply transform the args structure by listing every element. Even those that are not duplicate when a duplicate is present in the list.

```r
sample1 <- list(a = women, b = 1:4, c = mtcars)
sample2 <- list(a = women, b = 1:4, a = mtcars, c=mtcars)
expect_equal(combine_keys(sample1, FALSE), sample1)
# This one will fail
expect_equal(combine_keys(sample2, FALSE), list(a = list(women, mtcars), b = 1:4, c=mtcars))

# It would return
# list(a = list(women, mtcars), b = list(1:4), c=list(mtcars))
```

schloerke: This would only apply for `parser_multi`?  I think having each key point to a list of items is a great idea.

Makes everything consistent to parse. Good catch!

Yes. Let's makes it return like:
> ```
> # list(a = list(women, mtcars), b = list(1:4), c=list(mtcars))
> ```

meztez: I meant adding the edge case to the combine function.
You can check the PR files now.

Returning always a list for each key would make behavior inconsistent.

```R
#* @param dt:df
function(dt = mtcars) {
   class(dt)
   # data.frame
}

#* @param dt:df
#* @param files:[file]
function(dt = mtcars, files) {
   class(dt)
   # list
   class(dt[[1]])
   # data.frame
   class(files)
   # list
}
```

meztez: > Makes everything consistent to parse. Good catch!
You have a point there, if you send in only one file to an endpoint that can accept multiple files, you would not receive a list of file. Back to drawing board.

meztez:
After more thinking:

Removed the notion of `arraySupport` in api types, all type support array.

I think cohercing `list(nm=arg1` and `nm=arg2)` to `list(nm=list(arg1, arg2))` when `arg1` and `arg2` share the same name is ok in the case where they are themselves list. It should not impact args that do not share the same name. We should maintain their original structure.

It could be documented as such.

R/parse-body.R
```diff
@@ -29,8 +29,8 @@ parseRaw <- function(toparse) {
 }
 
 parserPicker <- function(content_type, first_byte, filename = NULL) {
-  #fast default to json when first byte is 7b (ascii {)
-  if (first_byte == as.raw(123L)) {
+  #fast default to json when first byte is 7b or 5b (ascii { or [)
```
meztez, line 32: https://stackoverflow.com/questions/5034444/can-json-start-with

https://i.stack.imgur.com/wAM3m.gif

meztez:
ready for review

R/parse-body.R
schloerke: good catch!

R/parse-body.R
```diff
@@ -128,7 +128,8 @@ parser_json <- function() {
     charset <- getCharacterSet(content_type)
     value <- rawToChar(value)
     Encoding(value) <- charset
-    safeFromJSON(value)
+    args <- safeFromJSON(value)
+    combine_keys(args, FALSE)
```
schloerke, line 132: I'm leaning towards saying we shouldn't do this here. There's nothing wrong with returning multiple keys with the same name. I would image the route should handle it.

```r
jsonlite::parse_json('{"a":1,"a":2}')
# $a
# [1] 1
# 
# $a
# [1] 2
```

R/parse-body.R
```diff
@@ -191,12 +192,14 @@ parser_multi <- function() {
...
     # content-type detection
-    lapply(toparse, function(x) {
+    args <- lapply(toparse, function(x) {
       if (!is.null(x$filename)) {
         x$content_type <- getContentType(tools::file_ext(x$filename))
       }
       parseRaw(x)
     })
+    # combine together args that share the same name
+    combine_keys(args, FALSE)
```
schloerke, line 202: Same as the json one above. Seems odd that a multi part would combine things and return a different shape only if the names are duplicated. Nothing wrong having a list with two items of the same name.

schloerke: Or is this where the magic is happening to be able to return `list(file1 = val1, file2 = val2, ....)`?

R/parse-body.R
meztez: It won't because only the first one will be matched there
https://github.com/rstudio/plumber/blob/20427b16eefed4a16952d22346ca41021d09cfce/R/plumber-step.R#L125

Also this gives an error:
```r
> a <- function(b, b) {}
Error: repeated formal argument 'b' on line 1
```

R/parse-body.R
meztez: > 
> 
> Or is this where the magic is happening to be able to return `list(file1 = val1, file2 = val2, ....)`?

Yes, for an arbitrary number of files. I think we discussed that all files should be returned in the form list(filename = raw).

It ends up in the args list as args = list(arg1 = 1, argfile1 = list(filename1 = raw1, ...))

R/parse-body.R
meztez: Are you saying we should do it somewhere else where we match args with the route?

R/parse-body.R
meztez: > If the value of a form field is a set of files rather than a single file, that value MUST be transmitted by supplying each in a separate part, but all with the same name parameter. 

https://tools.ietf.org/id/draft-ietf-appsawg-multipart-form-data-01.html

This is specific to the multipart content-type. I removed it from the json parser.

PR #583, opened 2020-07-03
add check before guessing content-type

meztez:
In reference to : https://community.rstudio.com/t/uploading-tsv-files-to-plumber-api/71705/3

PR task list:
- [NA] Update NEWS
- [X] Add tests


R/parse-body.R
```diff
@@ -192,7 +192,7 @@ parser_multi <- function() {
     toparse <- parse_multipart(value, boundary)
     # content-type detection
     lapply(toparse, function(x) {
-      if (!is.null(x$filename)) {
+      if (x$content_type %in% c(NULL, "application/octet-stream") && !is.null(x$filename)) {
         x$content_type <- getContentType(tools::file_ext(x$filename))
       }
```
schloerke, line 197: `c(NULL, "application/octet-stream") == c("application/octet-stream")`

```suggestion
      if (is.null(x$content_type) || isTRUE(x$content_type == "application/octet-stream")) {
        if (!is.null(x$filename)) {
          x$content_type <- getContentType(tools::file_ext(x$filename))
        }
      }
```

meztez:
Do you know of a way to prevent git from messing up a binary file that happens to be just text? It seems to strip the last two bytes at the end of the files.

schloerke:
From https://stackoverflow.com/a/39461324/591574... maybe add a `*.bin binary` line to `./gitattributes` ?

meztez:
Thanks. I'll go read a book for a bit. Should have duckducked that. Time for a break.

meztez:
Thanks @schloerke

PR #584, opened 2020-07-03
Restructure body parsers to use content type. Add many parsers

### LINKED ISSUE #586: Make body parsers opt in
Related https://github.com/rstudio/plumber/pull/584

Goals
* Users must provide parsers that they accept.
* Each parser must be defined by a single line containing 
* Each parsing function should have a name when registered.
* Allow for functions to be passed in directly.
* Add docs for rds that `rds` parsing should only be done from a trusted source. Do not accept `rds` blindly.
* Should a helper method be added for the return type of each parser? `function(mime_type, parser){}`

`/upload_white_list` is an example of how we could define 4 parsers for this route. By default, `@parse json` will be used.

`/upload_anything` should be allowed to use any parser that is registered. (Not recommended)

```
#* Parse Example
#*
#* @parse csv
#* @parse rds
#* @parse parser_json(simplifyVectors = FALSE) # returns a mime type and function
#* @parse plumber::parse_tsv(delim = "\t")
#* @post /upload_white_list
function(...) {
  # do stuff
  my_objects <- list(...)

  "parsed"
}

#* Opt in to parsing everything
#* @parse anything
#* @post /upload_anything
function(...) {
  # do stuff
  my_objects <- list(...)

  "parsed"
}


## Example
parse_xml <- function(...) {
  list(
    # mime type regex
    mime_type = "application/xml-*", # matches any thing that starts with `xml`
    parser = function(value, content_type = NULL, filename = NULL, ...) {
      # parse the xml value
    }
  )
}
```

cc @jcheng5

schloerke:
Fixes https://github.com/rstudio/plumber/pull/583
Fixes #586

Reference: https://community.rstudio.com/t/uploading-tsv-files-to-plumber-api/71705/3

Changes:
* Restructure parser list to use the pattern as the key and parser as the value. 
	* Allows the same parser to exist under different content type values.
	* Provide duplicate parsers for commonly seen content types
* Allow for arguments to be supplied to parsers
* Add tsv parser
* Add csv parser
* If a content-type of `application/octet` is provided, inspect the file name to see what it's mime-type is and use that as the content type value.
	* Allows for RDS files to be shipped using `application/octet`, but processed using `application/rds` which maps to `parser_rds()`
* Use other parsers to reduce copy/pasta
* Move the JSON first inspection after the content type has a chance to be used. Avoids possible situations where the plain text starts with a `{`, but is truly plain text and not JSON.
* Rename (new for next release) functions to snake case. (`add_parser()`, `parse_body()`, ...)

PR task list:
- [NA] Update NEWS
- [x] Add tests

-------------

@meztez Can you take a look at this PR.


R/parse-body.R
```diff
@@ -192,54 +243,56 @@ parser_multi <- function() {
...
   }
-}
 
+  # only one form of these parsers
+  add_parser("application/x-www-form-urlencoded", parser_query())
+  add_parser("application/rds",                   parser_rds())
+  add_parser("multipart/form-data",               parser_multi())
+  add_parser("application/octet",                 parser_octet())
 
+  add_parser("text/plain", parser_text())
```
meztez, line 290: shoud we add text/html? It will eventually be picked up by the picker in the next if check anyway?

schloerke: My thought is to leave it open as it isn't a _solution_ to `text/html`.

I'd expect `text/html` or `application/html` to turn into tag objects via something like https://rvest.tidyverse.org/reference/html.html . I'd like to not make that design decision for now.

meztez:
This is masterfully crafted code. Beautiful. Loves the generalization where parser reuse other parser. Last time I've seen that great of a collab was Snoop & Dre.

Seriously. Great stuff.

R/parse-body.R
meztez: Makes perfect sense.

schloerke:
😆😆 Made my day!

R/parse-body.R
```diff
@@ -192,54 +243,56 @@ parser_multi <- function() {
...
-    Encoding(value) <- charset
-    yaml::yaml.load(value)
   }
-}
 
+  # only one form of these parsers
+  add_parser("application/x-www-form-urlencoded", parser_query())
+  add_parser("application/rds",                   parser_rds())
+  add_parser("multipart/form-data",               parser_multi())
+  add_parser("application/octet",                 parser_octet())
```
schloerke, line 288:
```suggestion
  add_parser("application/octet",                 parser_octet())
  add_parser("application/octet-stream",          parser_octet())
```

R/parse-body.R
```diff
@@ -192,54 +243,57 @@ parser_multi <- function() {
...
```
schloerke, line 288: Remove invalid content type

```suggestion
```

Was originally there from carry over when using regex matching

R/parse-body.R
```diff
@@ -16,172 +16,223 @@ postBodyFilter <- function(req){
...
 
+#' @describeIn parsers YAML parser
+#' @export
+parser_yaml <- function(...) {
+  parser_text(function(val) {
+    if (!requireNamespace("yaml", quietly = TRUE)) {
+      stop("yaml must be installed for the yaml parser to work")
+    }
+    yaml::yaml.load(val, ...)
+  })
```
schloerke, line 179:
```suggestion
parser_yaml <- function(..., eval.expr = FALSE) {
  parser_text(function(val) {
    if (!requireNamespace("yaml", quietly = TRUE)) {
      stop("yaml must be installed for the yaml parser to work")
    }
    yaml::yaml.load(val, ..., eval.expr = eval.expr)
  })
```

R/json.R
```diff
@@ -1,8 +1,8 @@
 #' @importFrom jsonlite validate fromJSON toJSON
 #' @noRd
-safeFromJSON <- function(txt) {
+safeFromJSON <- function(txt, ...) {
   if (!validate(txt)) {
     stop("Argument 'txt' is not a valid JSON string.")
   }
-  fromJSON(txt)
+  fromJSON(txt, ...)
```
schloerke, line 7: Switch to `jsonlite::parse_json` here and default the parser to `simplifyVector = TRUE`

meztez: still relevant?

schloerke: Yes

`jsonlite::parse_json` assumes the input is a string to be parsed, not a string or file.  However, `jsonlite::parse_json` changes the default behavior of `simplifyVectors` to `FALSE` when compared `jsonlite::fromJSON`. It should be set to `TRUE` for backwards compatibility.

meztez: Done in #587

R/parse-body.R
```diff
@@ -16,230 +12,377 @@ postBodyFilter <- function(req){
...
+#' @export
+#' @describeIn add_parser List currently registered parsers
+list_parsers <- function() {
+  .globals$parsers
+}
 
+#' @describeIn add_parser Select from global parsers and create
+#' a formatted parsers list for programmatic use.
+#' @export
+select_parsers <- function(alias = character()) {
```
schloerke, line 202: Going to make this required arg

```suggestion
select_parsers <- function(alias) {
```

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+    tolower(toparse$content_type),
+    toparse$value[1],
+    toparse$filename,
+    toparse$parsers)
+  if (!is.null(parser)) {
+   return(do.call(parser, toparse))
+  } else {
+    message("No suitable parser found to handle request body type ", toparse$content_type, ".")
+    return(list())
+  }
```
cpsievert, line 42:
```suggestion
  parser <- parser_picker(
    # Lower case content_type for parser matching
    tolower(toparse$content_type),
    toparse$value[1],
    toparse$filename,
    toparse$parsers
  )
  if (is.null(parser)) {
    message("No suitable parser found to handle request body type ", toparse$content_type, ".")
    return(list())
  }
  do.call(parser, toparse)
```

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
-  addParser("text", parser_text, "text/")
-  addParser("rds", parser_rds, "application/rds")
-  addParser("multi", parser_multi, "multipart/form-data")
-  addParser("octet", parser_octet, "application/octet")
-  addParser("yaml", parser_yaml, "application/x-yaml")
+add_parsers_onLoad <- function() {
+
+  # parser alias names for plumbing
+  add_parser("csv", parser_csv, fixed = c("application/csv", "application/x-csv", "text/csv", "text/x-csv"))
+  add_parser("json", parser_json, fixed = c("application/json", "text/json"), regex = "json$", shortname = "json")
```
cpsievert, line 379:
```suggestion
  add_parser("json", parser_json, fixed = c("application/json", "text/json"), shortname = "json")
```

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+  if (!is.null(parser)) {
+    return(parser)
   }
-}
 
+  fpm <- stri_detect_regex(
+    content_type,
+    names(parsers$regex),
+    max_count = 1)
+  fpm[is.na(fpm)] <- FALSE
```
cpsievert, line 76: Both `any()` and `which()` are `NA` safe

```suggestion
  fpm <- stri_detect_regex(
    content_type,
    names(parsers$regex),
    max_count = 1
  )
```

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
 
+  # return known parser (first regex pattern match)
+  if (any(fpm)) {
+    return(parsers$regex[[which(fpm)]])
```
cpsievert, line 80:
```suggestion
    return(parsers$regex[[which(fpm)[1]]])
```

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+list_parsers <- function() {
+  .globals$parsers
+}
 
-#' QUERY STRING
-#' @rdname parsers
+#' @describeIn add_parser Select from global parsers and create
+#' a formatted parsers list for programmatic use.
+#' @export
+select_parsers <- function(alias) {
```
cpsievert, line 202:
```suggestion
get_parsers <- function(alias) {
```

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+          parsers$regex <- create_list(regex)
+        }
+
+        return(parsers)
+
+    }
+
+  } else {
+
+    parsers_list = parser
```
cpsievert, line 184: Remove and add a check above

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
-addParser <- function(name, parser, pattern = NULL) {
-  if (is.null(.globals$parsers)) {
-    .globals$parsers <- list()
-  }
-  if (!is.null(.globals$parsers$func[[name]])) {
-    stop("Already have a parser by the name of ", name)
-  }
-  if (is.null(pattern)) {
-    pattern <- paste0("application/", name)
+add_parser <- function(alias, parser, fixed = NULL, regex = NULL, shortname = NULL, verbose = TRUE) {
```
cpsievert, line 141: Exported function shouldn't include `shortname`?

cpsievert: Should this be named `register_parser()`?

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+    read_fn(tmp)
   }
 }
 
+#' @describeIn parsers CSV parser
+#' @export
+parser_csv <- function(...) {
+  parser_read_file(function(val) {
+    utils::read.csv(val, ...)
+  })
```
cpsievert, line 302:
```suggestion
  parser_read_file(function(tmpfile) {
    utils::read.csv(tmpfile, ...)
  })
```

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+}
+
 
+#' @describeIn parsers TSV parser
+#' @export
+parser_tsv <- function(...) {
+  parser_read_file(function(val) {
+    utils::read.delim(val, ...)
+  })
```
cpsievert, line 311:
```suggestion
  parser_read_file(function(tmpfile) {
    utils::read.delim(tmpfile, ...)
  })
```

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
 
 
-#" MULTI
-#' @rdname parsers
+#' @describeIn parsers RDS parser
+#' @export
+parser_rds <- function(...) {
+  parser_read_file(function(value) {
+    readRDS(value, ...)
+  })
```
cpsievert, line 320:
```suggestion
  parser_read_file(function(tmpfile) {
    readRDS(tmpfile, ...)
  })
```

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+    }, add = TRUE)
     writeBin(value, tmp)
-    readRDS(tmp)
+    read_fn(tmp)
   }
 }
 
+#' @describeIn parsers CSV parser
+#' @export
+parser_csv <- function(...) {
```
cpsievert, line 299: Use readr

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+  parser_read_file(function(val) {
+    utils::read.csv(val, ...)
+  })
+}
+
 
+#' @describeIn parsers TSV parser
+#' @export
+parser_tsv <- function(...) {
```
cpsievert, line 308: Use readr

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+#' @describeIn parsers YAML parser
+#' @export
+parser_yaml <- function(...) {
+  parser_text(function(val) {
+    if (!requireNamespace("yaml", quietly = TRUE)) {
+      stop("yaml must be installed for the yaml parser to work")
+    }
+    yaml::yaml.load(val, ..., eval.expr = FALSE)
+  })
+}
```
cpsievert, line 279:
```suggestion
parser_yaml <- function(...) {
  if (!requireNamespace("yaml", quietly = TRUE)) {
    stop("yaml must be installed for the yaml parser to work")
  }
  parser_text(function(val) {
    yaml::yaml.load(val, ..., eval.expr = FALSE)
  })
}
```

NAMESPACE
```diff
@@ -2,8 +2,8 @@
```
cpsievert, line 2: API:

* `register_parser()`
* `select_parsers()` -> `get_parsers()`
  * Defaults to everything
  *  Get rid of `list_parsers()` (it's just `names(get_parsers())`)

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+#' @describeIn add_parser Select from global parsers and create
+#' a formatted parsers list for programmatic use.
+#' @export
+select_parsers <- function(alias) {
+  parsers <- .globals$parsers[alias]
+  # remove to avoid infinite recursion
+  parsers$all <- NULL
+  return(
+    # Lambda function to get each parser `p()` list
+    Reduce(function(l, p) {utils::modifyList(l, p())}, parsers, init = list())
```
cpsievert, line 208: `p()` should accept arguments

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+#' @describeIn add_parser List currently registered parsers
+list_parsers <- function() {
+  .globals$parsers
+}
 
-#' QUERY STRING
-#' @rdname parsers
+#' @describeIn add_parser Select from global parsers and create
+#' a formatted parsers list for programmatic use.
+#' @export
```
meztez, line 201:
```r
#' @param aliases a vector of alias
```

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
+get_parsers <- function(aliases) {
+  parsers <- .globals$parsers[alias]
```
meztez, line 203: aliases instead of alias

R/parse-body.R
meztez: if you want to accept argument switch to a do.call, not all parsers accept arguments.

R/parse-body.R
```diff
@@ -16,230 +12,378 @@ postBodyFilter <- function(req){
...
-#' @rdname parsers
+#' @describeIn parsers All parsers (For internal use only)
 #' @export
-parser_octet <- function() {
-  function(value, filename = NULL, ...) {
-    attr(value, "filename") <- filename
-    return(value)
-  }
+parser_all <- function() {
+  select_parsers(names(.globals$parsers))
```
meztez, line 365: replace all : select_parsers > get_parsers

R/plumber-step.R
```diff
@@ -195,6 +198,9 @@ PlumberEndpoint <- R6Class(
       if (!missing(serializer) && !is.null(serializer)){
         self$serializer <- serializer
       }
+      if (!missing(parsers) && !is.null(parsers)){
```
meztez, line 201:
```r
if (is.character(parsers)) {
  parsers <- get_parsers(parsers)
}
```
plumb-block creates its own list because it can use arguments

R/plumber.R
```diff
@@ -490,8 +493,11 @@ plumber <- R6Class(
         if (missing(serializer)){
           serializer <- private$serializer
         }
+        if (missing(parsers)){
```
meztez, line 496:
```r
parsers <- get_parsers(private$parsers)
```

R/plumber-step.R
schloerke: ￼ correct. I am converting plum block to produce a named list of alias to arguments. 

This will allow plum block and programmatic end points to create parsers with arguments￼￼. 

------

If you don't mind, I will tag you tomorrow morning to review the changes￼

schloerke:
@meztez Do you mind taking a look through? 🙌 Thank you!

Notable differences:
* parse block constructs a named list of alias to list of arguments
* `select_parsers()` -> `make_parser()`
* `make_parser()` now accepts many different input types: `TRUE`, character vector, named list of alias to list of arguments
* `plumber private$parsers` -> `plumber private$default_parsers`

R/parse-body.R
```diff
@@ -16,230 +12,461 @@ postBodyFilter <- function(req){
...
-    charset <- getCharacterSet(content_type)
-    value <- rawToChar(value)
-    Encoding(value) <- charset
-    safeFromJSON(value)
+make_parser <- function(aliases) {
+  if (inherits(aliases, "plumber_parsed_parsers")) {
+    return(aliases)
+  }
+  if (isTRUE(aliases)) {
+    # use all available parsers except ("none")
```
meztez, line 213: except ("all", "none")

R/parse-body.R
```diff
@@ -16,230 +12,461 @@ postBodyFilter <- function(req){
...
+registered_parsers <- function() {
+  names(.globals$parsers)
+}
 
-#' JSON
-#' @rdname parsers
+#' @describeIn register_parser Select from global parsers and create a combined parser list for programmatic use.
+#' @param aliases Can be one of:
+#'   * A character vector of `alias` names.
+#'   * A named `list()` whose keysare `alias` names and values are arguments to be applied with [do.call()]
```
meztez, line 204: keysare?

R/parse-body.R
```diff
@@ -16,230 +12,461 @@ postBodyFilter <- function(req){
...
-
-#' OCTET
-#' @rdname parsers
+#' @describeIn parsers Enable all parsers. Not recommended due to security concerns.
 #' @export
-parser_octet <- function() {
-  function(value, filename = NULL, ...) {
-    attr(value, "filename") <- filename
-    return(value)
+parser_all <- function() {
```
meztez, line 436: I'm not sure you would need an all and a none parser since it could be all managed from make_parsers.
Also parser_all is broken

make_parsers("none") or make_parsers(character())
make_parsers("all")

R/parse-body.R
```diff
@@ -16,230 +12,461 @@ postBodyFilter <- function(req){
...
+  # parser alias names for plumbing
+  register_parser("csv", parser_csv, fixed = c("application/csv", "application/x-csv", "text/csv", "text/x-csv"))
+  register_parser("json", parser_json, fixed = c("application/json", "text/json"))
+  register_parser("multi", parser_multi, fixed = "multipart/form-data")
+  register_parser("octet", parser_octet, fixed = "application/octet-stream")
+  register_parser("query", parser_query, fixed = "application/x-www-form-urlencoded")
+  register_parser("rds", parser_rds, fixed = "application/rds")
+  register_parser("text", parser_text, fixed = "text/plain", regex = "^text/")
+  register_parser("tsv", parser_tsv, fixed = c("application/tab-separated-values", "text/tab-separated-values"))
+  register_parser("yaml", parser_yaml, fixed = c("application/yaml", "application/x-yaml", "text/yaml", "text/x-yaml"))
```
meztez, line 469: Maybe we should not register yaml parser since it requires yaml library. Using the all parser could lead to a yaml library error.

R/plumb-block.R
```diff
@@ -130,7 +131,7 @@ plumbBlock <- function(lineNum, file){
       }
 
       if (!is.na(s) && !s %in% names(.globals$serializers)){
```
meztez, line 133: consider one of the following
 - adding "all", "none" to the check here
 - removing the check here and letting the make_parsers inner check do it's job
 - make the check a function and use the same for make_parsers and here

R/plumber.R
```diff
@@ -733,7 +740,17 @@ plumber <- R6Class(
           if (!is.null(h$serializer)) {
             res$serializer <- h$serializer
           }
-          req$args <- c(h$getPathParams(path), req$args)
+          parsers <-
+            if (!is.null(h$parsers)) {
+              h$parsers
+            } else {
+              make_parser(private$default_parsers)
```
meztez, line 747: making parsers on each call to the API from a character vector could lower performance, consider pre-making these.

R/parse-body.R
schloerke:
```suggestion
#'   * A named `list()` whose keys are `alias` names and values are arguments to be applied with [do.call()]
```

R/parse-body.R
schloerke: Moved the library check to inside the parsing function.

R/parse-body.R
schloerke: Correct. Moved the logic inside for the `"all"` case.

I still want to maintain the `parser_none` to be a function.  This maintains the contract that a `parser` returns a function that can parse a value.  The only one that breaks this contract is `"all"`, which is now handled by `make_parser("all")`.

R/plumber.R
schloerke: Having `make_parser(foo)` automatically return if already parsed. So this works: `make_parser(make_parser(foo))`

R/plumb-block.R
schloerke: Added `registered_serializers()`. Going to leave the check against the `registered_parsers()` as `stopOnLine()` gives better context than what `make_parser()` can provide

R/plumber.R
meztez: Just saw you changed it in initialize, it is all good.

schloerke: 😄 It was a good catch! Spot on.

R/plumb-block.R
```diff
@@ -151,6 +152,30 @@ plumbBlock <- function(lineNum, file){
...
+        stopOnLine(lineNum, line, "No @parser specified")
+      }
+
+      if (!parser_alias %in% registered_parsers()){
+        stopOnLine(lineNum, line, paste0("No such @parser registered: ", parser_alias))
+      }
+
+      if (!is.na(parsersMat[1, 4]) && parsersMat[1,4] != ""){
+        # We have an arg to pass in to the parser
+        arg_list <- eval(parse(text=parsersMat[1,4]))
```
meztez, line 168: add envir to 
```r
arg_list <- eval(parse(text=parsersMat[1,4]))
...
arg_list <- eval(parse(text=parsersMat[1,4]), envir)
```

#620

schloerke: Should also prolly surround each `eval` statement in a `tryCatch` with a `stopOnLine` in the error part.

R/parse-body.R
```diff
@@ -16,230 +12,475 @@ postBodyFilter <- function(req){
...
+#' @describeIn parsers Query string parser
+#' @examples
+#' \dontrun{
+#' # Overwrite `text/json` parsing behavior to not allow JSON vectors to be simplified
+#' #* @parser json simplifyVector = FALSE
+#' # Activate `rds` parser in a multipart request
+#' #* @parser multi
+#' #* @parser rds
+#' pr <- plumber$new()
+#' pr$handle("GET", "/upload", function(rds) {rds}, parsers = c(parser_multi(), parser_rds()))
```
cpsievert, line 332:
```suggestion
#' pr$handle("GET", "/upload", function(rds) {rds}, parsers = c("multi", "rds"))
```

R/parse-body.R
```diff
@@ -16,230 +12,475 @@ postBodyFilter <- function(req){
...
 }
 
+#' @describeIn parsers CSV parser
+#' @export
+parser_csv <- function(...) {
+  parser_read_file(function(tmpfile) {
+    if (!requireNamespace("readr", quietly = TRUE)) {
+      stop("`readr` must be installed for `parser_csv` to work")
+    }
+    readr::read_csv(tmpfile, ...)
```
cpsievert, line 397: `file` takes a raw vector, so no need to write to disk

R/parse-body.R
```diff
@@ -16,230 +12,475 @@ postBodyFilter <- function(req){
...
+}
 
-#" MULTI
-#' @rdname parsers
+
+#' @describeIn parsers RDS parser
+#' @export
+parser_rds <- function(...) {
+  parser_read_file(function(tmpfile) {
+    readRDS(tmpfile, ...)
```
cpsievert, line 418: Could maybe use `rawConnection` to avoid writing to disk

R/plumber-step.R
```diff
@@ -176,6 +178,7 @@ PlumberEndpoint <- R6Class(
     #' @param expr endpoint expr
     #' @param envir endpoint environment
     #' @param serializer endpoint serializer
+    #' @param parsers endpoint parsers. If provided, the value will be sent processed by [make_parser()]
```
cpsievert, line 181: Don't export `make_parser()` and move relevant documentation here.

R/parse-body.R
```diff
@@ -16,230 +12,475 @@ postBodyFilter <- function(req){
...
+#'
+#' Functions signature should include `value`, `...` and
+#' possibly `content_type`, `filename`. Other parameters may be provided
+#' if you want to use the headers from [webutils::parse_multipart()].
+#'
+#' Parser function structure is something like below.
 #' ```r
-#' parser <- function() {
-#'   function(value, content_type = "ct", filename, ...) {
+#' parser <- (parser_arguments_here) {
```
cpsievert, line 121:
```suggestion
#' parser <- function(parser_arguments_here) {
```

R/parse-body.R
schloerke: Boo. `readRDS()` does not work with a `rawConnection()`.

meztez:
Pop a can of Dr Pepper, you did it!!!!

schloerke:
😌 Hope to get #598 this afternoon so we can test the two of them together.

PR #587, opened 2020-07-09
Have opt in parsers

### LINKED ISSUE #586: Make body parsers opt in

schloerke:
Fixes #586 

(cc @meztez ; Made an early PR to allow for commenting / discussion)

PR task list:
- [NA] Update NEWS
- [ ] Add tests


meztez:
I'm working on the doc and adding in @parser none and @parser all

R/parse-body.R
```diff
@@ -67,7 +77,7 @@ parser_picker <- function(content_type, first_byte, filename = NULL) {
     return(parsers$query)
```
schloerke, line 77: Do you think the parsers that are hard coded (`json`, `query`, `text`, `octet`) should always be available? (Pulled from `.globals$parsers, rather than `parsers` that's passed in.)

I'm leaning towards yes.  The only one that maybe not to be included is the `json` one. idk. Such as only do the first byte check on the json if the json parser is included. 🤷

meztez: I've only added the one that were available in 0.4.6 (json and query). The other ones are opt-in.

I had them in initially but upon reading your comment, I thought this was the direction you wanted to go.  I lean towards not overriding what the user defined by using the .globals parsers.

R/parse-body.R
```diff
@@ -159,24 +180,34 @@ parser_json <- function(...) {
...
     }
     yaml::yaml.load(val, ...)
-  })
+  })[[1]]
+  return(invisible(
+    list("application/yaml" = parse_func,
+         "application/x-yaml" = parse_func,
+         "text/yaml" = parse_func,
+         "text/x-yaml" = parse_func)
+  ))
```
schloerke, line 210: I'm on the fence as to which approach I'd for `plumber` to use

One side says we should have a fixed set of values. The other side says we should have a regex.

1. Fixed
```r
list(
  content_types = c("application/yaml", "application/x-yaml", "text/yaml", "text/x-yaml"),
  parser = parse_func
)
```
* Pro: O(1) lookup
* Con: Does **not** allow for dynamic values

2. Regex
```r
list(
  content_type_regex = "^(application|text)/(x-)?yaml$",
  parser = parse_func
)
```
* Con: O(n) look up. Slower. Must test each parser `content_type_regex`. (`n` isn't _that_ big though. Maybe up to 5 values regularly?)
* Pro: Allows for dynamic/partial matching content types

----------------------

Typing this out, I think we should change to using a regex and function combination. Since the content type is passed into the parsing function, the function can handle any particular differences at that point.

Should we use a helper function to validate what is being returned? Such as `plumber_parser(content_type_regex, parser)` or `init_parser(content_type_regex, parser)`? (I think `parser()` would be too generic.

Does this sound ok?

R/parse-body.R
schloerke: I like how you've set it up. It's great.  

Trying to figure out if there are some expected behavior that people **always** expect, regardless of what parsers are being used. 

Like `octet`, `query`, and `text` should **always** be available in the hard coded places in this function. I would even be fine if `text` was removed.

I'll update the defaults down below.

schloerke: Ok. I think I've caught up.  

You are correct.  The code above is clever in that it'll return early and possibly return NULL if the endpoint parser is missing.

The function that calls `parser_picker` will then check if a function was returned. Great!

(resolving)

R/plumber.R
```diff
@@ -220,6 +220,8 @@ plumber <- R6Class(
 
       # Initialize
       private$serializer <- serializer_json()
+      # Default parsers to maintain legacy features
+      private$parsers <- c(parser_json(), parser_query())
```
schloerke, line 224:
```suggestion
      private$parsers <- c(
        parser_json(), 
        parser_query(),
        parser_text(),
        parser_octet(),
        parser_multi()
      )
```

* Adding `text` and `octet` as they seem like fair expected behavior given the new capabilities
* Adding `multi` to parse multipart forms.

R/parse-body.R
meztez: I'll test if the way it is set up at the moment cause any performance decay. I agree it is not clean.

meztez: So using a single regex :
```R
microbenchmark::microbenchmark(stri_detect_regex("text/html", "(application|text)/(plain|html)"))
Unit: microseconds
                                                              expr   min
 stri_detect_regex("text/html", "(application|text)/(plain|html)") 7.501
    lq    mean median    uq    max neval
 7.701 9.07805  7.802 8.901 42.101   100
```

O(1) lookup using the current list
```R
> parsers <- parser_all()
> microbenchmark::microbenchmark(parsers$text)
Unit: nanoseconds
         expr min  lq   mean median  uq   max neval
 parsers$text 300 301 514.98    301 400 18601   100
> microbenchmark::microbenchmark(parsers$`application/json`)
Unit: nanoseconds
                       expr min  lq   mean median  uq  max neval
 parsers$`application/json` 100 200 258.98    201 201 5200   100
```

From a performance perspective, I would stick with a non regex solution.

schloerke: @jcheng5 There is a minimum 10x performance penalty for using a regex when validating a parser works for a given content type. As more parsers are added, more parsers would need to be tested, slowing down even further. 

Should we use a slower regex approach to gain regex matching capabilities or should we stay with the faster, white-list approach?

---------

Notes: Units for regex testing are `microseconds`. Units for key/value lookup (`parser_all()`) are `nanoseconds`.

jcheng5: That's still pretty fast. Can you really get away with exact matching? e.g. `application/json; charset=utf-8` is pretty common; the space after the semicolon is optional and the parameter name is case insensitive. I'm not saying you have to support all of that kind of stuff out of the gate, but I have a feeling either you or plumber users will need to deal with it eventually.

meztez: we only match on mime-type, charset is handled inside individual parser.

https://github.com/rstudio/plumber/blob/0d3b795949b892b2f5efadc6abf5d8f59a96f105/R/parse-body.R#L48

jcheng5: Oh OK, that takes care of a lot then. The only other thing I can think of is how XML has subtypes, like `application/atom+xml`, `application/xhtml+xml`. I don't have an opinion which you should choose but, if you choose fixed strings, I'd suggest you have an idea how you would introduce something more dynamic in the future if necessary (without breaking compatibility). Which maybe argues for the helper function approach.

meztez: I will check but I think all headers are lower cased when they reach us.
I also believe header value to be case sensitive per. Only the name is case insensitive. 
One exception is charset https://tools.ietf.org/html/rfc2616#section-3.4 but we do not use it. 

https://stackoverflow.com/questions/5258977/are-http-headers-case-sensitive
https://www.w3.org/Protocols/rfc2616/rfc2616-sec4.html#sec4.2

meztez: > Oh OK, that takes care of a lot then. The only other thing I can think of is how XML has subtypes, like `application/atom+xml`, `application/xhtml+xml`. I don't have an opinion which you should choose but, if you choose fixed strings, I'd suggest you have an idea how you would introduce something more dynamic in the future if necessary (without breaking compatibility). Which maybe argues for the helper function approach.

I'll see if we can up with something cuter.

For now it would be something like
```R
parser_xml <- function() {
  f <- parser_text(mockxmlparserfunction)
  return(list(invisible(
    "application/atom+xml" = f,
    "application/xhtml+xml" = f,
    "application/xml" = f,
    "application/xhtml" = f
  )))
add_parser("xml", parser_xml)
```

Then you could it 
```R
#* @parser xml
```

schloerke: If we had `parsers` be the list of endpoint parsers that contain fixed content types and/or a regex, we could iterate through them.



```
for (parser_info in parsers) {
  parser_content_types <- parser_info$content_types
  if (!is.null(parser_content_types)) {
    if (content_type %in% parser_content_types) {
      return(parser_info$parser)
    } 
  }
  parser_content_type_regex <- parser_info$content_type_regex
  if (!is.null(parser_content_type_regex)) {
    if (stri_detect(content_type, regex = parser_content_type_regex)) {
      return(parser_info$parser)
    }
  }
}
```

Right now, everything could be in the top `parser_content_types` set.  But it would also allow for the regex within xml. 

This would also allow for multiple parsers to have the same content type, so order of the parsers would matter.  

(`parser_text()` could also add a regex for `text/` in addition to the known type of `text/plain`)

meztez: When would you execute this loop?

schloerke: Np! I should be more explicit. Sorry about that

---------------------

I would replace this line with the loop above: 

https://github.com/rstudio/plumber/pull/587/files#diff-82431f6d2dc55b4007eff34ac0a9d103R60

meztez: @schloerke 

I ended up doing something similar but different, but similar. I kept the fixed match, if nothing is found, it does a first regex match. Hit me up on zoom if you want to discuss anything or here.

Ready for review, tests and parsers doc done.

schloerke: @meztez Sent you an email with a link

meztez: Example above now become:
```R
parser_xml <- function() {
  f <- parser_text(mockxmlparserfunction)
  ct <- c("application/xhtml+xml",
          "(xml|xhtml)$")
  return(make_parsers(f, ct))
}
add_parser("xml", parser_xml)
```

meztez: And before you suggest it. Do you think we should put `ct` as a parameter to the parser builder like so:

```R
parser_xml <- function(match_type =  c("application/xhtml+xml", "(xml|xhtml)$")) {
  f <- parser_text(mockxmlparserfunction)
  return(make_parsers(f, match_type))
}
add_parser("xml", parser_xml)
```

This way it could be exposed, better documented and usable directly from the `@parser` tag.

meztez:
I reworked the make_parsers to handle shortname, fixed and regex. After our last discussion, I've updated the parser_picker, and in some places replaced the `c(list(), list(), list()` call with something that does a `Reduce(utils::modifyList, parsers list)` instead. This is two fold.

- It makes sure there is no duplicated regex or fixed pattern and it will give priority to the latest added parsers
- It combines fixed and regex patterns across parsers lists.

I will push in a bit

R/parse-body.R
```diff
@@ -254,45 +333,44 @@ parser_multi <- function() {
...
     })
   }
+  return(make_parsers(f, fixed = "multipart/form-data"))
 }
 
+#' @describeIn parsers All parsers
+#' @export
+parser_all <- function() {
+  parsers <- .globals$parsers
+  parsers$all <- NULL
```
schloerke, line 347: Is this line needed? 

```suggestion
  parsers$all <- NULL
```

R/parse-body.R
```diff
@@ -76,54 +94,92 @@ parser_picker <- function(content_type, first_byte, filename = NULL) {
...
+      names = n
+    )
+  }
+  parsers <- m(shortname)
+  if (length(fixed) > 0) {
+    parsers <- c(parsers, list("fixed" = m(fixed)))
+  }
+  if (length(regex) > 0) {
+    parsers <- c(parsers, list("regex" = m(regex)))
+  }
```
schloerke, line 180: Small cosmetic things...

```suggestion
  create_list <- function(names) {
    stats::setNames(
      replicate(
        length(names),
        parser_function),
      names
    )
  }
  parsers <- create_list(shortname)
  if (length(fixed) > 0) {
    parsers$fixed <- create_list(fixed)
  }
  if (length(regex) > 0) {
    parsers$regex <- create_list(regex)
  }
```

R/parse-body.R
```diff
@@ -76,54 +94,92 @@ parser_picker <- function(content_type, first_byte, filename = NULL) {
...
+list_parsers <- function() {
+  .globals$parsers
+}
+
+#' @export
+#' @describeIn add_parser Make named list. Mapping content-type with parser.
+#' @param parser_function A single functions to map to one or more Content-Type.
+#' @param fixed A character vector of fixed string to be matched against a request Content-Type.
+#' @param regex A character vector of [regex] string to be matched against a request Content-Type.
+#' @param shortname A character value to reference a parser by a shortname.
```
schloerke, line 160:
```suggestion
#' @param shortname A character value to reference a parser by a shortname. (For internal use only)
```

R/parse-body.R
meztez: Yes, otherwise it does a recursive call to the `all` parsers and you end up with a C error stack.

schloerke:
@meztez Thank you very much for doing these changes! You read my mind.

--------------------------

When adding new parsers, there are two required steps: `make_parsers()` and `add_parser()`

Advantage: 
* Content types could be changed by arguments

Disadvantage:
* It is more complicated.


I believe the original reason was to move to `make_parsers()` as we might have functions in the `#* @parser PKG::FUNC()`. This approach has been scrapped in favor of registering and calling: `#* @parser ALIAS list(arg1 = foo)`

----------------

Proposal:
* Move the logic of `make_parsers()` inside `add_parser()`.
* Remove `make_parsers()`

```
parser_csv <- function(...) {
  parser_read_file(function(val) {
    utils::read.csv(val, ...)
  })
}
add_parser(
  parser_csv, 
  alias = "csv", 
  fixed = c("application/csv", "application/x-csv", "text/csv", "text/x-csv"),
  regex = c(),
  shorthand = c()
)
```

Thoughts?

R/parse-body.R
```diff
@@ -254,45 +333,44 @@ parser_multi <- function() {
...
-
-  add_parser("text/plain", parser_text())
-
-
-  # shorthand names for parser_picker
-  add_parser("text", parser_text())
-  add_parser("query", parser_query())
-  add_parser("octet", parser_octet())
-  add_parser("json", parser_json())
+  # shorthand names for parser plumbing
```
schloerke, line 364:
```suggestion
  # parser alias names for plumbing
```

R/parse-body.R
```diff
@@ -76,54 +94,92 @@ parser_picker <- function(content_type, first_byte, filename = NULL) {
...
+  }
+  create_list <- function(names) {
+    stats::setNames(
+      replicate(
+        length(names),
+        parser_function),
+      names
+    )
+  }
+  parsers <- create_list(shortname)
```
schloerke, line 174:
```suggestion
  parsers <- list()
  if (length(shortname) > 0) {
    parsers$shortname <- create_list(shortname)
  }
```

Also
* Remove the check above
* change how they are accessed above in `parser_picker`

meztez:
>     * Move the logic of `make_parsers()` inside `add_parser()`.

Now that you mention it, It seems a lot more logical.

R/parse-body.R
schloerke, line 350:
```suggestion
  # remove to avoid infinite recursion
  parsers$all <- NULL
```

R/parse-body.R
meztez: Would you be ok with 
```r
  if (length(shortname) > 0) {
    parsers$shortname <- parser_function
  }
```

schloerke: As long as I'm able to call `parsers$json` later, works for me! 🙌

meztez:
done

R/parse-body.R
```diff
@@ -152,36 +196,17 @@ list_parsers <- function() {
...
-    parsers$shortname <- create_list(shortname)
-  }
-  if (length(fixed) > 0) {
-    parsers$fixed <- create_list(fixed)
-  }
-  if (length(regex) > 0) {
-    parsers$regex <- create_list(regex)
-  }
-  invisible(parsers)
+select_parsers <- function(alias = character()) {
```
schloerke, line 202: Would prefer to make it a required argument

```suggestion
select_parsers <- function(alias) {
```

schloerke:
Going to add the comment above and merge into the base PR.

R/parse-body.R
schloerke: For some reason, github won't let me merge this. Will merge in #584

PR #611, opened 2020-07-20
Remove TODO flags from docs

blairj09:
This cleans up existing documentation by pulling each TODO item out of the docs and into an associated GH issue. This will help keep the docs clean while we continue to make improvements. It also helps us avoid the docs appearing as incomplete as they do now.

As it stands, I migrated each TODO into a separate issue so that they can be closed at will, rather than creating a giant issue spanning multiple vignettes. However, we could combine all issues into a single one if we wanted to.


vignettes/rendering-output.Rmd
```diff
@@ -17,7 +17,24 @@ knitr::include_graphics("files/images/plumber_output.png")
 
 ## The Response Object {#response-object}
 
-// TODO
+The plumber response object is stored as an environment, much like [the request object](./routing-and-input.html#the-request-object-1). The response object, which is accessible as `res` from within plumber functions, contains the following objects:
+
+Name | Example | Description
+---- | ------- | -----------------------
+`headers` | `list(header = "abc")` | A list of headers to include in the response
```
cpsievert, line 24:
```suggestion
`headers` | `list(header = "abc")` | A list of HTTP headers to include in the response
```

PR #618, opened 2020-07-24
Replace plus sign with space character when parsing a query string

### LINKED ISSUE #617: Space encoding in the GET input query parameter (+ vs %20)
I'm using a framework that automatically encodes spaces in the GET query parameters as "+" symbol. Looks like it isn't supported in Plumber. %20 works fine.

Example:
Plumber.R
#* Get info
#* @param stage:character stage
#* @get /info
function(stage) {
        print(stage)
        list(msg = paste0("The message is: '", stage, "'"))
}

In case of the following request:
http://host/info?stage=foo+bar
it will return "foo+bar" instead of "foo bar"

Related Stack Overflow question:
https://stackoverflow.com/questions/2678551/when-to-encode-space-to-plus-or-20
I think it would be nice to support both "+" and "%20"

meztez:
```r
plumber:::parseQS("?a=2&a=3&c=&&d=et+te")
```

Fixes #617 

PR task list:
- [x] Update NEWS
- [x] Add tests

### DIFF
--- a/tests/testthat/test-querystring.R
+++ b/tests/testthat/test-querystring.R
@@ -7,7 +7,7 @@ test_that("query strings are properly parsed", {
 })
 
 test_that("special characters in query strings are handled properly", {
-  expect_equal(parseQS("?a=1+.#"), list(a="1+.#"))
+  expect_equal(parseQS("?a=1+.#"), list(a="1 .#"))
   expect_equal(parseQS("?a=a%20b"), list(a="a b"))
   expect_equal(parseQS('?a=%2C%2B%2F%3F%25%26'), list(a=",+/?%&"))
 })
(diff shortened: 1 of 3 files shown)


tests/testthat/test-querystring.R
```diff
@@ -7,7 +7,7 @@ test_that("query strings are properly parsed", {
 })
 
 test_that("special characters in query strings are handled properly", {
```
schloerke, line 9: Adding a test to make sure path parameters are not converted to spaces. (Just for reassurance)

```suggestion
test_that("path parameters do not convert + to space", {
  r <- plumber$new(test_path("files/path-params.R"))
  expect_equal(r$route(make_req("GET", "/car/a+b"), PlumberResponse$new()), "a+b")
})

test_that("special characters in query strings are handled properly", {
```

meztez: Good, I did not think of that.

Issue #593, opened 2020-07-14
Port existing `analogsea` / digital ocean code to a new `plumberDigitialOcean` R package


schloerke:
The code on CRAN has been broken for at least 6 months.  Plumber can not update as fast as `analogsea` might want to update. 

This code was originally set up for showing how a toy example of plumber can be deployed to Digital Ocean.  However, this adds extra dependencies and is out of scope for the `plumber` R package.

I propose that 
* `plumber` Digital Ocean code be ported to an external R package (`plumberDigitalOcean`?). It could then be used in the `plumber` documentation on how to host on Digital Ocean.  
* The existing plumber functions should not really exist in outside scripts, so they may be gutted to proved a single message of `"Please visit USER/plumberDigitalOcean to run FUNC()"`. 
  * Maybe paired with message `remotes::install_github("USER/plumberDigitalOcean"); plumberDigitalOcean::FUNC()`

------------------

I would welcome this breaking change.

meztez:
And the answer to that is... yes.

I would even go as far as calling the new package plumberDeployment or something like that and support multiple deployment models.

schloerke:
I like the name `plumberDeployment`!

... want to take a run at it, @meztez ? (My plate is starting to fill up with `rstudio/learnr`.)

meztez:
yes yes, sometimes end of week or next week.

blairj09:
I'm happy to be involved here as well! Although I'll be mostly unavailable for the next month+ 👶 🎉

PR #620, opened 2020-07-28
change envir eval for plumb block

### LINKED ISSUE #619: Plumber comments should be evaluated in the correct environment
This should work:

```r
foo <- TRUE
#' @serializer json list(simplifyVector=foo)
...
```

```r
bar <- list(simplifyVector=TRUE)
#' @serializer json bar
...
```

meztez:
Fixes #619 

PR task list:
- [NA] Update NEWS
- [x] Add tests

- [ ] Do to the same for parsers once merged.

### DIFF
--- a/R/plumb-block.R
+++ b/R/plumb-block.R
@@ -109,7 +109,11 @@ plumbBlock <- function(lineNum, file, envir = parent.frame()){
 
       if (!is.na(serMat[1, 4]) && serMat[1,4] != ""){
         # We have an arg to pass in to the serializer
-        argList <- eval(parse(text=serMat[1,4]), envir)
+        argList <- tryCatch({
+          eval(parse(text=serMat[1,4]), envir)
+        }, error = function(e) {
+          stopOnLine(lineNum, line, e)
+        })
       } else {
         argList <- list()
       }
@@ -139,7 +143,11 @@ plumbBlock <- function(lineNum, file, envir = parent.frame()){
 
       if (shortSerAttr != "") {
         # We have an arg to pass in to the serializer
-        argList <- eval(parse(text=paste0("list", shortSerAttr)), envir)
+        argList <- tryCatch({
+          eval(parse(text=paste0("list", shortSerAttr)), envir)
+        }, error = function(e) {
+          stopOnLine(lineNum, line, e)
+        })
       } else {
         argList <- list()
       }
@@ -254,7 +262,11 @@ evaluateBlock <- function(srcref, file, expr, envir, addEndpoint, addFilter, pr)
         imageArgs <- NULL
         if (!identical(block$imageAttr, "")){
           call <- paste("list", block$imageAttr)
-          imageArgs <- eval(parse(text=call), envir)
+          imageArgs <- tryCatch({
+            eval(parse(text=call), envir)
+          }, error = function(e) {
+            stopOnLine(lineNum, file[lineNum], e)
+          })
         }
 
         if (block$image == "png"){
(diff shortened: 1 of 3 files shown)


meztez:
One thing I noticed is the treatment for the image arguments are evaluated in `evaluateBlock` instead of `plumbBlock`. It also add `list` in front of the arguments.

Could we move the eval to `plumbBlock` so we get a narrower stopOnLine message in case of errors. We could also make `list` optional by appending it to evals for serializers and parsers (in the other PR)

schloerke:
> Could we move the `eval` to `plumbBlock`?

Yes, please! (Feel free to do that in this PR.)

-------------------------------------
-------------------------------------

> The parsing is inconsistent.

Correct.

List of `#' @KIND TYPE list(ARGS)` setups and their equivalent shorthand
```
#' @parser json list(auto_unbox = TRUE)
## no short hand form of @parser

#' @serializer html list()
#' @serializer html
## shorthand for html specifically
#' @html()
#' @html

#' @serializer json list(auto_unbox = TRUE)
## shorthand for json specifically
#' @json(auto_unbox = TRUE)
#' @json

## No long hand form of images
#' @jpeg
#' @jpeg(w=1)
```


Should we have a separate PR for image serializers? (Will be MUCH easier once #584 is merged)

Something like...

* 
```r
register_image_serializer = function(name, content_type, dev_on, dev_off = grDevices::dev.off, verbose = TRUE) {
  .... # check for existing and warn if verbose
  .globals$image_serializers[[name]] <- list(
    content_type = content_type,
    dev_on = dev_on,
    dev_off = dev_off
  )
}
```
* `registered_image_serializer()`
* Export `image_png`, `image_jpeg`, `image_svg`
* Could maybe have an exported `ep_image(endpoint, image_name)` method to facilitate registering the hooks and serializer

Then the situation brings up how should we handle image serializers? Currently, there is only `#' @IMAGE(arg1 = val1, ..)`. Should we do `#' @image jpeg list(arg1 = val1, ..)`?

Does the `list` text seem redundant? Should we allow for `#' @KIND TYPE(ARGS)` rather than only `#' @KIND TYPE list(ARGS)`. Ex: `#' @image jpeg(arg1 = val1, ..)` or `#' @parser json(simplifyVector = TRUE)`

meztez:
I'll move the eval in this PR. Will do a separate PR for plumber syntax once #584 merged.

schloerke:
Thank you!

--------------

Does the proposed image make sense to implement? ...Seems to follow pattern of the serializers and parsers

meztez:
Yes I think the list is redundant, unsure about the parentheses, I quite like them. But the plumbing block need a tidbit more love, between the shortname for serializers, it is a bit hard to follow.

schloerke:
> unsure about the parentheses, I quite like them

Yes, let's keep the parentheses.

> between the shortname for serializers, it is a bit hard to follow

Yes, I'd like to remove all examples of `@json` and `@html` in the docs. IDK if we should soft deprecate it or just leave it there, hidden.

meztez:
I'm thinking about generalizing block parsing syntax so we have a list of block regex instead of whole blobby mess that it is at the moment. I think we already discussed it.

schloerke:
That'd be great. Would be much easier to read. (Don't know how the variables will be handled)

----------

If you're looking for plumber enhancements, I'd gladly take the removal of Digital Ocean code to an external package. (#593) ... would rather not keep those functions if I can help it. 😬😄

meztez:
I thought I had a PR for digital ocean stuff. Maybe I forgot to push.

cpsievert:
Generally agree with https://github.com/rstudio/plumber/pull/620#issuecomment-665089592. Maybe the naming should be `register_device()`,  `device_*()` (for png, jpeg, etc), and `#' @device png (width=640)`

schloerke:
This code is being merged in #623. (Added a news entry)

Closing here.

PR #627, opened 2020-08-03
`register_ui()` should accept arguments, not a list

schloerke:
`register_ui()` should accept arguments, not a list of arguments.

Going from 

```r
arg_list <- list(
  name = name_val,
  index = index_val,
  static = static_val
)
register_ui(arg_list)
```

to

```r
register_ui(
  name = name_val,
  index = index_val,
  static = static_val
)
```

Chose not to add `verbose` or `overwrite` as these should be done by packages and would always be overwriting.

-------------------

PR task list:
- [NA] Update NEWS
- [NA] Add tests

### DIFF
--- a/R/ui.R
+++ b/R/ui.R
@@ -128,17 +128,50 @@ unmount_openapi <- function(pr) {
 }
 
 #' Add UI for plumber to use
-#' @param ui A list of that plumber can use to mount
-#' a UI.
-#' @details [register_ui()] is used by other packages like `swagger`.
+#'
+#' [register_ui()] is used by other packages like `swagger`.
 #' When you load these packages, it calls [register_ui()] to provide a user
 #' interface that can interpret your plumber OpenAPI Specifications.
 #'
 #' @param name Name of the UI
 #' @param index A function that returns the HTML content of the landing page of the UI.
+#'   Parameters (besides `req` and `res`) will be supplied as if it is a regular `GET` route.
+#'   Default parameter values may be used when setting the ui.
+#'   Be sure to see the example below.
 #' @param static A function that returns the path to the static assets (images, javascript, css, fonts) the UI will use.
 #'
 #' @export
+#' @examples
+#' \dontrun{
+#' # Example from the `swagger` R package
+#' register_ui(
+#'   name = "swagger",
+#'   index = function(version = "3", ...) {
+#'     swagger::swagger_spec(
+#'       api_path = paste0(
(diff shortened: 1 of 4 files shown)


R/ui.R
```diff
@@ -134,23 +134,20 @@ unmount_openapi <- function(pr) {
...
-#' \item{static}{A function that returns the path to the static assets (images, javascript, css, fonts) the UI will use.}
-#' }
+#' @param name Name of the UI
+#' @param index A function that returns the HTML content of the landing page of the UI.
+#' @param static A function that returns the path to the static assets (images, javascript, css, fonts) the UI will use.
+#'
 #' @export
 #' @rdname register_ui
-register_ui <- function(ui) {
+register_ui <- function(name, index, static) {
```
cpsievert, line 143: I think you want `static = NULL`?

R/ui.R
```diff
@@ -134,23 +134,20 @@ unmount_openapi <- function(pr) {
...
-#' `ui` list expects the following values
-#' \describe{
-#' \item{name}{Name of the UI.}
-#' \item{index}{A function that returns the HTML content of the landing page of the UI.}
-#' \item{static}{A function that returns the path to the static assets (images, javascript, css, fonts) the UI will use.}
-#' }
+#' @param name Name of the UI
+#' @param index A function that returns the HTML content of the landing page of the UI.
+#' @param static A function that returns the path to the static assets (images, javascript, css, fonts) the UI will use.
+#'
```
cpsievert, line 140: Do we have any documentation for the input arguments to these functions?

meztez: Documentation should be in each UI package, maybe?

https://github.com/meztez/rapidoc/blob/d370734dcfb4772ffa98d7f165c40821cc10136a/R/paths.R#L40

Or are you talking about how they are called from inside plumber?

https://github.com/rstudio/plumber/blob/13f813400062c9a287af4602172a4232b4c6a545/R/ui.R#L161

It will feed of whatever you put in set_ui as extra arguments.

meztez: This whole thing probably requires a vignette in plumberDeploy

schloerke: > Or are you talking about how they are called from inside plumber?

This one. Have added an example and more docs.

-------------------------


@meztez If a vignette is made, I think it should stay in `plumber` as the content is for API UIs, not deployment of plumber APIs.

PR #629, opened 2020-08-03
Update tips-and-tricks.Rmd

### LINKED ISSUE #624: Add `Random Port` trick to Tips & Tricks doc
## Problem

My team has recently started to use Rstudio server deployed on an Azure VM. We've recently run into an issue trying to test a plumber API. When I click the "Run API" button in Rstudio, plumber will serve the API on a randomly selected port between 3k and 10k -- unless I have specified a specific port in the global options. I've been told by our engineers that we can setup rules for port forwarding in the Windows Adaptive Firewall but they have to be specified one at a time. :cry:. I would like to be able to set the range of ports that plumber will select from so that we restrict usage to a smaller range of ports.

## Potential Solution
I think a potential solution would be modify `getRandomPort` to check to see if getOption("plumber.port_range") is set before defaulting to range from 3k to 10k.

meztez:
Fixes #624 

Port range tips and tricks

### DIFF
--- a/vignettes/tips-and-tricks.Rmd
+++ b/vignettes/tips-and-tricks.Rmd
@@ -41,3 +41,28 @@ function(req, res){
 ```
 
 If you run this API locally and then visit the API in a web browser, you'll see your R session switch into debug mode when the request arrives, allowing you to look at the objects contained inside your `req` and `res` objects.
+
+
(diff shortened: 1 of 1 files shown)


schloerke:
🙌 Thank you, @meztez !

PR #641, opened 2020-08-10
Add support for Cookie SameSite attribute

### LINKED ISSUE #640: Add support for `SameSite` Cookie attribute
The ability to set the `SameSite` cookie attribute would be useful, especially with currently ongoing changes to browsers' handling of cookies without a `SameSite` policy set.

Modern browsers are currently rolling out changes setting the default for cookies to `SameSite=Lax`, which prevents cookies from being included in Cross-Origin requests. Without support for setting the `SameSite` attribute to `None`, it will be impossible to use  cookies in CORS requests, which is especially problematic for Cross-Origin APIs that rely on storing data in the Plumber session cookie.

The proposed changes include adding a new optional parameter called `sameSite` to the `sessionCookie()` function and the Response class's `setCookie()` and `removeCookie()` methods. If this parameter is set to a character value, the attribute `SameSite=<value>` will be added to the generated cookie.

chris-dudley:
Adds an optional `sameSite` attribute to the following functions:

- `sessionCookie()`
- `pr_cookie()`
- `res$setCookie()`
- `res$removeCookie()`
- `removeCookieStr`
- `cookieToStr`

If set to a character value `<value>`, the generated cookie will contain a `SameSite=<value>` attribute.

Fixes #640 

Allows the SameSite attribute to be set on cookies. This is especially important for Cross-Origin requests, as in the near future browsers will automatically default to the "SameSite=Lax" policy for cookies without an explicit policy set. This will prevent cookies from being sent on Cross-Origin requests unless the policy is set to "None".

See:
- https://www.chromestatus.com/feature/5088147346030592
- https://www.chromestatus.com/feature/5633521622188032

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -73,6 +73,8 @@ both UIs integration are available from https://github.com/meztez/rapidoc/ and h
 
 ### Minor new features and improvements
 
+* Added support for the `SameSite` Cookie attribute. (#640)
+
 * When calling `include_file()`, the `content_type` is automatically inferred from the file extension if `content_type` is not provided. (#631)
 
 * Added `serializer_feather()` and `parser_feather()` (#626)
--- a/NEWS.md
+++ b/NEWS.md
@@ -73,7 +73,7 @@ both UIs integration are available from https://github.com/meztez/rapidoc/ and h
 
 ### Minor new features and improvements
 
-* Added support for the `SameSite` Cookie attribute. (#640)
+* Added support for the `SameSite` Cookie attribute. (@chris-dudley, #640)
 
 * When calling `include_file()`, the `content_type` is automatically inferred from the file extension if `content_type` is not provided. (#631)
 
--- a/R/session-cookie.R
+++ b/R/session-cookie.R
@@ -43,7 +43,7 @@
 #' @param sameSite A character specifying the SameSite policy to attach to the cookie.
 #'   If specified, one of the following values should be given: "Strict", "Lax", or "None".
 #'   If "None" is specified, then the \code{secure} flag MUST also be set for the modern browsers to
-#'   accept the cookie.
+#'   accept the cookie. An error will be returned if \code{sameSite = "None"} and \code{secure = FALSE}.
 #'   If not specified or a non-character is given, no SameSite policy is attached to the cookie.
 #' @export
 #' @seealso \itemize{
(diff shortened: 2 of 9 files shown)


R/session-cookie.R
```diff
@@ -40,6 +40,11 @@
 #'   Defaults to \code{TRUE}.
 #' @param secure Boolean that adds the \code{Secure} cookie flag.  This should be set
 #'   when the route is eventually delivered over \href{https://en.wikipedia.org/wiki/HTTPS}{HTTPS}.
+#' @param sameSite A character specifying the SameSite policy to attach to the cookie.
+#'   If specified, one of the following values should be given: "Strict", "Lax", or "None".
+#'   If "None" is specified, then the \code{secure} flag MUST also be set for the modern browsers to
+#'   accept the cookie.
+#'   If not specified or a non-character is given, no SameSite policy is attached to the cookie.
```
schloerke, line 47: Can we implement this logic within `sessionCookie`?

Ex:
```r
if (is.character(sameSite)) {
  sameSite <- match.arg(sameSite, c("Strict", "Lax", "None"))
}
if (identical(sameSite, "None")) {
  if (!secure) {
    stop("informative message here")
  }
}
```

chris-dudley: Just added that logic into `sessionCookie`, also updated the documentation for `sessionCookie` and `pr_cookie` to note that an error will be raised if that particular set of parameters is given.

R/session-cookie.R
```diff
@@ -104,7 +110,17 @@ sessionCookie <- function(
   key <- asCookieKey(key)
 
   # force the args to evaluate
-  list(expiration, http, secure)
+  list(expiration, http, secure, sameSite)
+
+  # sanity check the sameSite and secure arguments
+  if (is.character(sameSite)) {
+    sameSite <- match.arg(sameSite, c("Strict", "Lax", "None"))
+  }
```
schloerke, line 118:
```suggestion
  } else {
    sameSite <- FALSE
  }
```

R/session-cookie.R
```diff
@@ -104,7 +110,17 @@ sessionCookie <- function(
...
+  if (identical(sameSite, "None")) {
+    if (!secure) {
+      stop("You must set secure = TRUE when sameSite is set to 'None'. See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie")
```
schloerke, line 121:
```suggestion
      stop("You must set `secure = TRUE` when `sameSite = \"None\"`. See: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie")
```

schloerke:
Thank you @chris-dudley ! Once I get `@jcheng5`'s approval, I'll merge. 😄

R/session-cookie.R
chris-dudley: Done in 605b265

R/session-cookie.R
chris-dudley: Done in 605b265

NEWS.md
```diff
@@ -73,6 +73,8 @@ both UIs integration are available from https://github.com/meztez/rapidoc/ and h
 
 ### Minor new features and improvements
 
+* Added support for the `SameSite` Cookie attribute. (#640)
```
schloerke, line 76:
```suggestion
* Added support for the `SameSite` Cookie attribute. (@chris-dudley, #640)
```

PR #621, opened 2020-07-28
Digital ocean

meztez:
Fixes #539 

Remove digital ocean stuff from docs hosting section and digital ocean function

PR task list:
- [ ] Update NEWS
- [NA] Add tests
- [ ] Update documentation with `devtools::document()`

PR #649, opened 2020-08-12
Defunct DigitalOcean functions

### LINKED ISSUE #593: Port existing `analogsea` / digital ocean code to a new `plumberDigitialOcean` R package

schloerke:
Fixes #593 
Updates of #621

PR task list:
- [x] Update NEWS
- [x] Add tests


R/digital-ocean.R
```diff
@@ -1,470 +1,42 @@
...
-  writeLines(forward, forwardfile)
-
-  analogsea::droplet_upload(droplet, forwardfile, "/etc/nginx/sites-available/plumber-apis/_forward.conf")
-
-  # TODO: add this as a catch()
-  file.remove(forwardfile)
-
-  invisible(droplet)
+#' @rdname digitalocean
+do_forward <- function(...) {
```
cpsievert, line 28: We should be nicer about these deprecations. Specifically, I'm worried about the `library(plumberDeploy); library(plumber)` case

R/digital-ocean.R
```diff
@@ -58,6 +58,10 @@ plumber_deploy_helper <- function(fn_name, args, new_fn_name = fn_name) {
...
   lifecycle::deprecate_warn("1.0.0", cur_fn, new_fn)
 
-  fn <- getFromNamespace(fn_name, "plumberDeploy")
+  fn <- utils::getFromNamespace(fn_name, "plumberDeploy")
   do.call(fn, args)
 }
+
+plumberDeploy_is_installed <- function() {
+    nzchar(system.file(package = "plumberDeploy"))
+}
```
cpsievert, line 67: Maybe it's time to have something like `shiny:::is_available`

PR #665, opened 2020-08-31
Make `postBody` a delayed assignment. Allows for removal of option.

schloerke:
Allows for dropping of the 

* Concern: By doing a delayed assign, it could cause the rook input read location to be reset and the whole file will be read.
  * While normally this is is not a worry, this can now happen much later than the body filter actually happens. (Such as in the middle of the endpoint.).  This would cause the rook read location to be finished and not where it was originally.  There is no way to restore it: https://github.com/jeffreyhorner/Rook/blob/871c797633a7e56919d003e059d81f86b8b251dc/R/Rhttpd.R#L41

PR task list:
- [NA] Update NEWS
- [x] Add tests

--------------

cc @meztez Thought this was a good trick. Add in a delayed assignment for `req$postBody`. So if no one uses it, it is never calculated. If someone touches it, it will be evaluated.


meztez:
@schloerke  Did not know about delayedAssign, great trick. I would not worry too much about Rook pointer reset. By adding both postbodyraw and postbody, I don't see a need for the original rook stream. I haven't been following closely but I like where you are going with this project and the breaking changes you decided to commit to. Great stuff.

schloerke:
@jcheng5 gentle ping

jcheng5 (approved):

PR #680, opened 2020-09-24
Annotations documentation

meztez:
Brief description of the original problem and the approach behind your solution.

Document annotations, fix bugs and conflicting behavior along the way.

Remove openapi 2.0 annotations, update to annotations to openapi 3.0.3

PR task list:
- [x] `@assets` args order
- [x] Endpoint annotations
- [ ] Update NEWS
- [x] Add tests

### DIFF
--- a/R/plumb-globals.R
+++ b/R/plumb-globals.R
@@ -32,37 +32,28 @@ plumbOneGlobal <- function(fields, argument){
            fields$info$termsOfService <- def
          },
          apiContact={
+           if (grepl("^list\\(", def)) {
+             def <- eval(parse(text = def))
+           }
            fields$info$contact <- def
          },
          apiLicense={
+           if (grepl("^list\\(", def)) {
+             def <- eval(parse(text = def))
+           }
            fields$info$license <- def
          },
          apiVersion={
            fields$info$version <- def
          },
-         apiHost={
-           fields$host <- def
-         },
-         apiBasePath={
-           fields$basePath <- def
-         },
-         apiSchemes={
-           fields$schemes <- strsplit(def, split="\\s+")[[1]]
-         },
-         apiConsumes={
-           fields$consumes <- strsplit(def, split="\\s+")[[1]]
-         },
-         apiProduces={
-           fields$produces <- strsplit(def, split="\\s+")[[1]]
-         },
          apiTag={
            tagMat <- stri_match(def, regex="^\\s*(\\w+)\\s+(\\S.+)\\s*$")
            name <- tagMat[1,2]
            description <- tagMat[1,3]
-           if(!is.null(fields$tags) && name %in% fields$tags$name) {
+           if(!is.null(fields$tags) && name %in% sapply(fields$tags, "[[", "name")) {
              stop("Error: '", argument, "' - ","Duplicate tag definition specified.")
            }
-           fields$tags <- rbind(fields$tags,data.frame(name=name, description=description, stringsAsFactors = FALSE))
+           fields$tags <- c(fields$tags, list(list(name=name, description=description)))
          })
   fields
 }
--- a/R/plumb-globals.R
+++ b/R/plumb-globals.R
@@ -4,8 +4,9 @@
 #' @param argument The line (including the plumber comment prefix) to append.
 #'   If this line represents what was once multiple lines, intermediate comment
 #'   prefixes should have been removed.
+#' @param envir An environment where to evaluate parsed expressions
 #' @noRd
-plumbOneGlobal <- function(fields, argument){
+plumbOneGlobal <- function(fields, argument, envir = parent.frame()){
   if (nchar(argument) == 0){
     return(fields)
   }
@@ -33,13 +34,13 @@ plumbOneGlobal <- function(fields, argument){
          },
          apiContact={
            if (grepl("^list\\(", def)) {
-             def <- eval(parse(text = def))
+             def <- eval(parse(text = def), envir)
            }
            fields$info$contact <- def
          },
          apiLicense={
            if (grepl("^list\\(", def)) {
-             def <- eval(parse(text = def))
+             def <- eval(parse(text = def), envir)
            }
            fields$info$license <- def
          },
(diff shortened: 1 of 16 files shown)


R/plumb-globals.R
```diff
@@ -32,37 +32,28 @@ plumbOneGlobal <- function(fields, argument){
...
          apiLicense={
+           if (grepl("^list\\(", def)) {
+             def <- eval(parse(text = def))
+           }
            fields$info$license <- def
          },
          apiVersion={
            fields$info$version <- def
          },
-         apiHost={
```
meztez, line 43: Update to this function were made to match openapi 3.0.3 spec.

R/plumber.R
```diff
@@ -118,7 +118,7 @@ Plumber <- R6Class(
           private$disable_run <- FALSE
         }, add = TRUE)
 
-        for (i in 1:length(private$parsed)){
+        for (i in seq_len(length(private$parsed))) {
```
meztez, line 121: While plumbing an empty file or a file without any block expression, this would create an error since the loop would go from 1:0 instead of not looping.

R/plumber.R
```diff
@@ -817,7 +817,8 @@ Plumber <- R6Class(
       private$docs_info <- list(
         enabled = enabled,
         docs = docs,
-        args = list(...)
+        args = list(...),
+        has_not_been_set = FALSE
```
meztez, line 821: Saw that this was set during router initialization, I think it should be set to FALSE when `setDocs` is called. Otherwise it stays TRUE at all time.

R/pr.R
```diff
@@ -499,21 +499,24 @@ pr_run <- function(pr,
...
 #' pr() %>%
-#'   pr_static("/path", "./my_folder/location") %>%
+#'   pr_static("./my_folder/location", "/path") %>%
 #'   pr_run()
 #' }
 #'
 #' @export
+# TODO: Validate if this makes sense?
+# Arguments path and direc are in reverse order compared to plumber annotation `@assets direc {path}`.
+# Default path is `/public` for annotation `@assets`
```
meztez, line 515: What the comment says

R/plumb-globals.R
```diff
@@ -32,37 +32,28 @@ plumbOneGlobal <- function(fields, argument){
...
-         },
-         apiProduces={
-           fields$produces <- strsplit(def, split="\\s+")[[1]]
-         },
          apiTag={
            tagMat <- stri_match(def, regex="^\\s*(\\w+)\\s+(\\S.+)\\s*$")
            name <- tagMat[1,2]
            description <- tagMat[1,3]
-           if(!is.null(fields$tags) && name %in% fields$tags$name) {
+           if(!is.null(fields$tags) && name %in% sapply(fields$tags, "[[", "name")) {
```
schloerke, line 53: I try to avoid `sapply` at all costs due to the inconsistent simplifying behavior.  If we know it's a vector, we should use `vapply`. If it's a vector, but may not have a `name` in each item, then we can `unlist(lapply())`

```suggestion
           if(!is.null(fields$tags) && name %in% unlist(lapply(fields$tags, "[[", "name"))) {
```

R/plumber.R
schloerke: I'm good with reversing the name / logic. Ex: ~ `vanilla = TRUE`

R/pr.R
schloerke: I think `@assets`'s order is wrong. I'd like to keep the original `pr_static()` setup.  

It might be nice to change the docs to not use `@assets` (due to the confusing order) and instead use....

```r
#' @plumber
function(pr) {
  pr %>%
    pr_static(path, direc)
}
```

R/plumber.R
meztez: Let me rephrase that. `has_not_been_set` is not set anywhere else than initialization. So the value will always be what you set it to when you check it later for the purpose of lifecycle messaging. This boolean value should be reversed when you actually set the doc. This is what I was alluding to.

R/pr.R
meztez: My hypothesis is that the initial `@asset` args order was this way because `path` was optional. It default to `/public` when not provided. I could change the order and still making `path` optional by assuming:

One argument on line, set as directory
#* @asset directory
Two arguments on line, set as path and directory
#* @asset path directory

Is it ok?

meztez: I will assume this is ok and go ahead with the changes.

meztez: Also, what a sweeping job you did. Good job.

schloerke: Unfortunately, we can not change the behavior of the tag `@assets` tag. This would definitely be breaking behavior. 

If anything we'd have to add a new tag (ex: `@static) and I'd like to pass on that for now.

When we add support for httpuv static paths (which don't support filters), we can make sure the new tag names are good.

vignettes/annotations.Rmd
```diff
@@ -0,0 +1,199 @@
...
+    pr_set_docs("swagger")
+}
+```
+
+##### Equivalent programmatic usage
+```{r, eval = FALSE, echo = TRUE}
+pr() %>%
+  pr_set_debug(TRUE) %>%
+  pr_set_docs("swagger")
+```
```
schloerke, line 199: This vignette is awesome. Thank you!

R/plumber.R
schloerke: I'm up for changing the name / logic as long as we can tell the difference between the initialization and customized `docs_info`.  (I know my current implementation was a lil bit of a hack.)

schloerke (changes requested):
If we can revert the `@assets` changes, I'm happy to merge! Thank you!

R/pr.R
meztez: Let's keep them in their original order then, even if mismatched.

schloerke (approved):

R/plumb-globals.R
```diff
@@ -32,37 +32,28 @@ plumbOneGlobal <- function(fields, argument){
            fields$info$termsOfService <- def
          },
          apiContact={
+           if (grepl("^list\\(", def)) {
+             def <- eval(parse(text = def))
```
cpsievert, line 36: This should probably have an `envir` coming in from the router

cpsievert (approved):
LGTM, pending suggested `envir` change

schloerke:
@meztez Let me know when it's ready and I'll merge. (Please feel free to add a news entry for you adding the annotations vignette. Thank you!!)

meztez:
> @meztez Let me know when it's ready and I'll merge. (Please feel free to add a news entry for you adding the annotations vignette. Thank you!!)

It is ready, let me finish investigate what github is complaining about.

PR #685, opened 2020-10-02
Allow for spaces in tags (surrounded by quotes)

### LINKED ISSUE #350: [Feature Request] Allow spaces in @apiTag
This was pulled from [this PR](https://github.com/trestletech/plumber/pull/230#discussion_r163072201). Well, here's one voice saying "yes please!"

meztez:
Fixes #350 

Allow to use spaces in tabs (or other character) as long as they are surrounded by quotes or double quotes.

Fix regex to allow one character description in description for response and tags.

Also reversed order of list elements to match plumber file written order, as appear on screen. This applies to plumber decoration that can be repeated in the same block.

```r
#* @apiTag "a b" Space with double quote
#* @apiTag 'c d' Space with single quote

#* @serializer csv
#* @get /test
#* @tag 'a b'
#* @tag "c d"
function() {
  mtcars
}
```

In reference to #350

PR task list:
- [x] Update NEWS
- [x] Add tests


R/plumb-globals.R
```diff
@@ -48,8 +48,8 @@ plumbOneGlobal <- function(fields, argument, envir = parent.frame()){
            fields$info$version <- def
          },
          apiTag={
-           tagMat <- stri_match(def, regex="^\\s*(\\w+)\\s+(\\S.+)\\s*$")
-           name <- tagMat[1,2]
+           tagMat <- stri_match(def, regex="^\\s*(\"[^\"]+\"|'[^']+'|\\w+)\\s+(\\S.+)\\s*$")
+           name <- stri_trim_both(tagMat[1,2], pattern = "[[\\P{Wspace}]-[\"']]")
```
schloerke, line 52: Can we make this a function so `plumb-block.R` can also use it?

```suggestion
           tagMat <- stri_match(def, regex="^\\s*(\"[^\"]+\"|'[^']+'|\\S+)\\s+(\\S.+)\\s*$")
           name <- stri_trim_both(tagMat[1,2], pattern = "[[\\P{Wspace}]-[\"']]")
```

schloerke (changes requested):

R/plumb-globals.R
meztez: I do not think so.
The block tag regex does not accept parameters while the global api one does.
Let's keep it like this for now. I will revisit this when I gather the courage to refactor plumb-globals and plumb-block.

PR #696, opened 2020-10-09
Fix #692, set openapi spec with file path

### LINKED ISSUE #692: Support OpenAPI Spec Files
Version 1.0.0 introduced a new way to modify the API UI via `pr_set_api_spec()`. This function accepts either a list object defining the spec or a function that modifies the existing spec and returns it. However, if the API spec is defined externally, there's no good way to provide it without serializing it into a list first.

For example, if I have a file `openapi.yml`:
```yaml
servers:
  - url: ''
openapi: 3.0.3
info:
  description: API Description
  title: Foo
  version: 1.0.0
paths:
  /health-check:
    get:
      summary: ' Determine if the API is running and listening as expected'
      responses:
        default:
          description: Default response.
      parameters: []
```

And I want to use that to define my API UI, currently, I need to do something like the following:
```r
pr %>%
  pr_set_api_spec(yaml::read_yaml("openapi.yaml"))
```

But this feels a bit odd since I'm serializing the spec file into a list, just so it can be turned back into a spec file. It would be nice if I could specify a file path to the spec file:

```r
pr %>%
  pr_set_api_spec("openapi.yaml")
```

meztez:
Fix #692, set openapi spec with file path

PR task list:
- [x] Update NEWS
- [x] Add tests


R/plumber.R
```diff
@@ -887,9 +887,20 @@ Plumber <- R6Class(
...
+    #'   * a path to an OpenAPI Specification
     #'
     #'  The value returned will not be validated for OAS compatibility.
     setApiSpec = function(api = NULL) {
+      if (is.character(api)) {
+        if (tools::file_ext(api) %in% c("yaml", "yml")) {
+          if (!requireNamespace("yaml", quietly = TRUE)) {
+            stop("yaml must be installed to read yaml format")
+          }
+          api <- yaml::read_yaml(api)
```
schloerke, line 899: Want to be explicit on this variable.
```suggestion
          api <- yaml::read_yaml(api, eval.expr = FALSE)
```

R/plumber.R
```diff
@@ -887,9 +887,20 @@ Plumber <- R6Class(
...
     #'  The value returned will not be validated for OAS compatibility.
     setApiSpec = function(api = NULL) {
+      if (is.character(api)) {
+        if (tools::file_ext(api) %in% c("yaml", "yml")) {
+          if (!requireNamespace("yaml", quietly = TRUE)) {
+            stop("yaml must be installed to read yaml format")
+          }
+          api <- yaml::read_yaml(api, eval.expr = FALSE)
+        } else {
+          api <- jsonlite::parse_json(api, simplifyVector = TRUE)
```
schloerke, line 901:
```suggestion
          api <- jsonlite::read_json(api, simplifyVector = TRUE)
```

R/plumber.R
```diff
@@ -887,9 +887,20 @@ Plumber <- R6Class(
     #' @param api This can be
     #'   * an OpenAPI Specification formatted list object
     #'   * a function that accepts the OpenAPI Specification autogenerated by `plumber` and returns a OpenAPI Specification formatted list object.
+    #'   * a path to an OpenAPI Specification
     #'
     #'  The value returned will not be validated for OAS compatibility.
     setApiSpec = function(api = NULL) {
+      if (is.character(api)) {
```
schloerke, line 894:
```suggestion
      if (is.character(api) && length(api) == 1 && file.exists(api)) {
```

schloerke (approved):

R/plumber.R
meztez: I'll have to check, I was not sure about checking for file.exist since it returns FALSE is you use a URL (non local file)

R/plumber.R
meztez: Same reason I went with parse_json, to allow for URL, read_json only work from disk files. If we do not need to support remote API file, I'm all for it.

schloerke:
Closing in favor of https://github.com/rstudio/plumber/pull/735

PR #710, opened 2020-10-26
Checking if router modified in plumb block is the one the argument

### LINKED ISSUE #707: @plumber tag should verify that the returned router matches the supplied router
The `@plumber` tag was introduced to enable increased flexibility while defining routers in the traditional Plumber style with tags. It's expected that it's used as follows:

```r
#* @plumber
function(pr) {
  # Modify and return pr
}
```

However, it's easy to accidentally do the following:
```r
#* @plumber
function(pr) {
  pr() %>%
    pr_get()
}
```

In this case, the supplied router isn't modified, but instead a new router is created. This leads to unexpected end results when the router is fully assembled. It would be helpful if Plumber verified that the output of a `@plumber` function was indeed a modification of the input instead of the instantiation of an entirely new router.

meztez:
Closes #707 

Draft idea, will finalize depending on comments.

PR task list:
- [ ] Update NEWS
- [ ] Add tests

### DIFF
--- a/R/plumb-block.R
+++ b/R/plumb-block.R
@@ -304,7 +304,12 @@ evaluateBlock <- function(srcref, file, expr, envir, addEndpoint, addFilter, pr)
         stopOnLine(lineNum, file[lineNum], e)
       })
       if (is.function(func)) {
-        func(pr)
+        assign("__plump_block_check__", TRUE, envir = envir)
+        on.exit(rm("__plump_block_check__", envir = envir), add = TRUE)
+        fpr <- func(pr)
+        if (inherits(fpr, "Plumber") && !isTRUE(fpr$environment[["__plump_block_check__"]])) {
+          stopOnLine(lineNum, file[lineNum], "Plumber object returned is not the same as the one in argument.")
+        }
         return()
       }
     }
--- a/R/plumb-block.R
+++ b/R/plumb-block.R
@@ -304,10 +304,10 @@ evaluateBlock <- function(srcref, file, expr, envir, addEndpoint, addFilter, pr)
         stopOnLine(lineNum, file[lineNum], e)
       })
       if (is.function(func)) {
-        assign("__plump_block_check__", TRUE, envir = envir)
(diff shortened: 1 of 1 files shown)


R/plumb-block.R
```diff
@@ -304,7 +304,12 @@ evaluateBlock <- function(srcref, file, expr, envir, addEndpoint, addFilter, pr)
         stopOnLine(lineNum, file[lineNum], e)
       })
       if (is.function(func)) {
-        func(pr)
+        assign("__plump_block_check__", TRUE, envir = envir)
+        on.exit(rm("__plump_block_check__", envir = envir), add = TRUE)
+        fpr <- func(pr)
+        if (inherits(fpr, "Plumber") && !isTRUE(fpr$environment[["__plump_block_check__"]])) {
```
schloerke, line 310: If we start with a `.`, then it won't show up with `ls()` by default.

```suggestion
        assign(".__plump_block_check__", TRUE, envir = envir)
        on.exit(rm(".__plump_block_check__", envir = envir), add = TRUE)
        fpr <- func(pr)
        if (inherits(fpr, "Plumber") && !isTRUE(fpr$environment[[".__plump_block_check__"]])) {
```

schloerke (commented):
Looks like it'll work! 

I think if we run into another 'flag' situation, I'd like to support it directly on the `Plumber` object itself. 

```r
pr$flag() # return all flags as named list
pr$flag(key) # return value
pr$flag(key, value) # set key with value
```
(No 'tidy' interface)

meztez:
Only thing that still itch me is if you create a second router that share the same environment
```R
function(pr) {
 pr(envir = pr$environment)
}
```

schloerke:
> Only thing that still itch me is if you create a second router that share the same environment

Flag route? Can also use it with printing as well.

`pr$flag("is_printing", TRUE)` / `pr$flag("is_printing")`

meztez:
So a flag facility per router, will see what can be turned into flags instead of attributes.

schloerke:
Closing in favor of #738

PR #719, opened 2020-12-01
Use `dev_set()` helper method

schloerke:
avoid opening a new device when setting to the NULL device.

Learned via thematic

PR task list:
- [NA] Update NEWS
- [x] Add tests

### DIFF
--- a/R/serializer.R
+++ b/R/serializer.R
@@ -652,8 +652,8 @@ createGraphicsDevicePromiseDomain <- function(which = dev.cur()) {
       force(onFulfilled)
       function(...) {
         old <- dev.cur()
-        dev.set(which)
-        on.exit(dev.set(old))
+        dev_set(which)
+        on.exit(dev_set(old))
 
         onFulfilled(...)
       }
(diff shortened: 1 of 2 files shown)


R/serializer.R
```diff
@@ -652,8 +652,8 @@ createGraphicsDevicePromiseDomain <- function(which = dev.cur()) {
       force(onFulfilled)
       function(...) {
         old <- dev.cur()
-        dev.set(which)
-        on.exit(dev.set(old))
+        dev_set(which)
+        on.exit(dev_set(old))
```
cpsievert, line 656: I *think* what you want is closer to:

```r
old <- dev.cur()
on.exit({
  if (old > 1) dev.set(old) else dev.off()
})
dev.set(which)
```

schloerke: Dev off is handled elsewhere. This is for handling the graphics device when changing promise domains

schloerke: When I do call dev off, I am telling which device to turn off

cpsievert: Ok, maybe you could just remove the `on.exit()` then? Let's talk through it later this week

R/serializer.R
```diff
@@ -652,8 +652,8 @@ createGraphicsDevicePromiseDomain <- function(which = dev.cur()) {
       force(onFulfilled)
```
cpsievert, line 652: Up above add `stopifnot(which >= 2)` (to make the rest of this easier to reason about)

R/serializer.R
```diff
@@ -662,18 +662,25 @@ createGraphicsDevicePromiseDomain <- function(which = dev.cur()) {
...
       force(expr)
     }
   )
 }
+    
+dev_set <- function(i) {
+  # make sure to not open a device when calling `dev.set(1)`
+  if (i > 1) {
+    dev.set(i)
+  }
```
cpsievert, line 685: Let's add a warning here (and maybe drop the stopifnot above?) just so we can be aware of this happening if it actually does (it shouldn't)

cpsievert (approved):

PR #720, opened 2020-12-09
Fix the description of $setHeader() in rendering-output.Rmd

s-fleck:
The description of the argument of `$setHeader()` was (slightly) wrong. Also `that's available in res$headers` suggests to me that it can only change existing headers, but that is not the case.

### DIFF
--- a/vignettes/rendering-output.Rmd
+++ b/vignettes/rendering-output.Rmd
@@ -31,7 +31,7 @@ Name | Example | Description
 ---- | ------- | -----------------------
 `setCookie` | `res$setCookie("foo", "bar")` | Sets an HTTP cookie on the client
 `removeCookie` | `res$removeCookie("foo")` | Removes an HTTP cookie
-`setHeader` | `res$setHeader(foo = "bar")` | Sets a header that's available in `res$headers`
+`setHeader` | `res$setHeader("foo", "bar")` | Sets an HTTP header
 `toResponse` | `res$toResponse()` | Renders the response object as a list containing `status`, `headers`, and `body`.
 
 The other methods (`clone`, `initialize`, and `serializer`) should not be directly invoked on the response object.


vignettes/rendering-output.Rmd
```diff
@@ -31,7 +31,7 @@ Name | Example | Description
 ---- | ------- | -----------------------
 `setCookie` | `res$setCookie("foo", "bar")` | Sets an HTTP cookie on the client
 `removeCookie` | `res$removeCookie("foo")` | Removes an HTTP cookie
-`setHeader` | `res$setHeader(foo = "bar")` | Sets a header that's available in `res$headers`
+`setHeader` | `res$setHeader("foo", "bar")` | Sets an HTTP header
```
meztez, line 34: ... or replace an existing one ...

s-fleck: right.. arguably that is also pretty unexpected behaviour. Are duplicated header names even legal in http?

Why not just `self$header[[name]] <- value`. I guess there must be a reason?

meztez: Seems like they may be legal https://tools.ietf.org/html/rfc7230#section-3.2.2 and I guess this is why it is not enforced. Maybe adding a parameter like `replace = FALSE`.

Replace current header when TRUE, append when FALSE

s-fleck:
```
A sender MUST NOT generate multiple header fields with the same field
   name in a message unless either the entire field value for that
   header field is defined as a comma-separated list [i.e., #(values)]
   or the header field is a well-known exception (as noted below).
```

So you are right, under certain circumstances it is legal. I like the `replace = TRUE` idea, but this pull request is just about fixing the documentation, and not about adding new features. Maybe change the description to `"Adds an HTTP header "foo" with the value "bar". Note that HTTP headers may contain several fields with the same name."

schloerke:
Thank you!

PR #722, opened 2020-12-11
Plumber block parsing order

### LINKED ISSUE #721: Resolve R devel `all.equal()` issue
> The  NEWS file  in "R-devel" (the development version of R)  now has
> in  'NEW Features'
> 
> > • all.equal(f, g) for functions now by default also compares their
> > environment(.)s, notably via new all.equal method for class
> > function.  Comparison of nls() fits, e.g., may now need
> > all.equal(m1, m2, check.environment=FALSE).
> 
> and all your packages need at least one such
> 'check.environment=FALSE'   in (at least one of)  your
> all.equal() calls  --- or  expect_equal(..)  testthat calls.

> See this R-devel (mailing list) thread on why we made the
> change:
> 
>   https://stat.ethz.ch/pipermail/r-devel/2020-December/080172.html
> 
> and please ask for help / questions on R-devel (the mailing
> list), or possibly me, if necessary.

> Note that I've investigated 'tvthemes' package in particular, as
> its use of testthat ((as usual in such cases, because of all the
> 'testthat' tricks leading to obfuscation)) gave an error message
> which was not at all directly pointing to the issue:
> 
> There, I've even opened a github issue with the solution (for
> 'tvthemes') here:
> 
>     https://github.com/Ryo-N7/tvthemes/issues/15
> 
> The "gist" of it is to define a simple wrapper function

> ## For R-devel (aka R 4.1.x), but also works in older R
> expect_eqNe <- function(...) expect_equal(..., check.environment=FALSE)
> 
> 
> and then -- in a dozen cases in that case -- use
> 
> expect_eqNe() instead of expect_equal()
> 
> Alternatively, of course, you'd add
>   " , check.environment=FALSE "  (without quotes)
> to your all.equal() / expect_equal()  calls.
> 
> In cases as above, it's faster to provide the wrapper and use
> "global replace everything".
> 
> Thank you for helping to make R more powerful and useful by
> maintaining your packages on CRAN!

> With best regards,
> Martin

meztez:
Fixes #721
Fixes https://github.com/rstudio/plumber/issues/187#issuecomment-743313501

Reverse responses, tags and comments before returning to respect plumber file order.

PR task list:
- [x] Update NEWS
- [x] Add tests


schloerke (approved):
🙌

NEWS.md
```diff
@@ -7,6 +7,7 @@ plumber 1.0.0.9999 Development version
 
 ### Bug fixes
 
+* Block parsing comments, tags and responses ordering match plumber api ordering.
```
schloerke, line 10:
```suggestion
* Block parsing comments, tags and responses ordering match plumber api ordering. (#722)
```

schloerke (approved):

PR #724, opened 2020-12-15
ws experiment

meztez:
It requires quite a bit of comprehension to build something useful out of if.

You get an `Error in wsSend(private$wsObj, msg) : invalid state` when you try to use websockets while a request is being processed by the regular REST API.

I've yet to find a way to leverage the utility of enabling websockets over regular plumbing, this need more thinking.

PR task list:
- [ ] Update NEWS
- [ ] Add tests

### DIFF
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -755,7 +755,10 @@ Plumber <- R6Class(
     #' @description httpuv interface onWSOpen function. (Required for \pkg{httpuv})
     #' @param ws WebSocket object
     onWSOpen = function(ws){
-      warning("WebSockets not supported.")
+      if (!is.null(private$ws_open)) {
+        private$ws_open(ws)
+      }
+      invisible(self)
     },
     #' @description Sets the default serializer of the router.
     #'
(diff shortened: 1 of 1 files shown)


meztez:
I think this is more interesting.

api
```r
#* @post /sleeper
#* @get /sleeper
#* @param payload:df
#* @parser json
#* @serializer unboxedJSON
function(payload = matrix(1:8, 2, 4), req) {
  a <- rowSums(payload)
  req$ws$send("Almost done")
  b <- colSums(payload)
  req$ws$send("Calculation completed")
  list(a, b)
}
```

Using websockets
```r
ws <- websocket::WebSocket$new("ws://127.0.0.1:3946/sleeper")
ws$onMessage(function(res) { cat(res$data, "\n") })
ws$send('{"payload":[[1,3,5,7],[2,4,6,8]]}')
ws$close()
```

results
```r
> ws <- websocket::WebSocket$new("ws://127.0.0.1:3946/sleeper")
> ws$onMessage(function(res) { cat(res$data, "\n") })
> ws$send('{"payload":[[1,3,5,7],[2,4,6,8]]}')
Almost done 
Calculation completed 
[[16,20],[3,7,11,15]] 
> ws$close()
```
Then you can get intermediate response before getting the final response. Still use case to cover, I think this is a good start.

R/websocket.R
```diff
@@ -0,0 +1,27 @@
...
+          if (binary) rawToChar(message) else message
+        },
+        assign.env = req
+      )
+      req$.internal$bodyHandled <- TRUE
+      res <- PlumberResponse$new(ser)
+      pr$serve(req, res)
+      ws$send(paste("_status_", res$status))
+      ws$send(paste("_headers_", paste(names(res$headers), unlist(res$headers), sep = "=", collapse = ";")))
+      ws$send("_body_ nextmessage")
```
meztez, line 23: It probably makes more sense to only return `res$body` in websockets context?

meztez:
Would websockets bypass cookie or apikey filter?

Most likely they should not be enabled by default.

PR #727, opened 2020-12-15
Update README.md curl example

### LINKED ISSUE #725: Add content type header for json arguments in documentation example
### tl;dr: 

Change:

```
curl --data '{"a":4, "b":5}' http://localhost:8000/sum
```

To:

```
curl -H "Content-Type: application/json" --data '{"a":4, "b":5}' http://localhost:8000/sum
```

In the [plumber docs](https://www.rplumber.io/index.html).

### System details

```
R version 4.0.2 (2020-06-22)
Platform: x86_64-apple-darwin17.0 (64-bit)
Running under: macOS Catalina 10.15.6

locale:
[1] en_US.UTF-8/en_US.UTF-8/en_US.UTF-8/C/en_US.UTF-8/en_US.UTF-8

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] plumber_1.0.0.9999

loaded via a namespace (and not attached):
 [1] compiler_4.0.2      magrittr_2.0.1      R6_2.5.0            later_1.1.0.1      
 [5] promises_1.1.1.9001 tools_4.0.2         swagger_3.33.1      yaml_2.2.1         
 [9] Rcpp_1.0.5          stringi_1.5.3       jsonlite_1.7.2      webutils_1.1       
[13] httpuv_1.5.4        lifecycle_0.2.0     rlang_0.4.9.9000   

```

According to the [plumber docs](https://www.rplumber.io/index.html), I should be able to pass function arguments in JSON format. Given the following server:

```R
#* Return the sum of two numbers
#* @param a The first number to add
#* @param b The second number to add
#* @post /sum
function(a, b) {
  as.numeric(a) + as.numeric(b)
}
```

If I then start the server:

```R
library(plumber)
library(dplyr)
pr("plumber.R") %>%
     pr_run(port=8000)
```

This works as expected:

```
curl --data "a=4&b=3" "http://localhost:8000/sum"
[7]
```

But this does not work, although the docs linked above seem to state it should:

```
curl --data '{"a":4, "b":5}' http://localhost:8000/sum
```

The error I get is:

```
<simpleError in (function (a, b) {    as.numeric(a) + as.numeric(b)})(): argument "a" is missing, with no default>
```

### Solution

The solution is simply to declare the content type: 

```
curl  -H "Content-Type: application/json" --data '{"a":4, "b":5}' http://localhost:8000/sum
```
 Can this be clarified in the documentation reference above?

meztez:
Fixes #725 

specify header for curl example

second issue for the same thing, miss it last time around.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ $ curl --data "a=4&b=3" "http://localhost:8000/sum"
 You can also send your data as JSON:
 
 ```
-$ curl --data '{"a":4, "b":5}' http://localhost:8000/sum
+$ echo {"a":4, "b":5} > call.json & curl --data @call.json "http://localhost:8000/sum" -H "content-type: application/json"
  [9]
 ```
 
--- a/README.md
+++ b/README.md
@@ -76,7 +76,7 @@ $ curl --data "a=4&b=3" "http://localhost:8000/sum"
 You can also send your data as JSON:
 
 ```
-$ echo {"a":4, "b":5} > call.json & curl --data @call.json "http://localhost:8000/sum" -H "content-type: application/json"
+$ curl -H "Content-Type: application/json" --data '{"a":4, "b":5}' http://localhost:8000/sum
  [9]
 ```
 


README.md
```diff
@@ -76,7 +76,7 @@ $ curl --data "a=4&b=3" "http://localhost:8000/sum"
 You can also send your data as JSON:
 
 ```
-$ curl --data '{"a":4, "b":5}' http://localhost:8000/sum
+$ echo {"a":4, "b":5} > call.json & curl --data @call.json "http://localhost:8000/sum" -H "content-type: application/json"
```
schloerke, line 79: To keep the examples similar, let's avoid writing to a file

```suggestion
$ curl -H "Content-Type: application/json" --data '{"a":4, "b":5}' http://localhost:8000/sum
```

schloerke (approved):

meztez:
There is another instance in the doc using a file. We might want to change it too

schloerke:
I could not find the other instance. LGTM

PR #730, opened 2020-12-17
Update Dockerfile

meztez:
substitute deprecated argument swagger for setDocs method

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -23,7 +23,7 @@ ARG PLUMBER_REF=master
 RUN Rscript -e "remotes::install_github('rstudio/plumber@${PLUMBER_REF}')"
 
 EXPOSE 8000
-ENTRYPOINT ["R", "-e", "pr <- plumber::plumb(rev(commandArgs())[1]); pr$setDocs(TRUE); pr$run(host='0.0.0.0', port=8000)"]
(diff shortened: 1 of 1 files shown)


schloerke:
So with this update, it'll break when using plumber tag v0.4.6 as `$setDocs()` is not supported.

I'm wondering if we should abandon republishing tag plumber v0.4.6 with the latest R updates. It would make maintenance MUCH easier.  We would have to remove support in the docker workflow.

meztez:
Good point, I see 2.8k pulls, I don't know if you have access to a split statistic. If we avoid republishing, the image won't benefit from OS updates. Maybe there should be a warning.

Dockerfile
```diff
@@ -23,7 +23,7 @@ ARG PLUMBER_REF=master
 RUN Rscript -e "remotes::install_github('rstudio/plumber@${PLUMBER_REF}')"
 
 EXPOSE 8000
-ENTRYPOINT ["R", "-e", "pr <- plumber::plumb(rev(commandArgs())[1]); pr$run(host='0.0.0.0', port=8000, swagger=TRUE)"]
+ENTRYPOINT ["R", "-e", "pr <- plumber::plumb(rev(commandArgs())[1]); pr$setDocs(TRUE); pr$run(host='0.0.0.0', port=8000)"]
```
schloerke, line 26: Maybe we could check the package version, then we can accommodate both for now. Don't want to disable legacy support if we can get around it with a few characters. Once it starts to look like a hack, we can disable it.

```suggestion
ENTRYPOINT ["R", "-e", "pr <- plumber::plumb(rev(commandArgs())[1]); args <- list(host = '0.0.0.0', port = 8000); if (packageVersion('plumber') >= 1.0.0) { pr$setDocs(TRUE) } else { args$swagger <- TRUE }; do.call(pr$run, args)"]
```

(untested)

schloerke (approved):

Issue #734, opened 2020-12-28
Mounting at `/` exhibits odd behavior


blairj09:
Reprex:
```r
  pr() %>% 
    pr_mount("/", 
             pr() %>% 
               pr_get("/hello", function() "hi")
    ) %>% 
    pr_run()
```

When this is executed, docs doesn't render and returns a 404.

meztez:
Probably because the docs are mounted on the first router and the route being served is on the nested router.

meztez:
I have a clumsy fix 

in plumber.R
```r
...
# We aren't going to serve this endpoint; see if any mounted routers will
      mountSteps <- lapply(sort(names(private$mnts), decreasing = TRUE), function(mountPath) {
       
```

It creates problem because we use this condition to see if a mounted router can be used to serve request.
```r
if (nchar(path) >= nchar(mountPath) && substr(path, 0, nchar(mountPath)) == mountPath) {
```

I will try to think of a better solution before submitting a patch. Decrease sorting works because it will evaluates the longest path first.

meztez:
This could also happens if we had two routers, one at "/aaa/" and one at "/aaa/bbb/".

```r
pr <- pr() %>% 
  pr_mount("/aaa", pr()) %>%
  pr_mount("/aaa/bbb",
           pr() %>% 
             pr_get("/hello", function() "hi")
  ) %>% pr_run()
```

I'm thinking either we sort mounts on mounting with something like
```r
mount = function(path, router) {
      # Ensure that the path has both a leading and trailing slash.
      if (!grepl("^/", path)) {
        path <- paste0("/", path)
      }
      if (!grepl("/$", path)) {
        path <- paste0(path, "/")
      }

      private$mnts[[path]] <- router
      private$mnts <- private$mnts[order(names(private$mnts), decreasing = TRUE)]
    }
```

or what I suggested above.

It still is not clear to reviewer reading the code why we need to sort so I think a better alternative should be found.

I still do not like this solution.

schloerke:
👀

schloerke:
@meztez Great analysis and fix! 

In the case of `/aaa/bbb/foo` and mounts ordered c(`/aaa/bbb`, `/aaa`), I think it is fair to say that a request of `/aaa/bbb/foo` should attempt through `/aaa/bbb` (vs `/aaa`). ... the mount with the largest match.

[...]

PR #748, opened 2020-12-31
When processing a route, use most specific mount paths first

schloerke:
Fixes https://github.com/rstudio/plumber/issues/734

* More specific mount paths will be found by using this order

Reprex:
```r
pr <- pr() %>% 
  pr_mount("/aaa", pr()) %>%
  pr_mount("/aaa/bbb",
           pr() %>% 
             pr_get("/hello", function() "hi")
  ) %>% pr_run()
```
... Which mount processes `/aaa/bbb/hello`?
* Before pr: `/aaa`
* With pr: `/aaa/bbb`

PR task list:
- [ ] Update NEWS
- [x] Add tests

Questions:
* Is it ok to return different order for `pr$mounts`? (Now it is alpha sorted. Before, it was added order)

### DIFF
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -1041,9 +1041,10 @@ Plumber <- R6Class(
     filters = function(){
       private$filts
     },
-    #' @field mounts Plumber router mounts read-only
+    #' @field mounts Plumber router mounts sorted by mount location. Read-only.
     mounts = function(){
-      private$mnts
+      mnts <- private$mnts
+      mnts[sort(names(mnts), decreasing = FALSE)]
     },
     #' @field environment Plumber router environment read-only
     environment = function() {
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -242,7 +242,9 @@ Plumber <- R6Class(
         path <- paste0(path, "/")
       }
 
+      # order the mounts so that the more specific mount paths are ahead of the less specific mount paths
       private$mnts[[path]] <- router
+      private$mnts <- private$mnts[order(names(private$mnts), decreasing = TRUE)]
     },
     #' @description Unmount a Plumber router
     #' @param path a character string. Where to unmount router.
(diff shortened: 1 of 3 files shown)


R/plumber.R
```diff
@@ -1021,23 +1039,25 @@ Plumber <- R6Class(
...
     filters = function(){
       private$filts
     },
-    #' @field mounts Plumber router mounts read-only
+    #' @field mounts Plumber router mounts sorted by mount location. Read-only.
     mounts = function(){
-      private$mnts
+      mnts <- private$mnts
+      if (length(mnts) == 0) return(mnts)
+      mnts[sort(names(mnts), decreasing = FALSE)]
```
schloerke, line 1054:
```suggestion
      mnts[order(names(mnts), decreasing = FALSE)]
```

R/plumber.R
```diff
@@ -704,8 +712,18 @@ Plumber <- R6Class(
...
-      mountSteps <- lapply(names(private$mnts), function(mountPath) {
+      # This router can not serve this endpoint; See if any mounted routers will.
+      # `private$mnts` is reverse alpha-sorted to have more specific mount paths be found first.
+      # `self$mounts` is alpha-sorted to have less specific mount paths be found first.
+      # Ex:
+      # * `/aaa/bbb/foo` is requested.
+      # * Mounts `/aaa` and `/aaa/bbb` exist.
+      # * We want to use mount `/aaa/bbb` as it is more specific
+      # TODO
+      # * If `/aaa/bbb` mount does not support `/aaa/bbb/foo`, then try mount `/aaa`.
```
schloerke, line 723:
```suggestion
      # * If `/aaa/bbb` mount does not support `/aaa/bbb/foo`, then try mount `/aaa` with route `/bbb/foo`.
```

R/plumber.R
```diff
@@ -242,7 +248,9 @@ Plumber <- R6Class(
         path <- paste0(path, "/")
       }
 
+      # order the mounts so that the more specific mount paths are ahead of the less specific mount paths
```
schloerke, line 251:
```suggestion
      # Add the mount to the original mount order
      private$mnt_order <- c(private$mnt_order, path)
      
      # Order the mounts so that the more specific mount paths are ahead of the less specific mount paths
```

R/plumber.R
schloerke, line 1054:
```suggestion
      private$mnts[private$mnt_order]
```

schloerke:
This change is deemed too dangerous. Maybe approach this with `pr_merge()` (#https://github.com/rstudio/plumber/issues/749) instead?

cc @cpsievert

PR #751, opened 2021-01-04
Quote the version number in Dockerfile

### LINKED ISSUE #750: v1.0.0 Plumber Docker Entrypoint

1. Run rstudio/plumber:v1.0.0.

### Problem 1 -- Pushed Master onto Docker v1.

2. **Expectation**:  to see an output line like [this, off of the v1.0.0 github tag](https://github.com/rstudio/plumber/blob/c051c6b6940c3887992caf9be925d551956271d9/Dockerfile#L26).
3. **Actual**:  [See this as an output instead](https://github.com/rstudio/plumber/blob/6499ed98f989d6040c9247c4df836b6f5bbd1f04/Dockerfile#L26). Clearly the same entrypoint off of master. 

This makes me think that we are actually 1.0.0+ for the 1.0.0 image. Even if they were minor changes, shouldn't there have been a new github release instead?

### Problem 2 - Entrypoint errors

1. `docker run --rm -it  -p 8000:8000/tcp rstudio/plumber:v1.0.0`
2. **Expectation**: app should be running.
3. **Actual**:  `Error: unexpected numeric constant in " if (packageVersion('plumber') >= 1.0.0" Execution halted`

schloerke:
Fixes #750

### DIFF
--- a/.github/workflows/docker.yaml
+++ b/.github/workflows/docker.yaml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - "master"
-      - "docker/**"
+      - "docker**"
   schedule:
     - cron:  '0 0 1 * *' # first of every month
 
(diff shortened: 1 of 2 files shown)


schloerke:
@kassemsandarusi Can you please try again. I have verified that `rstudio/plumber:latest` works.

schloerke:
Going to merge as I can not run the master version, but I can run this PR's version.

PR #754, opened 2021-01-07
decode static URI before serving

### LINKED ISSUE #753: Maybe support Chinese character in req$PATH_INFO
From https://community.rstudio.com/t/pr-static-with-filename-contains-chinese-character/92499/2, a user asked if plumber supported Chinese characters directly in the PATH_INFO to serve static documents. Currently no.

This would work but it might 
 - break some stuff (all tests still pass but you never know).
 - add to execution overhead.

There are others places where this conversion could be made.

`plumber.R`
```r
    #' @description \pkg{httpuv} interface call function. (Required for \pkg{httpuv})
    #' @param req request object
    call = function(req) {
      # Set the arguments to an empty list
      req$pr <- self
      req$.internal <- new.env()
      req$PATH_INFO <- httpuv::decodeURIComponent(req$PATH_INFO)

      res <- PlumberResponse$new(private$default_serializer)
      req$args <- list()

      # maybe return a promise object
      self$serve(req, res)
    },
```

less intrusive, regular plumber routes would still not support Chinese characters
`plumber-static.R`
```r
        path <- httpuv::decodeURIComponent(path)
        abs.path <- resolve_path(direc, path)
        if (is.null(abs.path)){
          # TODO: Should this be inherited from a parent router?
          val <- private$notFoundHandler(req=req, res=res)
          return(val)
        }
```

Something to think about

meztez:
Closes #753

PR task list:
- [x] Update NEWS
- [x] Add tests


meztez:
I'm thinking of creating the file on the flies for the test. Since I get an R check error for the filename.

schloerke:
Will need to check if file exists as some systems won't be able to write chinese characters. 😞 

Or maybe skip on cran and only test on macos?  Then the logic will at least be tested inside GHA.

meztez:
I'll wait to see which one fails

meztez:
so skip on CRAN

PR #759, opened 2021-02-05
Add ability to have custom 405 handler

### LINKED ISSUE #757: HTTP 405 response bodies/headers cannot be customised

Every Plumber API sent an incorrect method to a valid route (e.g. `PATCH /route` where only `GET /route` is supported) is affected by this issue.

Plumber has `pr_set_404()` and `pr_set_error()` functions (among other, older approaches) that allow users to control the response body/headers of errors returned from an API. We use this extensively to emit common error structures. However, since #746 the HTTP 405 response has been [hardcoded internally](https://github.com/rstudio/plumber/blob/e9dac935dab37d318f9627caafe17c8ea128023d/R/plumber.R#L759-L768) and it is no longer possible to control it.

This is not a bug per se (since it is intended behaviour) but it *is* a regression, since it was previously possible to control HTTP 405 responses from the 404 handler.

I originally proposed in #493 that Plumber have a dedicated handler for HTTP 405 responses (presumably `pr_set_405()`), and still believe that that is the right level of control here -- I would be happy to implement this myself, too.

meztez:
Fixes #757

I could not explain it better than #757

This is crude, I would still prefer to have a more global approach to custom error handling

PR task list:
- [x] Update NEWS
- [x] Add tests


NEWS.md
```diff
@@ -23,6 +23,8 @@ plumber 1.0.0.9999 Development version
 
 * OpenAPI Specification can be set using a file path. (@meztez #696)
 
+* Add `$set_505_handler()` and `pr_set_405()` to customize 405 error type.
```
schloerke, line 26:
```suggestion
* Add `$set_405_handler()` and `pr_set_405()` to customize 405 error type. (#759)
```

schloerke (approved):

R/plumber.R
```diff
@@ -853,10 +850,23 @@ Plumber <- R6Class(
...
+    #'
+    #' See also: [pr_set_405()]
+    #' @param fun a handler function.
+    #' @examples
+    #' \dontrun{
+    #' pr <- pr()
+    #' pr$set405Handler(function(req, res) {cat(req$PATH_INFO)})
+    #' }
+    set405Handler = function(fun){
+      private$unallowedMethodHandler <- fun
```
cpsievert, line 864: `set404Handler` should maybe do the same (to avoid errors that aren't caught until runtime)?

```suggestion
      stopifnot(is.function(fun))
      args <- names(formals(fun))
      if (!all(c("req", res") %in% args)) {
        stop("fun must have formal arguments for req and res")
      }
      private$unallowedMethodHandler <- fun
```

meztez: There a couple other plumber functions that take a function as input and are executed at runtime with potential failure. This might be material for another PR

cpsievert (approved):

R/plumber.R
schloerke: @meztez Would you like / have time to do a PR for the other method parameter checks?

meztez: Eventually, I still have the regular job thing to do 😄

meztez:
Scrath this PR, I'm thinking long term we should have a better way to handler errors.

Something like serializers and parsers, a collection of error handler. This PR is more a bandaid than a real solution.

PR #760, opened 2021-02-05
Fix Content-Length header on HTTP responses which forbid it

### LINKED ISSUE #758: Plumber sets a Content-Length header on HTTP responses which forbid it (e.g. 1xx, 204, 304)

Given the following Plumber API which returns [HTTP 204 No Content](https://tools.ietf.org/html/rfc7231#section-6.3.5):

```R
library(plumber)

pr() %>%
  pr_get("/", function(req, res) {
    res$status <- 204
    res
  }) %>%
  pr_run(port = 4444)
```

Querying this with `curl` yields:

```
$ curl -vs localhost:4444/
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connected to localhost (127.0.0.1) port 4444 (#0)
> GET / HTTP/1.1
> Host: localhost:4444
> User-Agent: curl/7.58.0
> Accept: */*
> 
< HTTP/1.1 204 No Content
< Date: Fri, 05 Feb 2021 05:07:53 GMT
< Content-Length: 0
< 
* Connection #0 to host localhost left intact
```

You can see in the response above that Plumber sets the `Content-Length` header. RFC 7230 (part of the HTTP 1.1 specification) [says the following](https://tools.ietf.org/html/rfc7230#section-3.3.2) about the `Content-Length` header:

> A server MAY send a Content-Length header field in a 304 (Not
> Modified) response to a conditional GET request (Section 4.1 of
> [RFC7232]); a server MUST NOT send Content-Length in such a response
> unless its field-value equals the decimal number of octets that would
> have been sent in the payload body of a 200 (OK) response to the same
> request.
>
> A server MUST NOT send a Content-Length header field in any response
> with a status code of 1xx (Informational) or 204 (No Content).

Ergo we're currently in violation of the HTTP spec.

The reason this happens is actually because the underlying `httpuv` package previously set the `Content-Length` header in all cases. I [fixed this recently](https://github.com/rstudio/httpuv/pull/288) and the fix is now on CRAN (as v1.5.5).

Plumber should check if the status code is `1xx`, `204`, or `304` and set the response body to `NULL` in those cases, as well as require `httpuv >= 1.5.5` so that this actually works.

meztez:
Fixes #758 

Once again, I could not have said it better than #758 

```r
library(plumber)

pr() %>%
  pr_get("/", function(req, res) {
    res$status <- 204
    res
  }) %>%
  pr_run(port = 8004)
```

PR task list:
- [x] Update NEWS
- [x] Add tests


tests/testthat/test-response.R
```diff
@@ -44,3 +44,10 @@ test_that("response properly sets cookies with multiple options", {
   head <- res$toResponse()$headers
   expect_equal(head[["Set-Cookie"]], "abc=two%20words; HttpOnly; Secure; SameSite=None")
 })
+
+test_that("Content-Length header on HTTP responses which forbid it", {
```
atheriel, line 48: Should probably be "Body on HTTP responses which forbid it" instead.

NEWS.md
```diff
@@ -45,6 +45,7 @@ plumber 1.0.0.9999 Development version
 
 * Changed `future::plan()` from `multiprocess` to `multisession` in example API `14-future` as "Strategy 'multiprocess' is deprecated in future (>= 1.20.0)". (#747)
 
+* Fix Content-Length header on HTTP responses which forbid it (e.g. 1xx, 204, 304)
```
atheriel: Should probably be something more like

> Remove response bodies (and therefore the `Content-Length` header) for status codes which forbid it under the HTTP specification (e.g. 1xx, 204, 304). (#758, #760)

meztez: Maybe you did, maybe you didn't. I wanted to point out that you could suggest code change directly using a suggestion block in your comment (```suggestion ... ```)
https://docs.github.com/en/github/collaborating-with-issues-and-pull-requests/commenting-on-a-pull-request

Then lazy me could click on a button and isn't that one the greatest achievement in human history? Click a button, works get done.

schloerke (approved):

PR #766, opened 2021-02-07
Setting `plumber.docs.callback` to `NULL` will also set legacy `plumber.swagger.url`

meztez:
…cated but supported option `plumber.swagger.url`.

Currently this does nothing since the router default to `plumber.swagger.url` when `plumber.docs.callback` is NULL

```r
library(plumber)
options_plumber(docs.callback = NULL)
pr() %>% pr_run()
```

This PR will make sure  `plumber.swagger.url` is set to `NULL` when calling `options_plumber(docs.callback = NULL)`.

We could also stop supporting `plumber.swagger.url` at some point.

PR task list:
- [x] Update NEWS
- [x] Add tests

### DIFF
--- a/R/options_plumber.R
+++ b/R/options_plumber.R
@@ -67,6 +67,7 @@ options_plumber <- function(
 ) {
   ellipsis::check_dots_empty()
 
+  # Make sure all fallback options are disabled
   if (!missing(docs.callback) && is.null(docs.callback)) {
     options("plumber.swagger.url" = NULL)
   }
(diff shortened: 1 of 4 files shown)


R/options_plumber.R
```diff
@@ -67,6 +67,10 @@ options_plumber <- function(
 ) {
   ellipsis::check_dots_empty()
 
+  if (!missing(docs.callback) && is.null(docs.callback)) {
+    options("plumber.swagger.url" = NULL)
+  }
```
schloerke, line 72:
```suggestion
  # Make sure all fallback options are disabled
  if (!missing(docs.callback) && is.null(docs.callback)) {
    options("plumber.swagger.url" = NULL)
  }
```

schloerke (approved):
LGTM

PR #809, opened 2021-06-02
Shared secret rejections are not errors

### LINKED ISSUE #808: shared secret mismatches throw errors
Given any Plumber API, for versions of plumber up to 1.1.0:

```r
options(`plumber.sharedSecret`="abcdefg")
plumber::plumb(file='plumber.R')$run()
# On connecting to this application:
# <simpleError in (function (req, res) {    secret <- getOption("plumber.sharedSecret", NULL)    if (!is.null(secret)) {        supplied <- req$HTTP_PLUMBER_SHARED_SECRET        if (!identical(supplied, secret)) {            res$status <- 400            stop("The provided shared secret did not match expected secret.")        }    }    forward()})(req = <environment>, res = <environment>): The provided shared secret did not match expected secret.>
```

Interacting with that API gives:

```console
curl http://127.0.0.1:3063/echo
# {"error":"Internal error","message":"Error in (function (req, res) : The provided shared secret did not match expected secret.\n"}
```

It is not necessary to use `stop`; returning an error response is sufficient.

aronatkins:
This change has the shared secret filter no longer log an error and returns a cleaner error response.

```r
options(`plumber.sharedSecret`="abcdefg")
p <- plumber::Plumber$new(file = NULL)
p$run()
# Running plumber API at http://127.0.0.1:8050
# Running swagger Docs at http://127.0.0.1:8050/__docs__/
```

```console
curl http://127.0.0.1:8050/
# {"error": "Shared secret mismatch."}
```

Fixes #808 

CC @blairj09

### DIFF
--- a/R/shared-secret-filter.R
+++ b/R/shared-secret-filter.R
@@ -7,7 +7,14 @@ sharedSecretFilter <- function(req, res){
       res$status <- 400
       # Force the route to return as unboxed json
       res$serializer <- serializer_unboxed_json()
-      return(list(error = "Shared secret mismatch."))
+      # Using output similar to `defaultErrorHandler()`
+      li <- list(error = "400 - Bad request")
+      
+      # Don't overly leak data unless they opt-in
+      if (is.function(req$pr$getDebug) && isTRUE(req$pr$getDebug())) {
+        li$message <- "Shared secret mismatch"
+      }
+      return(li)
     }
   }
 
--- a/R/shared-secret-filter.R
+++ b/R/shared-secret-filter.R
@@ -5,7 +5,7 @@ sharedSecretFilter <- function(req, res){
     supplied <- req$HTTP_PLUMBER_SHARED_SECRET
     if (!identical(supplied, secret)){
       res$status <- 400
-      stop("The provided shared secret did not match expected secret.")
+      return(list(error = "Shared secret mismatch."))
     }
   }
 
--- a/R/shared-secret-filter.R
+++ b/R/shared-secret-filter.R
@@ -5,6 +5,8 @@ sharedSecretFilter <- function(req, res){
     supplied <- req$HTTP_PLUMBER_SHARED_SECRET
     if (!identical(supplied, secret)){
       res$status <- 400
+      # Force the route to return as unboxed json
+      res$serializer <- serializer_unboxed_json()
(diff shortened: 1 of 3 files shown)


blairj09:
Thanks @aronatkins!

meztez:
What should be the http status of these?

schloerke:
> What should be the http status of these?

From `options_plumber`: https://github.com/rstudio/plumber/blob/e846cb3edf980cfea3e37325c54d69c814db2194/R/options_plumber.R#L38-L40

... so `400`.

R/shared-secret-filter.R
```diff
@@ -5,7 +5,7 @@ sharedSecretFilter <- function(req, res){
     supplied <- req$HTTP_PLUMBER_SHARED_SECRET
     if (!identical(supplied, secret)){
       res$status <- 400
```
schloerke, line 7:
```suggestion
      res$status <- 400
      # Force the route to return as unboxed json
      res$serializer <- serializer_unboxed_json()
```

R/shared-secret-filter.R
```diff
@@ -5,7 +5,9 @@ sharedSecretFilter <- function(req, res){
...
-      stop("The provided shared secret did not match expected secret.")
+      # Force the route to return as unboxed json
+      res$serializer <- serializer_unboxed_json()
+      return(list(error = "Shared secret mismatch."))
```
schloerke, line 10: Move behavior similar to the current default error handler. Only provide the message if the `debug` is set.

```suggestion
      # Using output similar to `defaultErrorHandler()`
      li <- list(error = "400 - Bad Request")
      
      # Don't overly leak data unless they opt-in
      if (is.function(req$pr$getDebug) && isTRUE(req$pr$getDebug())) {
        li["message"] <- "Shared secret mismatch"
      }
      return(li)
```

schloerke: Reference: https://github.com/rstudio/plumber/blob/e846cb3edf980cfea3e37325c54d69c814db2194/R/default-handlers.R#L7-L34

schloerke (approved):
LGTM pending suggestion

(Windows devel is unrelated to failure)
(Please squash+merge when ready)

cpsievert (approved):

schloerke:
Thank you for the tests @aronatkins!

PR #830, opened 2021-09-22
Create GeoJSON 

JosiahParry:
## Pull Request

This pull requests implements a geojson serializer. The serializer utilizes [geojsonsf](https://github.com/SymbolixAU/geojsonsf). The serializer only supports objects of class `sf` and `sfc`.

Please note that `devtools::check()` fails due to `hookable.R#3` and it cannot find the namespace. As such I cannot provide check results without fiddling with code that is not mine. 

No code was modified, only added. As such there should be no breaking features or breaking backwards compatibility. 
---

At present, the only way to provide geojson—that I am aware of—via plumber API is to utilize sf to write to a temporary file. read that tempfile via `readLines()` and use a `cat` serializer to return the exact json that was written by sf. This serializer removes the need to create hacky solutions for those who would like to return geojson from a plumber API. 

```r
library(sf)
library(plumber)

nc <- st_read(system.file("shape/nc.shp", package="sf"))

#* @apiTitle GeoJSON serializer

#* Test sfc object
#* @param id The row ID to print. Must be between 1 - 100.
#* @get /sfc
#* @serializer geojson
function(id = "") {
  nc[["geometry"]][as.integer(id)]
}

#* Test sf object
#* @param id The row ID to print. Must be between 1 - 100.
#* @get /sf
#* @serializer geojson
function(id = "") {
  nc[as.integer(id),]
}

```

PR task list:
- [ ] Update NEWS
- [x] Add tests


meztez:
As an alternative. You can register a serializer or a parser directly on the fly. Here is a an example using an xml parser.

```r
library(plumber)
library(xml2)
library(CreditScore)

# make sure the pods does not get up if we are not able to
# calculate credit score
CreditScore:::warm_up()

serializer_xml <- function(..., type = "application/xml") {
  serializer_content_type(type, function(val) {
    as.character(xml2::as_xml_document(val, ...))
  })
}
register_serializer("xml", serializer_xml)

parser_xml <- function(...) {
  parser_text(function(txt_value) {
    xml2::as_list(xml2::read_xml(txt_value, ...))
  })
}
register_parser("xml", parser_xml, fixed = c("text/xml", "application/xml"), regex = "/xml$")

#* @apiTitle Credit Score
#* @apiDescription Compute credit score from a credit report
#* @apiVersion 1.0.6

#* Credit Score / xml formatted response
#* @post /score/xml
#* @serializer xml
#* @parser xml
#* @parser multi
#* @param policyInfo:file
#* @param report:file
#* @param debug
function(policyInfo, report, req) {
  req$extradata$CorrelationID <- req$HTTP_X_CORRELATION_ID
  CreditScore:::process_input(policyInfo, report)
}
```

DESCRIPTION
```diff
@@ -49,7 +50,8 @@ Suggests:
     future,
     rstudioapi,
     spelling,
-    mockery (>= 0.4.2)
+    mockery (>= 0.4.2),
+    geojsonsf
```
schloerke, line 54: Tests use `sf` as well. This is causing the check failures.

```suggestion
    geojsonsf,
    sf
```

DESCRIPTION
```diff
@@ -10,6 +10,7 @@ Authors@R: c(
   person("Bruno", "Tremblay", role = "ctb", email = "cran@neoxone.com"),
   person("Frans", "van Dunné", role = "ctb", email = "frans@ixpantia.com"),
   person("Sebastiaan", "Vandewoude", role="ctb", email = "sebastiaanvandewoude@gmail.com"),
+  person("Josiah", "Parry", role = "ctb", email = "josiah.parry@gmail.com", comment = c(ORCID = "0000-0001-9910-865X")),
```
schloerke, line 13: Could you add a NEWS entry and tag your GitHub handle with the PR number? Thank you!

```suggestion
```

R/serializer.R
```diff
@@ -231,6 +231,26 @@ serializer_unboxed_json <- function(auto_unbox = TRUE, ..., type = "application/
   serializer_json(auto_unbox = auto_unbox, ..., type = type)
 }
 
+#' @describeIn serializers GeoJSON serializer. See also [geojsonsf::sf_geojson()] and [[geojsonsf::sfc_geojson()]].
+#' @inheritParams geojsonsf::sf_geojson
+#' @inheritParams geojsonsf::sfc_geojson
```
schloerke, line 236: No parameters will be matched. Removing param inheritance.

```suggestion
```

---------------------

Maybe in the description on the line above, mention that the `...` will be passed into `geojsonsf::sfc_geosjon()` or `geojsonsf::sf_geojson()` respectively?

tests/testthat/test-serializer-geojson.R
```diff
@@ -0,0 +1,37 @@
+context("geojson serializer")
+
+test_that("GeoJSON serializes properly", {
+  skip_if_not_installed("geojsonsf")
+  skip_if_not_installed("sf")
```
schloerke, line 5: Thank you for the tests!

tests/testthat/test-serializer-geojson.R
```diff
@@ -0,0 +1,37 @@
+context("geojson serializer")
```
schloerke, line 1:
```suggestion
```

R/serializer.R
```diff
@@ -231,6 +231,26 @@ serializer_unboxed_json <- function(auto_unbox = TRUE, ..., type = "application/
...
+
+      geojsonsf::sfc_geojson(val, ...)
+
+    } else if (any(class(val) == "sf")) {
+
+      geojsonsf::sf_geojson(val, ...)
+
+    } else {
+      stop('Object must be of class "sf" or "sfc"', call. = FALSE)
+    }
```
schloerke, line 250: Two changes:
* Function change: Let's use `inherits(val, classVal)` instead of `any(class(val) == classVal)`
* Cosmetic: I'm a fan of early returns, rather than `else if` of death.

```suggestion
    if (inherits(val, "sfc")) return(geojsonsf::sfc_geojson(val, ...))
    if (inherits(val, "sf"))   return(geojsonsf::sf_geojson(val, ...))
    stop('Object must be of class "sf" or "sfc"', call. = FALSE)
```

schloerke (changes requested):
Notes: https://datatracker.ietf.org/doc/html/rfc7946#section-12 defines `application/geo+json` as the standard Content-Type for geojson data

JosiahParry:
@schloerke thanks for the comments! 
Will do. I'll make your changes and also add the parser.

meztez:
geojsonsf use a different R json library. I wonder if this could be avoided to reduce plumber depencies?

schloerke:
@meztez , do you think these codes should live within the geojson pkg?

I don't remember the setup for `broom`'s integration with other packages. I'm guessing this is closed to that situation

JosiahParry:
@meztez No libraries were added to `Imports` as such the dependency tree remains the same. There is no effect on plumber dependencies.

meztez:
> @meztez No libraries were added to `Imports` as such the dependency tree remains the same. There is no effect on plumber dependencies.

Great! In that case, should we prepend call to geojsonsf function with a requireNamespace check?
https://github.com/rstudio/plumber/search?q=requireNamespace&type=code

@JosiahParry thanks a lot for this contribution

R/parse-body.R
```diff
@@ -386,6 +386,16 @@ parser_json <- function(...) {
   })
 }
 
+#' @describeIn parsers GeoJSON parser. See [geojsonsf::geojson_sf()] for more details.
+#' @export
+parser_geojson <- function(...) {
+  parser_text(function(val) {
+    if (!requireNamespace("geojsonsf", quietly = TRUE)) {
+      stop("`geojsonsf` must be installed for `parser_geojson` to work")
+    }
```
schloerke, line 395: Let's move the require check earlier in the code so the error happens immediately, not when the parser is finally executed.

```suggestion
  if (!requireNamespace("geojsonsf", quietly = TRUE)) {
    stop("`geojsonsf` must be installed for `parser_geojson` to work")
  }
  parser_text(function(val) {
```

R/serializer.R
```diff
@@ -231,6 +231,17 @@ serializer_unboxed_json <- function(auto_unbox = TRUE, ..., type = "application/
...
+#' @export
+serializer_geojson <- function(..., type = "application/geo+json") {
+  serializer_content_type(type, function(val) {
+    if (inherits(val, "sfc")) return(geojsonsf::sfc_geojson(val, ...))
+    if (inherits(val, "sf"))  return(geojsonsf::sf_geojson(val, ...))
+    if (!requireNamespace("geojsonsf", quietly = TRUE)) {
+      stop("`geojsonsf` must be installed for `serializer_geojson` to work")
+    }
+  })
+}
```
schloerke, line 244: Let's move the `requireNamespace()` check up so that if the serializer is called, the error could be thrown immediately.

Let's also add a generic "this isn't a geojson object" error message if not an `sf` or `sfc` object. (Feel free to change the wording)

```suggestion
serializer_geojson <- function(..., type = "application/geo+json") {
  if (!requireNamespace("geojsonsf", quietly = TRUE)) {
    stop("`geojsonsf` must be installed for `serializer_geojson()` to work")
  }
  serializer_content_type(type, function(val) {
    if (inherits(val, "sfc")) return(geojsonsf::sfc_geojson(val, ...))
    if (inherits(val, "sf"))  return(geojsonsf::sf_geojson(val, ...))
    stop("Did not receive a `sf` or `sfc` object")
  })
}
```

R/parse-body.R
```diff
@@ -386,6 +386,16 @@ parser_json <- function(...) {
   })
 }
 
+#' @describeIn parsers GeoJSON parser. See [geojsonsf::geojson_sf()] for more details.
+#' @export
```
schloerke, line 390: This isn't exported in the NAMESPACE. Please run `devtools::document()`

R/parse-body.R
```diff
@@ -564,6 +574,7 @@ register_parsers_onLoad <- function() {
   # yaml types: https://stackoverflow.com/a/38000954/591574
   register_parser("yaml",    parser_yaml,    fixed = c("text/vnd.yaml", "application/yaml", "application/x-yaml", "text/yaml", "text/x-yaml"))
   register_parser("none",    parser_none,    regex = "*")
+  register_parser("geojson", parser_geojson, fixed = c("application/geo+json", "geo+json"))
```
schloerke, line 577: Should we also add `application/vnd.geo+json`?

Removing invalid content type. (Must have `text` or `application` to start)
```suggestion
  register_parser("geojson", parser_geojson, fixed = c("application/geo+json"))
```

schloerke (changes requested):

R/serializer.R
```diff
@@ -231,6 +231,26 @@ serializer_unboxed_json <- function(auto_unbox = TRUE, ..., type = "application/
   serializer_json(auto_unbox = auto_unbox, ..., type = type)
 }
 
+#' @describeIn serializers GeoJSON serializer. See also [geojsonsf::sf_geojson()] and [[geojsonsf::sfc_geojson()]].
+#' @inheritParams geojsonsf::sf_geojson
+#' @inheritParams geojsonsf::sfc_geojson
+#' @export
+serializer_geojson <- function(..., type = "application/geo+json") {
```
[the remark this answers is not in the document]
schloerke: Addressed in lower comment. Resolving.

JosiahParry:
All requested changes should be addressed in most recent commit.

JosiahParry:
Exciting!!! Thanks all for your guidance :)

PR #843, opened 2021-11-26
New Dockerfile based on Rocker Project Ubuntu LTS

mbuet2ner:
Originally, only a Dockerfile based on Rocker Project's Debian Docker image was provided.
This PR also adds a Dockerfile for the Rocker Project's Ubuntu LTS image based on the newest [focal](https://github.com/rocker-org/rocker/tree/master/r-ubuntu) release.

I was able to built the image without any issues and access the mini example API.

PR task list:
- [X] Update NEWS
- [ ] Add tests (does not apply)
- [ ] Update documentation with `devtools::document()` (does not apply)

### DIFF
--- a/Dockerfile_ubuntu
+++ b/Dockerfile_ubuntu
@@ -1,0 +1,62 @@
+
+ARG LTS_VERSION=20.04
+
+FROM rocker/r-ubuntu:${LTS_VERSION}
+LABEL maintainer="barret@rstudio.com"
+
+# BEGIN rstudio/plumber layers
+RUN apt-get update -qq && apt-get install -y --no-install-recommends \
+  git-core \
+  libssl-dev \
+  libcurl4-gnutls-dev \
+  curl \
+  libsodium-dev \
+  libxml2-dev
+
+RUN install2.r remotes
+
+## Remove this comment to always bust the Docker cache at this step
+## https://stackoverflow.com/a/55621942/591574
+#ADD https://github.com/rstudio/plumber/commits/ _docker_cache
+
+ARG PLUMBER_REF=main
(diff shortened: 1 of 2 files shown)


mbuet2ner:
I copied the "readme" comments from the Debian Docker file. However, the `docker build` commands etc. have to be adapted to the new image and the preferred tags. Let me know if you want me to change anything.

schloerke:
I completely agree with the goal of simplifying routines. However, I'm going to close this PR. 

Reasoning:
* `rocker/r-ver` is built on a floating OS that changes to `ubuntu:focal` when `R >= 4.0.0`. 
	* This is not well documented. 
	* By default, `rstudio/plumber`'s Docker file is are pointing to `rocker/r-ver:latest` which [is using R 4.2.0](https://hub.docker.com/layers/r-ver/rocker/r-ver/latest/images/sha256-077e6f5a2d7d8b1f44f8c0da4580edd68983e94a16b2040b3924435e5318b3d7?context=explore) and [ubuntu focal](https://github.com/rocker-org/rocker-versioned2/blob/master/dockerfiles/r-ver_4.2.0.Dockerfile#L1).
* Switching to `rstudio/r-base` (which would be fair to match the other existing rstudio infrastructure) would also install TinyTeX and Pandoc which isn't really needed for most users.
	* `rocker/r-ver` is also pointing their CRAN to the public RSPM for faster binary installs
* If we moved to either `rstudio/r-base` or `rocker/r-ubuntu`, both the R version and operating system would need to be defined.
	* This leads me to more complication than I'd like to have for such a small Docker file example

Thank you for starting the discussion and clear goals!

mbuet2ner:
Makes sense! Thanks for the detailed explanation @schloerke !

Issue #854, opened 2022-03-30
Using parquet parser seems to fail


Giqles:
### System details

Output of `sessioninfo::session_info()()`:

```
R version 4.1.3 (2022-03-10)
Platform: x86_64-pc-linux-gnu (64-bit)
Running under: Ubuntu 20.04.4 LTS

Matrix products: default
BLAS:   /usr/lib/x86_64-linux-gnu/openblas-pthread/libblas.so.3
LAPACK: /usr/lib/x86_64-linux-gnu/openblas-pthread/liblapack.so.3

locale:
 [1] LC_CTYPE=en_US.UTF-8       LC_NUMERIC=C              
 [3] LC_TIME=en_US.UTF-8        LC_COLLATE=en_US.UTF-8    
 [5] LC_MONETARY=en_US.UTF-8    LC_MESSAGES=en_US.UTF-8   
 [7] LC_PAPER=en_US.UTF-8       LC_NAME=C                 
 [9] LC_ADDRESS=C               LC_TELEPHONE=C            
[11] LC_MEASUREMENT=en_US.UTF-8 LC_IDENTIFICATION=C       

attached base packages:
[1] stats     graphics  grDevices utils     datasets  methods   base     

other attached packages:
[1] plumber_1.1.0.9000 grf_2.1.0          data.table_1.14.2  arrow_7.0.0       
[5] qs_0.25.3          jsonlite_1.8.0    

loaded via a namespace (and not attached):
 [1] stringfish_0.15.5   Rcpp_1.0.8.3        magrittr_2.0.2     
 [4] webutils_1.1        tidyselect_1.1.2    bit_4.0.4          
 [7] RApiSerialize_0.1.0 lattice_0.20-45     R6_2.5.1           
[10] rlang_1.0.2         tools_4.1.3         grid_4.1.3         
[13] bit64_4.0.5         RcppParallel_5.1.5  assertthat_0.2.1   
[16] lifecycle_1.0.1     Matrix_1.4-0        purrr_0.3.4        
[19] later_1.3.0         vctrs_0.3.8         promises_1.2.0.1   
[22] glue_1.6.2          stringi_1.7.6       swagger_3.33.1     
[25] compiler_4.1.3
```

### Example application *or* steps to reproduce the problem

This is trying to set up a container to use with Sagemaker batch transforms. From the sagemaker examples, that will be sent as a post like:

```sh
[...]

PR #855, opened 2022-04-08
Update parquet parser

Giqles:
This is a fix for #854.

```r
# insert reprex here
```

PR task list:
- [x] Update NEWS
- [ ] Add tests

### DIFF
--- a/R/parse-body.R
+++ b/R/parse-body.R
@@ -455,8 +455,10 @@ parser_tsv <- function(...) {
 #' @describeIn parsers Helper parser that writes the binary body to a file and reads it back again using `read_fn`.
 #'   This parser should be used when reading from a file is required.
 #' @param read_fn function used to read a the content of a file. Ex: [readRDS()]
+#' @param write_fn function used to write the body content to a file. Defaults to writeBin(), the main alternative is 
+#' readr::write_file(). This can help with some (rare) issues with using writeBin(), and might be faster for larger files.
 #' @export
-parser_read_file <- function(read_fn = readLines) {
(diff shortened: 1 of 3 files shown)


R/parse-body.R
```diff
@@ -483,23 +485,39 @@ parser_rds <- function(...) {
...
+      if (!requireNamespace("arrow", quietly = TRUE)) {
+        stop("`arrow` must be installed for `parser_parquet` to work")
+      }
+      arrow::read_parquet(tmpfile, ...)
+    },
+    function(value, tmpfile) {
+      if (!requireNamespace("readr", quietly = TRUE)) {
+        stop("`readr` must be installed for `parser_parquet` to work")
+      }
+      readr::write_file(value, tmpfile)
```
schloerke, line 518: What advantage is there to use `readr::write_file()` over `writeBin()`?

schloerke: Do you have an example of a _rare_ issue?

schloerke:
@Giqles Any other changes you'd like to do to make this PR not a draft?

R/parse-body.R
Giqles: I was trying to write a test to for the issue (I thought) I was having with the `writeBin` implementation. The original data I was having this issue (#854) with is not public, and some alternatives I've tried didn't seem to reproduce either. 

I've just gone back to the machine I was trying this on, and it doesn't reproduce there either -- I think this must've been caused by something else. I'll close this PR for now.

Giqles:
Closing this for now. Does not reproduce consistently.

PR #858, opened 2022-05-04
Allow `path` specification in `pr_cookie()`

jtlandis:
This pull request will address #850 as the user will be able to specify the Path attribute of the cookie.

This attribute was already present in the main branches `res$setCookie` function, I was merely exposing this code. As such all edits were made such that default calls to `pr_cookie()` will not affect backwards compatibility.

I haven't yet added tests as I do not know how to test plumber endpoints in `testthat`.

Only one test failed when running `devtools::check()` and that was `test-serializer-htmlwidgets.R` testing "htmlwidgets get serialized properly".

PR task list:
- [x] Update NEWS
- [ ] Add tests
- [x] Update documentation with `devtools::document()`

PR #862, opened 2022-06-03
Add `pr_cookie(path=)`

### LINKED ISSUE #850: Let pr_cookie set cookie path for sharing sessions between multiple mounted routers
Multiple routers could share a session if pr_cookie allows to set an explicit path in the cookie to the common root of all routers, e.g. "/".  So having this parameter would be nice.

Currently when multiple routers are mounted, they cannot see each other's cookies and thus sessions. This is because each router sets a cookie whose visibility is restricted to each router's own URI path. Pr_cookie sets no path information with the cookie, and the restriction is the default browser behaviour then.

schloerke:
Continuation of https://github.com/rstudio/plumber/pull/858
Fixes #850 

PR task list:
- [x] Update NEWS
- [x] Add tests


schloerke:
Reviewed with `@rich-iannone` and `@wch`

PR #870, opened 2022-07-08
Debug `detritus` temp files bug on windows devel

schloerke:
PR task list:
- [x] Pass on `devtools::check_win_devel()`
- [x] Add tests


schloerke:
The win devel note of 
```
    * checking for detritus in the temp directory ... NOTE
    Found the following files/directories:
      'Rscript114b08aee34ea' 'Rscript4b588aee34ea'
```

is caused by a similar situation of https://stat.ethz.ch/pipermail/r-devel/2021-June/080830.html. (Basically starting a `{future}` process and not explicitly killing it / cleaning it up)

while it is not involving `{plumber}` directly, it does impact `{plumber}`'s submission. 

Solution: Disable the test for windows r >= 4.0. If it `plumb()`s for other platforms, I am ok with it not being tested on recent versions of R on windows

schloerke:
Passes on win devel; Merging

PR #871, opened 2022-07-12
Tweak Dockerfile; Reduce final size

eitsupi:
- Use the `org.opencontainers.image.authors` annotation instead of the `maintainer` label. (See <https://docs.docker.com/engine/reference/builder/#maintainer-deprecated>)
- Remove temp files to reduce image size. (c.f. Dockerfile best practices <https://docs.docker.com/develop/develop-images/dockerfile_best-practices/#run>)
  - `apt`'s cache `/var/lib/apt/lists/*`
  - `install2.r`'s cache `/tmp/downloaded_packages`
- Fix docs link

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -2,7 +2,6 @@
 ARG R_VERSION=latest
 
 FROM rocker/r-ver:${R_VERSION}
-LABEL maintainer="barret@rstudio.com"
 LABEL org.opencontainers.image.authors="barret@rstudio.com"
 
 # BEGIN rstudio/plumber layers
--- a/Dockerfile
+++ b/Dockerfile
@@ -5,6 +5,8 @@ FROM rocker/r-ver:${R_VERSION}
 LABEL org.opencontainers.image.authors="barret@rstudio.com"
 
 # BEGIN rstudio/plumber layers
+
+# `rm` call removes `apt` cache
 RUN apt-get update -qq && apt-get install -y --no-install-recommends \
   git-core \
   libssl-dev \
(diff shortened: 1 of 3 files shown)


schloerke (changes requested):
@eitsupi Awesome! Can you add two comments about the two `rm -rf` statements? I can see them performing them in the docs, but I don't immediately see what it is for.

Thank you!

eitsupi:
> Can you add two comments about the two `rm -rf` statements?

I updated the PR comment, but I didn't know what you wanted.
Do "two comments" mean comments in the source code?

Dockerfile
```diff
@@ -11,9 +11,12 @@ RUN apt-get update -qq && apt-get install -y --no-install-recommends \
   libcurl4-gnutls-dev \
   curl \
   libsodium-dev \
-  libxml2-dev
+  libxml2-dev \
+  && rm -rf /var/lib/apt/lists/*
 
-RUN install2.r remotes
+RUN install2.r --error --skipinstalled --ncpus -1 \
```
schloerke, line 17: Add Dockerfile comment here about the `rm -rf`

Dockerfile
```diff
@@ -2,7 +2,7 @@
 ARG R_VERSION=latest
 
 FROM rocker/r-ver:${R_VERSION}
-LABEL maintainer="barret@rstudio.com"
+LABEL org.opencontainers.image.authors="barret@rstudio.com"
 
 # BEGIN rstudio/plumber layers
 RUN apt-get update -qq && apt-get install -y --no-install-recommends \
```
schloerke, line 8: Add Dockerfile comment here about the `rm -rf`

schloerke:
```suggestion

# `rm` call removes `apt` cache
RUN apt-get update -qq && apt-get install -y --no-install-recommends \
```

Dockerfile
schloerke:
```suggestion
# `rm` call removes install2.r's cache
RUN install2.r --error --skipinstalled --ncpus -1 \
```

schloerke (approved):

schloerke:
Thank you!

PR #882, opened 2022-09-28
Mount docs as first mount

### LINKED ISSUE #881: root static file handler causes /__docs__/ to 404
### System details

Output of `sessioninfo::session_info()()`:

```
# sessioninfo::session_info() output goes here
```

The `plumber.R` file.

```r
#* @assets ./static /
list()

#* Say hello.
#* @get /hello
function() {
  list(msg = "Hello.")
}
```

The `static/index.html` file.

```html
<html>
<head>
<title>static</title>
</head>
<body>
this is static.
</body>
</html>
```

This is the R command run by the RStudio IDE "Run API" button, but it can also be run from the R console.

```R
plumber::plumb(file='plumber.R')$run()
# => Running plumber API at http://127.0.0.1:6385
# => Running swagger Docs at http://127.0.0.1:6385/__docs__/
```

The RStudio IDE attempts to launch the documentation in its Viewer pane, which shows the 404 message. You can also see the 404 from `curl`:

```bash
curl http://127.0.0.1:6385/__docs__/
# => {"error":"404 - Resource Not Found"}
```

The `/__docs__/` route responds with a 404 rather than serving the API documentation. The server indicates that documentation is available during startup.

schloerke:
Fixes #881 

Other changes:
* working directory is restored last, not before the `exit` hooks are run.

PR task list:
- [ ] Update NEWS
- [x] Add tests

cc @aronatkins

### DIFF
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -269,6 +269,8 @@ Plumber <- R6Class(
           quiet = quiet
         )
         # Unmount the docs before restoring the wd
+        # Unmount needs to happen after exit hooks
+        # No guarantee that new exit hooks are not added after running API
         on.exit(unmount_docs(self, docs_info), add = TRUE)
       }
 
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -255,6 +255,9 @@ Plumber <- R6Class(
         old_wd <- setwd(dirname(private$filename))
       }
 
+      # Run user exit hooks given docs and wd are still set
+      on.exit(private$runHooks("exit"), add = TRUE)
+
       if (isTRUE(docs_info$enabled)) {
         mount_docs(
           pr = self,
@@ -265,12 +268,14 @@ Plumber <- R6Class(
           callback = swaggerCallback,
           quiet = quiet
         )
+        # Unmount the docs before restoring the wd
         on.exit(unmount_docs(self, docs_info), add = TRUE)
-        on.exit({
-        }, add = TRUE)
       }
 
       # Restore the prior working directory after exit hooks have run
+      if (!is.null(old_wd)) {
+        on.exit(setwd(old_wd), add = TRUE)
+      }
 
       httpuv::runServer(host, port, self)
     },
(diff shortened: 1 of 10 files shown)


R/plumber.R
```diff
@@ -250,24 +250,34 @@ Plumber <- R6Class(
       }
 
       # Set and restore the wd to make it appear that the proc is running local to the file's definition.
+      old_wd <- NULL
       if (!is.null(private$filename)) {
         old_wd <- setwd(dirname(private$filename))
-        on.exit({setwd(old_wd)}, add = TRUE)
```
jcheng5, line 255: Not sure what you're doing here with the `on.exit` changes, but presuming it has to do with the order of the cleanup happening backwards? Have you considered `on.exit(..., add = TRUE, after = FALSE)`?

jcheng5:
Sorry, I don't understand these changes--can you walk me through them tomorrow?

R/plumber.R
schloerke: I didn't know about `after=`!

schloerke:
Will do!

jcheng5:
I suggest adding a `prepend = TRUE` argument to `pr$mount()` and then not having to pass `pr_private`

R/plumber.R
```diff
@@ -299,7 +307,15 @@ Plumber <- R6Class(
         path <- paste0(path, "/")
       }
 
-      private$mnts[[path]] <- router
+      # Remove prior mount if it exists
+      self$unmount(path)
```
jcheng5, line 311: I don't think you want to do this unmount...? With https://github.com/rstudio/plumber/pull/883, you'll be able to mount two subrouters to one path and have one fall through to the other.

schloerke: I'll be sure to address that in the PR. There's a lot of decisions to be made in #883 

For now, I'd like to keep existing behavior where it was overwritten.

jcheng5: And if you still do want to unmount, I think you would then need to factor in how it affects the `after` index if it's a non-NULL, non-zero value.

schloerke: Hmmmm. Let's keep this in draft to see if we're even going to merge #883

PR #883, opened 2022-09-29
Mounted router fall back to parent router when route is not found

schloerke:
A different solution to #881 

Situation:
* Static handler is mounted at `/`;
* Route `/b/z` is mounted at `/a`
* Route `/c` is mounted at `/a/b`

Goal: Request `/a/b/c`

Current behavior: 
* The route of `/a/b/c` matches against mount location of `/`.
* The static file handler attempts to handle `/a/b/c`. 
	* No file is found, a 404 is returned.

New behavior: 
* The route of `/a/b/c` matches against mount location of `/`.
* The static file handler attempts to handle `/a/b/c`
	* No file is found, route is forwarded
	* Mount's `$route()` is marked as `routeNotFound()`
* The route of `/a/b/c` matches against mount location `/a`
	* The route is trimmed to `/b/c`
	* The route `/b/c` does not match any routes
	* Mount's `$route()` is marked as `routeNotFound()`
* The route of `/a/b/c` matches against mount location `/a/b`
	* The route is trimmed to `/c`
	* The route `/c` matches route location `/c`
	* Result of `/c` is returned

This would allow for #882 to not be merged, keep the docs as the last mount location. This benefits the user in that their content will always have preference over anything mounted in `/__docs__/`

PR task list:
- [ ] Should this behavior be opt in? Opt out?
- [ ] How should the user opt in / out? 
	* `$set_fallback(fallback = TRUE)`?
	* `$mount(path, router, fallback = TRUE)`?
- [ ] Update NEWS
- [x] Add tests

### DIFF
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -768,12 +768,31 @@ Plumber <- R6Class(
           resetForward()
           # TODO: support globbing?
 
-          if (nchar(path) >= nchar(mountPath) && substr(path, 0, nchar(mountPath)) == mountPath) {
-            # This is a prefix match or exact match. Let this router handle.
+          # Keep track of how many mount levels deep we are.
+          # Can not use a boolean as a nested mount will unwrap before the parent mount is done.
+          # Using a counter allows us to know when we are at the top level router (0).
+          if (is.null(req$`_MOUNT_COUNT`)) {
+            req$`_MOUNT_COUNT` <- 0
+          }
 
+          if (nchar(path) >= nchar(mountPath) && substr(path, 0, nchar(mountPath)) == mountPath) {
+            # This is a prefix match or exact match. Let mount attempt handle.
+
+            # Mark that the route is within a mount. Allows for the mount to forward instead of 404.
+            req$`_MOUNT_COUNT` <- req$`_MOUNT_COUNT` + 1
             # First trim the prefix off of the PATH_INFO element
+            curPathInfo <- req$PATH_INFO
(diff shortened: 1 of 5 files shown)


jcheng5:
I suggest removing the `IS_MOUNT` tracking variable (tricky to make this work with async responses anyway) and instead have the `route()` method (which is recursive) never do 404 handling; instead, leave that to `serve()` (which is not recursive).

R/plumber.R
```diff
@@ -767,13 +767,31 @@ Plumber <- R6Class(
...
+            cur_path_info <- req$PATH_INFO
             req$PATH_INFO <- substr(req$PATH_INFO, nchar(mountPath), nchar(req$PATH_INFO))
-            return(private$mnts[[mountPath]]$route(req, res))
+
+            # Handle route
+            ret <- private$mnts[[mountPath]]$route(req, res)
+
+            # Undo path info and mount status changes
+            req$PATH_INFO <- cur_path_info
+            req$`_IS_MOUNT` <- prev_mount_status
```
jcheng5, line 787: req$PATH_INFO reset needs to happen in a separate step (so that async is supported)

jcheng5:
As written, you can choose to either have 1) 405 and missing-trailing-slash automatically handled for you, or 2) a custom top-level 404 handler, but you cannot have both.

schloerke:
Going to wait on this PR for a bit as the changes are fairly drastic compared to current functionality

PR #884, opened 2022-10-05
make fastmap dependency explicit

edavidaja:
`fastmap` must be available to call `promises::future_promises()`; since it is only a `Suggests`ed dependency, renv / packrat will not identify it as necessary, causing runtime failures.

```r
r$> renv::dependencies("inst/plumber/14-future/plumber.R")
Finding R package dependencies ... Done!
                                                                               Source  Package Require Version   Dev
1 /Users/edavidaja/Documents/rstudiopbc/open/plumber/inst/plumber/14-future/plumber.R     coro                 FALSE
2 /Users/edavidaja/Documents/rstudiopbc/open/plumber/inst/plumber/14-future/plumber.R   future                 FALSE
3 /Users/edavidaja/Documents/rstudiopbc/open/plumber/inst/plumber/14-future/plumber.R promises                 FALSE
```

PR task list:
- [ ] Update NEWS
- [ ] Add tests

### DIFF
--- a/inst/plumber/14-future/plumber.R
+++ b/inst/plumber/14-future/plumber.R
@@ -1,6 +1,7 @@
 
 library(promises)
 library(coro)
+library(fastmap)
 future::plan("multisession") # a worker for each core
 # future::plan(future::multisession(workers = 2)) # only two workers
 


edavidaja:
I caught it in deployment when I implemented an API based on this example and couldn't figure out why it was returning 500s on Connect even though it worked locally; I think fastmap was in my local environment but the dependency doesn't get detected by renv / packrat.

Issue #869, opened 2022-06-28
"boolean" param not defaulting consistently


r2evans:
The selector pulldowns for `:boolean` class will default to `true` when true but `--` (not `false`) when false. Can this be fixed? My guess is that the `FALSE` is being interpreted as "do not display the default value" instead of "display this of the candidate values".

### Steps to demonstrate/reproduce

```R
#* @get /quux
#* @param regex:boolean Whether the pattern is regex (true) or glob (false).
function(regex = FALSE) {
  regex
}
```

![image](https://user-images.githubusercontent.com/5815808/176263681-8dee6940-9f1f-4ed4-b1c1-786b04f07712.png)

```r
#* @get /quux
#* @param regex:boolean Whether the pattern is regex (true) or glob (false).
function(regex = TRUE) {
  regex
}
```

![image](https://user-images.githubusercontent.com/5815808/176264094-e8304bd1-054b-4470-8046-277474dd51a9.png)

### System details

Output of `sessioninfo::session_info()()`:

```
sessioninfo::session_info()
# - Session info  --------------------------------------------------------------
#  hash: parrot, open hands: dark skin tone, man firefighter: medium skin tone
#  setting  value
#  version  R version 4.1.2 (2021-11-01)
#  os       Windows 10 x64 (build 22000)
#  system   x86_64, mingw32
#  ui       RTerm
#  language (EN)
#  collate  English_United States.1252
#  ctype    English_United States.1252
#  tz       America/New_York
#  date     2022-06-28
#  pandoc   2.17.1.1 @ C:/Users/r2/AppData/Local/Pandoc/ (via rmarkdown)
# - Packages -------------------------------------------------------------------
#  package     * version date (UTC) lib source
#  cli           3.3.0   2022-04-25 [1] CRAN (R 4.1.3)
#  digest        0.6.28  2021-09-23 [1] CRAN (R 4.1.2)
#  evaluate      0.14    2019-05-28 [1] CRAN (R 4.1.2)
#  fastmap       1.1.0   2021-01-25 [1] CRAN (R 4.1.2)
#  htmltools     0.5.2   2021-08-25 [1] CRAN (R 4.1.2)
#  jsonlite      1.8.0   2022-02-22 [1] CRAN (R 4.1.3)
#  knitr         1.36    2021-09-29 [1] CRAN (R 4.1.2)
#  later         1.3.0   2021-08-18 [1] CRAN (R 4.1.2)
[...]

Issue #762, opened 2021-02-05
Parse parameters if a OpenAPI type is specified


schloerke:
Would need to undo https://github.com/rstudio/plumber/pull/756 , but it is fair that this be the default behavior.

If nothing is specified, then no type conversion should be attempted.

meztez:
That makes a lot of sense

schloerke:
This is a breaking change. But I believe it is a welcome one

slodge:
Is help still wanted on this?

Because I'd love to see this happen and might be able to help...

schloerke:
@slodge Help is still wanted

slodge:
Looking ... will start with working out what steps and hooks are :)

If there's anything I should be reading re: contributing or general coding guidelines for this repo then please point me at them - otherwise I'll just try to "fit in" 👍

schloerke:
There has been a lot of work in https://github.com/rstudio/plumber/blob/06e46f3ff5119e5f1cb8af29ef49aecb3cbb932a/R/parse-query.R#L141

Currently only the path params are being coerced: https://github.com/rstudio/plumber/blob/76cecc58925897c91cb88203ab485a0c512a0d26/R/plumber-step.R#L278-L280

Will prolly need to parse `req$argsQuery `/`req$args` and `req$argsBody` as well: https://github.com/rstudio/plumber/blob/06e46f3ff5119e5f1cb8af29ef49aecb3cbb932a/R/plumber.R#L697-L702

But prolly may just want to perform the conversions in https://github.com/rstudio/plumber/blob/d932ac8fdaad9a5613a484eac9c088b827bbd0b4/R/plumber-step.R#L119 before the value is returned: https://github.com/rstudio/plumber/blob/d932ac8fdaad9a5613a484eac9c088b827bbd0b4/R/plumber-step.R#L164 . 

----------------------------

I guess we need to double check if we _should_ convert query parameters or body parameters. I'm leaning towards maybe only query parameters as body parameters should be parsed already. Query parameters are always strings. This makes sense as path parameters would behave the same way (only as strings).

@slodge Could you parse the query parameters near here: https://github.com/rstudio/plumber/blob/06e46f3ff5119e5f1cb8af29ef49aecb3cbb932a/R/parse-query.R#L6 . Can we leverage the converters from above?

------------------

[...]

PR #889, opened 2022-11-29
Converting parameter values to their openapi specified types

slodge:
This is a PR for #762  - automagically converting parameters to their openApi types for all endpoints

## PR task list:

Functional:
- [ ] Discuss design and make sure the maintainer is happy...
- [ ] Discuss implementation and make sure the maintainer is happy...

Admin:

- [x] Sign RStudio CLA 
- [ ] Update NEWS
- [ ] Update articles
- [ ] Add tests
  - [x] non-typed parameters
  - [x] typed parameters
  - [ ] parameters which don't parse
- [ ] Run `devtools::check()`

```r
# Just a demo Plumber API. 
devtools::load_all()

#* Echo back the input
#* @get /echo
function(msg = "") {
}

#* Plot a histogram
#* @serializer png
#* @get /plot
function() {
    rand <- rnorm(100)
    hist(rand)
}

#* Return the sum of two numbers
#* @param a:int The first number to add
#* @param b:int The second number to add
#* @get /sum
function(a, b) {
  message("/sum processing a:", a, " of class ", class(a), " and b:", b, " of class ", class(b))
  a + b
}


#* Return the sum of two sets of numbers
#* @param a:[float] The first set of numbers to add
#* @param b:[float] The second set of numbers to add
#* @get /sumAll
function(a, b) {
  message("/sumAll processing a:", a, " of class ", class(a), " and b:", b, " of class ", class(b))
  sum(a, b)
}

#* Return a filtered set of strings
#* @param a:[logical] The filter map
#* @param b:[string] The set of string to filter
#* @get /filterText
function(a, b) {
  browser()
  message("/filterText processing a:", a, " of class ", class(a), " and b:", b, " of class ", class(b))
  b[a]
}

#* Is this working?
#* @param yes:bool Yes?
#* @get /boolTest
function(yes) {
  browser()
  if (yes) {
    "YES!"
  } else {
    "How sad.."
  }
}

# Programmatically alter your API
#* @plumber
function(pr) {
    pr %>%
        # Overwrite the default serializer to return unboxed JSON
        pr_set_serializer(serializer_unboxed_json())
}
```

Current file used for "manual testing" so far is:

```
> plumb(file='ExampleApi/plumber.R')$run(port = 7638)
ℹ Loading plumber
Running plumber API at http://127.0.0.1:7638
Running swagger Docs at http://127.0.0.1:7638/__docs__/

curl -X POST "http://127.0.0.1:7638/sum?a=1&b=22" -H "accept: */*" -d ""
/sum processing a:1 of class integer and b:22 of class integer

curl -X POST "http://127.0.0.1:7638/sumAll?a=1&a=33&a=22&b=44&b=11&b=-123" -H "accept: */*" -d ""
/sumAll processing a:13322 of class numeric and b:4411-123 of class numeric

http://127.0.0.1:7638/filterText?a=true&a=false&a=false&b=life&b=is%2Clike%2Ca&b=box
/filterText processing a:TRUEFALSEFALSE of class logical and b:lifeis,like,abox of class character
```

**Note**: when using the swagger page, I'm seeing some issues with selecting bools - perhaps similar to #869

### DIFF
--- a/R/options_plumber.R
+++ b/R/options_plumber.R
@@ -41,10 +41,12 @@
 #' \item{`plumber.legacyRedirects`}{Plumber will redirect legacy route `/__swagger__/` and
 #' `/__swagger__/index.html` to `../__docs__/` and `../__docs__/index.html`. You can disable this
 #' by settings this option to `FALSE`. Defaults to `TRUE`}
+#' \item{`plumber.typedParameters`}{Plumber will convert all parameters according to openApi
+#' supplied type hints.  Defaults to`TRUE`}
 #' }
 #'
 #' @param ... Ignored. Should be empty
-#' @param port,docs,docs.callback,trailingSlash,methodNotAllowed,apiScheme,apiHost,apiPort,apiPath,apiURL,maxRequestSize,sharedSecret,legacyRedirects See details
+#' @param port,docs,docs.callback,trailingSlash,methodNotAllowed,apiScheme,apiHost,apiPort,apiPath,apiURL,maxRequestSize,sharedSecret,legacyRedirects,typedParameters See details
 #' @return
 #' The complete, prior set of [options()] values.
 #' If a particular parameter is not supplied, it will return the current value.
@@ -64,7 +66,8 @@ options_plumber <- function(
   apiPath              = getOption("plumber.apiPath"),
   maxRequestSize       = getOption("plumber.maxRequestSize"),
   sharedSecret         = getOption("plumber.sharedSecret"),
-  legacyRedirects      = getOption("plumber.legacyRedirects")
+  legacyRedirects      = getOption("plumber.legacyRedirects"),
+  typedParameters      = getOption("plumber.typedParameters")
 ) {
   ellipsis::check_dots_empty()
 
@@ -86,6 +89,7 @@ options_plumber <- function(
     plumber.apiPath              =   apiPath,
     plumber.maxRequestSize       =   maxRequestSize,
     plumber.sharedSecret         =   sharedSecret,
-    plumber.legacyRedirects      =   legacyRedirects
+    plumber.legacyRedirects      =   legacyRedirects,
+    plumber.typedParameters      =   typedParameters
   )
 }
--- a/R/plumber-step.R
+++ b/R/plumber-step.R
@@ -346,35 +346,24 @@ PlumberEndpoint <- R6Class(
         return(args)
       }
 
-      # TODO - I suspect someone with more base-R could turn this
-      # for into some wonderful s/lapply thing...
-      converted_args <- list()
-
-      for (i in 1:length(args)) {
-        key <- args_names[[i]]
-        arg <- args[[i]]
-        if (key == "") {
-          # not sure why... but we allow unnamed arguments through
-          converted_args <- c(converted_args, arg)
-          next;
-        }
-
-        param <- params[[key]]
-        converter <- identity
-        if (!is.null(param)) {
-          # NOTE: areArrays is FALSE here as we do not want to parse comma
-          #    separated strings (or do we?)
-          converter <- typesToConverters(param$type, areArrays = FALSE)
-          if (is.null(converter)) {
-            converter <- identity
-          } else {
-            converter <- converter[[1]]
+      converted_args <-
+        Map(
+          key = args_names,
+          arg = args,
+          param = params[args_names],
+          f = function(key, arg, param) {
+            # not sure why... but we allow unnamed arguments through
+            if (key == "") return(arg)
+            if (is.null(param)) return(arg)
+            
+            # NOTE: areArrays is FALSE here as we do not want to parse comma
+            #    separated strings (or do we?)
+            converter <- typesToConverters(param$type, areArrays = FALSE)[[1]]
+            if (is.null(converter)) return(arg)
+            converter(arg)
           }
-        }
-        new_args <- list()
-        new_args[[key]] <- converter(arg)
-        converted_args <- c(converted_args, new_args)
-      }
+        )
+      names(converted_args) <- args_names
 
       converted_args
     }
--- a/R/plumber-step.R
+++ b/R/plumber-step.R
@@ -61,6 +61,13 @@ PlumberStep <- R6Class(
         self$serializer <- serializer
       }
     },
+    #' @description convert args to specific types
+    #' @param args List of args to convert
+    convertArgs = function(args) {
+      # a general step doesn't convert args
+      # but a subclass can choose to override and convert
+      args
+    },
     #' @description step execution function
     #' @param req,res Request and response objects created by a Plumber request
     exec = function(req, res) {
@@ -81,7 +88,8 @@ PlumberStep <- R6Class(
       execStep <- function(...) {
         private$runHooksAround("aroundexec", args_for_formal_matching(), .next = function(...) {
           relevant_args <- getRelevantArgs(list(...), func = private$func)
-          do.call(private$func, relevant_args, envir = private$envir)
+          converted_args <- self$convertArgs(relevant_args)
+          do.call(private$func, converted_args, envir = private$envir)
         })
       }
       postexecStep <- function(value, ...) {
@@ -307,6 +315,37 @@ PlumberEndpoint <- R6Class(
       self$path <- path
       private$regex <- createPathRegex(path, self$getFuncParams())
       path
+    },
+    #' @description convert args to specific types
+    #' @param args List of args to convert
+    convertArgs = function(args) {
+      if (!isTRUE(getOption("plumber.typedParameters", TRUE))) {
+        return(args)
+      }
+
+      params <- self$getEndpointParams()
+
+      # TODO - I suspect someone with more base-R could turn this
+      # for into some wonderful s/lapply thing...
+      converted_args <- list()
+      for (i in names(args)) {
+        arg <- args[[i]]
+        param <- params[[i]]
+        converter <- identity
+        if (!is.null(param)) {
+          # NOTE: areArrays is FALSE here as we do not want to parse comma
+          #    separated strings (or do we?)
+          converter <- typesToConverters(param$type, areArrays = FALSE)
+          if (is.null(converter)) {
+            converter <- identity
+          } else {
+            converter <- converter[[1]]
+          }
+        }
+        converted_args[[i]] <- converter(arg)
+      }
+
+      converted_args
     }
   ),
   private = list(
--- a/R/plumber-step.R
+++ b/R/plumber-step.R
@@ -335,7 +335,7 @@ PlumberEndpoint <- R6Class(
 
       args_names <- names(args)
       if (is.null(args_names)) {
-        # no names - won't be able to dp any conversions
+        # no names - won't be able to do any conversions
         return(args)
       }
 
--- a/R/plumber-step.R
+++ b/R/plumber-step.R
@@ -323,14 +323,43 @@ PlumberEndpoint <- R6Class(
         return(args)
       }
 
+      if (is.null(args)) {
+        # no args to convert
+        return(args)
+      }
+
+      if (length(args) == 0) {
+        # no args to convert
+        return(args)
+      }
+
+      args_names <- names(args)
+      if (is.null(args_names)) {
+        # no names - won't be able to dp any conversions
+        return(args)
+      }
+
       params <- self$getEndpointParams()
+
+      if (length(params) == 0) {
+        # no information to convert by
+        return(args)
+      }
 
       # TODO - I suspect someone with more base-R could turn this
       # for into some wonderful s/lapply thing...
       converted_args <- list()
-      for (i in names(args)) {
+
+      for (i in 1:length(args)) {
+        key <- args_names[[i]]
         arg <- args[[i]]
-        param <- params[[i]]
+        if (key == "") {
+          # not sure why... but we allow unnamed arguments through
+          converted_args <- c(converted_args, arg)
+          next;
+        }
+
+        param <- params[[key]]
         converter <- identity
         if (!is.null(param)) {
           # NOTE: areArrays is FALSE here as we do not want to parse comma
@@ -342,7 +371,9 @@ PlumberEndpoint <- R6Class(
             converter <- converter[[1]]
           }
         }
-        converted_args[[i]] <- converter(arg)
+        new_args <- list()
+        new_args[[key]] <- converter(arg)
+        converted_args <- c(converted_args, new_args)
       }
 
       converted_args
--- a/R/plumber-step.R
+++ b/R/plumber-step.R
@@ -352,7 +352,7 @@ PlumberEndpoint <- R6Class(
           arg = args,
           param = params[args_names],
           f = function(key, arg, param) {
-            # not sure why... but we allow unnamed arguments through
+            # Early return on unnamed args or `NULL` values
             if (key == "") return(arg)
             if (is.null(param)) return(arg)
             
--- a/tests/testthat/test-plumber.R
+++ b/tests/testthat/test-plumber.R
@@ -2,7 +2,10 @@ context("Plumber")
 
 exec_endpoint <- function(pr, ep_pos, subset = 1) {
   # This is a poor setup of `req` and `res`. But it works for testing purposes
-  pr$endpoints[[subset]][[ep_pos]]$exec(make_req(), PlumberResponse$new())
+  pr$endpoints[[subset]][[ep_pos]]$exec(
+    as.list(make_req()),
+    PlumberResponse$new()
+  )
 }
 
 test_that("Endpoints are properly identified", {
--- a/tests/testthat/test-plumber.R
+++ b/tests/testthat/test-plumber.R
@@ -2,10 +2,7 @@ context("Plumber")
 
 exec_endpoint <- function(pr, ep_pos, subset = 1) {
   # This is a poor setup of `req` and `res`. But it works for testing purposes
-  pr$endpoints[[subset]][[ep_pos]]$exec(
-    as.list(make_req()),
-    PlumberResponse$new()
-  )
+  pr$endpoints[[subset]][[ep_pos]]$exec(make_req(), PlumberResponse$new())
 }
 
 test_that("Endpoints are properly identified", {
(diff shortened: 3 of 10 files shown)


slodge:
This test isn't working for me 

https://github.com/rstudio/plumber/blob/76cecc58925897c91cb88203ab485a0c512a0d26/tests/testthat/test-options.R#L79

Should it be:
```
      expect_null(getOption("plumber.swagger.url"))
      expect_null(getOption("plumber.docs.callback"))
```

slodge:
I'm also looking at some R4.2 && complaints 

https://github.com/rstudio/plumber/blob/76cecc58925897c91cb88203ab485a0c512a0d26/tests/testthat/test-options.R#L31

Can I adjust to:
```
  for (r_file in dir(r_folder, full.names = T)) {
    file_content <- paste0(readLines(r_file, warn = F), collapse = "")
    match <- stringi::stri_match_all_regex(file_content, "getOption\\([^,\\)]+,?\\)?")[[1]][,1]
    match <- gsub("\\s", "", match)
    match <- match[!is.na(match)]
    if (length(match) > 0) {
      matches <- c(matches, match)
    }
  }
```

schloerke:
For specific code changes, please feel free to make them. For follow up
questions, add a comment to the PR’s code from within the Files tab. It
makes it much easier to compare and resolve any questions.

If something is unrelated to the PR’s goal (ex the option change above),
please comment on it in the Files tab.

Thank you!

slodge:
:+1:

I think I'm actually done on design and implementation... so it's over to you/rstudio whether this is the right sort of approach for Plumber... then:
- (assuming it is) there's just a few more tests and docs to write.
- otherwise... what need's doing?

Have added some initial tests. Will step back now and give you a chance to review/think 👍

R/options_plumber.R
```diff
@@ -41,10 +41,12 @@
 #' \item{`plumber.legacyRedirects`}{Plumber will redirect legacy route `/__swagger__/` and
 #' `/__swagger__/index.html` to `../__docs__/` and `../__docs__/index.html`. You can disable this
 #' by settings this option to `FALSE`. Defaults to `TRUE`}
+#' \item{`plumber.typedParameters`}{Plumber will convert all parameters according to openApi
+#' supplied type hints.  Defaults to`TRUE`}
```
slodge, line 45: One option here might be to default to FALSE - this would keep existing functionality intact... and the option could be changed to TRUE when Plumber is ready to make a bigger version release...

(although personally I want this set to TRUE in every API I write 👍 )

R/plumber-step.R
```diff
@@ -307,6 +315,37 @@ PlumberEndpoint <- R6Class(
...
+    #' @param args List of args to convert
+    convertArgs = function(args) {
+      if (!isTRUE(getOption("plumber.typedParameters", TRUE))) {
+        return(args)
+      }
+
+      params <- self$getEndpointParams()
+
+      # TODO - I suspect someone with more base-R could turn this
+      # for into some wonderful s/lapply thing...
```
slodge, line 329: One option here would be to start embracing tidyverse... I'd be so much better at writing this code in `purrr` :)

R/plumber-step.R
```diff
@@ -307,6 +315,37 @@ PlumberEndpoint <- R6Class(
...
+          # NOTE: areArrays is FALSE here as we do not want to parse comma
+          #    separated strings (or do we?)
+          converter <- typesToConverters(param$type, areArrays = FALSE)
+          if (is.null(converter)) {
+            converter <- identity
+          } else {
+            converter <- converter[[1]]
+          }
+        }
+        converted_args[[i]] <- converter(arg)
```
slodge, line 345: if we had access to a `can_convert` as well as a `converter` here then we could trigger a 400 from here... although I'm not entirely sure how I'd trigger a response from this function - I guess we could return some kind of union of *args plus errors* from this function and then `plumberStep$exec` could return a 400 error `res` instead of calling the actual endpoint handler? I might need a little nudge in the right direction on that code (or someone else can code it)

slodge:
Current test output is https://gist.github.com/slodge/174e6a6d27785d5c01abeecdd531aad0

R/plumber-step.R
```diff
@@ -307,6 +315,68 @@ PlumberEndpoint <- R6Class(
...
+      # TODO - I suspect someone with more base-R could turn this
+      # for into some wonderful s/lapply thing...
+      converted_args <- list()
+
+      for (i in 1:length(args)) {
+        key <- args_names[[i]]
+        arg <- args[[i]]
+        if (key == "") {
+          # not sure why... but we allow unnamed arguments through
+          converted_args <- c(converted_args, arg)
```
schloerke, line 358: > `            # not sure why... but we allow unnamed arguments through`

Do you have an example? Or where did you see this in other code?

I'd understand empty values, but not empty keys.  `?=1` .. doesn't immediately make sense to me as it being valid.

R/plumber-step.R
```diff
@@ -61,6 +61,13 @@ PlumberStep <- R6Class(
         self$serializer <- serializer
       }
     },
+    #' @description convert args to specific types
+    #' @param args List of args to convert
+    convertArgs = function(args) {
+      # a general step doesn't convert args
+      # but a subclass can choose to override and convert
```
schloerke, line 68: Clever. But I think I'd like to do this earlier in the process. I'll see if I can find a better location. (It definitely won't be cleaner than this.)

R/plumber-step.R
```diff
@@ -307,6 +315,68 @@ PlumberEndpoint <- R6Class(
...
+          # not sure why... but we allow unnamed arguments through
+          converted_args <- c(converted_args, arg)
+          next;
+        }
+
+        param <- params[[key]]
+        converter <- identity
+        if (!is.null(param)) {
+          # NOTE: areArrays is FALSE here as we do not want to parse comma
+          #    separated strings (or do we?)
```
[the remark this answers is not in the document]
slodge: Agreed - cross reference with https://github.com/rstudio/plumber/blob/76cecc58925897c91cb88203ab485a0c512a0d26/R/openapi-types.R#L31 - what scenario is `converterArray` for?

R/plumber-step.R
slodge: No opinion from me - this was just where I had the args and the params conveniently together.

I think you have permissions to edit this branch and contents - so please do :+1: - I also won't be overly offended if you through everything away and do it better :+1:

R/options_plumber.R
slodge: If you are making a more major version change here... can I suggest we might also want to deprecate the typing from the dynamic routes - we don't want type information in both the param comments and in the dynamic routes?

R/plumber-step.R
```diff
@@ -307,6 +315,68 @@ PlumberEndpoint <- R6Class(
...
+          if (is.null(converter)) {
+            converter <- identity
+          } else {
+            converter <- converter[[1]]
+          }
+        }
+        new_args <- list()
+        new_args[[key]] <- converter(arg)
+        converted_args <- c(converted_args, new_args)
+      }
```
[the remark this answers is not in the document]
slodge: Tested... seems good - merging

```
══ Results ═════════════════════════════════════════════════════════════════════════════════════════════════════
Duration: 74.0 s

── Skipped tests  ──────────────────────────────────────────────────────────────────────────────────────────────
• mockery cannot be loaded (3)
• On Windows (7)
• sf cannot be loaded (3)

[ FAIL 2 | WARN 0 | SKIP 13 | PASS 2024 ]
```

R/options_plumber.R
schloerke: Yes. Let's go for it. But if `plumber.typedParameters` is not `TRUE`, then the old behavior should exist (if possible).

R/plumber-step.R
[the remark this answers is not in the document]
schloerke: (... I don't see the code in the PR)

tests/testthat/test-plumber.R
```diff
@@ -2,7 +2,10 @@ context("Plumber")
 
 exec_endpoint <- function(pr, ep_pos, subset = 1) {
   # This is a poor setup of `req` and `res`. But it works for testing purposes
-  pr$endpoints[[subset]][[ep_pos]]$exec(make_req(), PlumberResponse$new())
+  pr$endpoints[[subset]][[ep_pos]]$exec(
+    as.list(make_req()),
```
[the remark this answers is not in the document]
schloerke: I haven't seen this cause a crash.  Do you have a `traceback()`? (But seems ok for testing, maybe we should add a small comment if we keep it.)

R/options_plumber.R
slodge: I think that's what https://github.com/slodge/plumber/blob/cafc91ea24423306b19f0f71f750cb4b6ddf7aeb/R/plumber-step.R#L322 does

tests/testthat/test-plumber.R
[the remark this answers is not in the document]
slodge: Amazingly yes I do
```
>   expect_equal(exec_endpoint(r, 1), 5)
 Error in converter(arg) : 
cannot coerce type 'environment' to vector of type 'character' 
> exec_endpoint(r, 1)
 Error in converter(arg) : 
cannot coerce type 'environment' to vector of type 'character'
9.
stop(error) at plumber-step.R#102
8.
value[[3L]](cond)
7.
tryCatchOne(expr, names, parentenv, handlers[[1L]])
6.
tryCatchList(expr, classes, parentenv, handlers)
5.
tryCatch(runStep(), error = errorHandlerStep) at async.R#104
4.
runStepsUntil(initialValue, errorHandlerStep, function(x) {
FALSE
}, steps) at async.R#26
3.
runSteps(NULL, function(error) {
stop(error)
}, list(preexecStep, execStep, postexecStep)) at plumber-step.R#98
2.
pr$endpoints[[subset]][[ep_pos]]$exec(make_req(), PlumberResponse$new())
1.
exec_endpoint(r, 1)

> debugonce(exec_endpoint)
> exec_endpoint(r, 1)
debugging in: exec_endpoint(r, 1)
debug at #1: {
    pr$endpoints[[subset]][[ep_pos]]$exec(make_req(), PlumberResponse$new())
}
Browse[2]> debugonce(make_req)
Browse[2]> n
debug at #3: pr$endpoints[[subset]][[ep_pos]]$exec(make_req(), PlumberResponse$new())
Browse[2]> n
debugging in: make_req()
debug at #1: {
    req <- as.environment(list(...))
    req$REQUEST_METHOD <- toupper(verb)
    req$PATH_INFO <- path
    req$QUERY_STRING <- qs
    if (is.character(body)) {
        body <- charToRaw(body)
    }
    stopifnot(is.raw(body))
    req$rook.input <- list(read_lines = function() {
        rawToChar(body)
    }, read = function() {
        body
    }, rewind = function() {
        length(body)
    })
    req$bodyRaw <- body
    req$pr <- pr
    req
}
Browse[3]> n
debug at #2: req <- as.environment(list(...))
Browse[3]> n
debug at #3: req$REQUEST_METHOD <- toupper(verb)
Browse[3]> n
debug at #4: req$PATH_INFO <- path
Browse[3]> n
debug at #5: req$QUERY_STRING <- qs
Browse[3]> n
debug at #7: if (is.character(body)) {
    body <- charToRaw(body)
}
Browse[3]> n
debug at #8: body <- charToRaw(body)
Browse[3]> n
debug at #10: stopifnot(is.raw(body))
Browse[3]> n
debug at #11: req$rook.input <- list(read_lines = function() {
    rawToChar(body)
}, read = function() {
    body
}, rewind = function() {
    length(body)
})
Browse[3]> n
debug at #14: req$bodyRaw <- body
Browse[3]> n
debug at #15: req$pr <- pr
Browse[3]> n
debug at #16: req
Browse[3]> req
<environment: 0x000001b2a10b44f8>
Browse[3]> n
exiting from: make_req()
Error in converter(arg) : 
  cannot coerce type 'environment' to vector of type 'character'
```

R/plumber-step.R
schloerke: Notes for me....

This method is only being applied to the handle's non-`req`/`res` arguments. It will not apply to `pr$argsPath`, `pr$argsQuery` or any other stored values.  Just the functions helper parameters that have typically been discouraged.

Since we'll have the location of `query` and `path` in the parameter definitions, we should process the parameters as soon as possible (such as in the [query filter](https://github.com/rstudio/plumber/blob/76cecc58925897c91cb88203ab485a0c512a0d26/R/parse-query.R#L6) and [path param extraction](https://github.com/rstudio/plumber/blob/76cecc58925897c91cb88203ab485a0c512a0d26/R/plumber.R#L689)).

This method should just be a helper method that is applied to the args given.

... however... the query filter does not know which endpoint is going to be executed. This makes typing the arguments impossible (or not something we should do). 

-----------------

This still feels weird to me, but we're being caught by the query parameters being stored on `pr$args` too early.

I'm worried about having a path parameter being parsed as an integer and the user adding a matching parameter query string value that is not an integer. Ex: `/iteration/5?i=A` where the route is defined as `@param i:int iteration`. This would set `req$args` to `list(i = "A", i = 5)` and `req$args[["i"]]` would resolve to `"A"`. 

This is just a problem in general about which param is returned. So maybe it's no worse that what it currently is.

--------------------

Therefore, this **is** the correct location to parse the args and only apply them to the helper params (at the cost of possibly overlapping which parameter is actually retrieved).

schloerke:
I'll need to review this with Joe. It'll be a bit before it is merged. Hopefully sooner

R/plumber-step.R
slodge: One possibility here is to deprecate putting the type in the routes - suggest all type information is placed in the @param

Obviously that's not pain free for existing code, but should be possible to keep support and to just add a warning that this will get switched off at some point in the future.

But this is probably easier for me to say as I don't have lots of existing plumber APIs to maintain and update.

R/plumber-step.R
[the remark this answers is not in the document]
slodge: I think I'd like to get rid of this - I think it can only be for supporting arrays in dynamic routes - since queryparams do arrays using repeated key value pairs and JSON bodies do arrays using Jason (not sure about forms though?)

slodge:
I suspect I might have overdone the PRs here... Sorry for the overflow - just filling some spare cycles before Xmas!



Let me know if there's anything else you want/need on #899, #891, #892 - apart from that I'll keep out the way and wait for Posit to work out what direction they want plumber to head in :+1:

tests/testthat/test-shared-secret.R
```diff
@@ -11,7 +11,7 @@ test_that("requests with shared secrets pass, w/o fail", {
   res <- PlumberResponse$new()
   output <- pr$route(req, res)
   expect_equal(res$status, 400)
-  expect_equal(output, list(error = "400 - Bad request"))
+  expect_equal(output, list(error = "400 - Bad request", message = "Shared secret mismatch"))
```
schloerke, line 14: This should be the original code. I'd rather have it error and fixed elsewhere

```suggestion
  expect_equal(output, list(error = "400 - Bad request")
```

tests/testthat/test-plumber.R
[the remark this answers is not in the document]
schloerke: Something is happening in the converter code. I'll handle this from here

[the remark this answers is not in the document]
schloerke: Going to revert the code change so the error pops up. Going to resolve this issue as well

tests/testthat/test-plumber.R
```diff
@@ -2,7 +2,10 @@ context("Plumber")
...
+    PlumberResponse$new()
+  )
```
schloerke, line 8: Reverting to surface error

```suggestion
  pr$endpoints[[subset]][[ep_pos]]$exec(make_req(), PlumberResponse$new())
```

tests/testthat/test-shared-secret.R
```diff
@@ -11,7 +11,7 @@ test_that("requests with shared secrets pass, w/o fail", {
   res <- PlumberResponse$new()
   output <- pr$route(req, res)
   expect_equal(res$status, 400)
-  expect_equal(output, list(error = "400 - Bad request"))
+  expect_equal(output, list(error = "400 - Bad request")
```
schloerke, line 14: Fix my typo

```suggestion
  expect_equal(output, list(error = "400 - Bad request"))
```

R/plumber-step.R
```diff
@@ -307,6 +315,57 @@ PlumberEndpoint <- R6Class(
...
+        return(args)
+      }
+
+      converted_args <-
+        Map(
+          key = args_names,
+          arg = args,
+          param = params[args_names],
+          f = function(key, arg, param) {
+            # not sure why... but we allow unnamed arguments through
```
schloerke, line 355:
```suggestion
            # Early return on unnamed args or `NULL` values
```

R/plumber-step.R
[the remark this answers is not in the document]
schloerke: I had a funny state / commit-view while reviewing. 🙄 Found it 🤦‍♂️

R/options_plumber.R
jcheng5: Discussed with @schloerke in realtime, we think it makes sense to default to `FALSE` for now and then introduce the concept of editions, like testthat (https://testthat.r-lib.org/articles/third-edition.html).

@slodge, how do you feel about that?

R/plumber-step.R
jcheng5: @slodge I like the idea of deprecating the type in the routes, we could tie this new behavior to an edition.

R/options_plumber.R
slodge: Happy for Posit/RStudio to choose how you version plumber and its features.

---------

I've not read about editions before, and I don't know if I understand right now how they'd work for a plumber app or for a helper package (e.g. how would the package specify that it relies on a plumber edition as well as its version?)

In this particular case, I suspect that you could get away with setting this to `TRUE` and incrementing the major plumber version (I also don't think this would actually break that many apps... but I've been wrong about that before!). 

But this is **100% the call for whoever maintains plumber** - aka **you** 👍

slodge:
Just checking back in... Does Posit want these changes? Is now just too busy with other work? Should I go away for another couple of weeks? Or months? Does the team need longer to choose/plan? Are you expecting more from me at this time? (Apologies if I've missed anything!)

We're currently contemplating merging our own internal developments into a single branch of our own repo, and possibly switching to use a plumber version hosted in our internal CRAN so that we can take advantage of some of these features. I'd much prefer to edit/merge this code into "plumber proper", but I appreciate that making and maintaining a product for the wider community is harder and requires more thought and care than maintaining an internal package... :+1:

slodge:
Closing - after 3 months, this branch/PR too hard to maintain.

Issue #890, opened 2022-11-29
Add `Date` types to openapi (and conversions)


slodge:
Much of the work I do involves time series

For these, date, and to a lesser extent, datetime parameters are very common.

It would be very useful to be able to supply `date` hints in the api definitions - e.g.

```
#* Get the stock prices
#* @param ticker:string The ticker
#* @param from:date The first date to include (inclusive)
#* @param to:date The last date to include (inclusive)
#* @get /prices
function(ticker, from, to) {
   # code here...
}
```

From https://swagger.io/docs/specification/data-models/data-types/#string this can be done using:

> ### String Formats
> An optional format modifier serves as a hint at the contents and format of the string. OpenAPI defines the following built-in string formats:
>  - date – full-date notation as defined by [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6), for example, 2017-07-21
>  - date-time – the date-time notation as defined by [RFC 3339, section 5.6](https://tools.ietf.org/html/rfc3339#section-5.6), for example, 2017-07-21T17:32:28Z
>  - password – a hint to UIs to mask the input
>  - byte – base64-encoded characters, for example, U3dhZ2dlciByb2Nrcw==
>  - binary – binary data, used to describe files (see [Files](https://swagger.io/docs/specification/data-models/data-types/#file) below)

slodge:
There's easy workarounds to this - date parameters can be cast from incoming strings inside the API functions.

Leaving this open for maintainers to decide if/when to implement. Happy for them to close it if required.

PR #891, opened 2022-11-29
Draft changes to support Dates and DateTime

slodge:
This is a PR for #890 - providing Date and DateTime support

## PR task list:

Functional:
- [x] Discuss design and make sure the maintainer is happy...
- [x] Discuss implementation and make sure the maintainer is happy...
- [x] I suspect the regexes "could be better"...

Admin:

- [x] Sign RStudio CLA 
- [ ] Update NEWS
- [ ] Update articles
- [ ] Add tests
  - [ ] GET and POST query parameters
  - [ ] body parameters
  - [ ] dynamic route parameters
  - [ ] parameters which don't parse
- [ ] Run `devtools::check()`

```r

#* Consider this...
#* @param when:date the date
#* @param t:date-time The date and time
#* @get /dates
function(when, t) {

  browser()
  paste0("The date was ", when, " and the time was ", t)
}
```

That seems to work:
image

Note: have also tested this alongside #889

### DIFF
--- a/R/openapi-types.R
+++ b/R/openapi-types.R
@@ -7,122 +7,172 @@ defaultApiType <- structure("string", default = TRUE)
 defaultIsArray <- structure(FALSE, default = TRUE)
 
 add_api_info_onLoad <- function() {
-  addApiInfo <- function(apiType, plumberTypes,
-                         regex = NULL, converter = NULL,
-                         format = NULL,
-                         location = NULL,
-                         realType = NULL) {
-    apiTypesInfo[[apiType]] <<-
-      list(
-        regex = regex,
-        converter = converter,
-        format = format,
-        location = location,
-        realType = apiType %||% realType,
-        # Q: Do we need to safe guard against special characters, such as `,`?
-        # https://github.com/rstudio/plumber/pull/532#discussion_r439584727
-        # A: https://swagger.io/docs/specification/serialization/
-        # > Additionally, the allowReserved keyword specifies whether the reserved
-        # > characters :/?#[]@!$&'()*+,;= in parameter values are allowed to be sent as they are,
-        # > or should be percent-encoded. By default, allowReserved is false, and reserved characters
-        # > are percent-encoded. For example, / is encoded as %2F (or %2f), so that the parameter
-        # > value quotes/h2g2.txt will be sent as quotes%2Fh2g2.txt
-        regexArray = paste0("(?:(?:", regex, "),?)+"),
-        converterArray = function(x) {converter(stri_split_fixed(x, ",")[[1]])}
-      )
+  addApiInfo <- function(
+    keys,
+    location = c("body", "route", "query"),
+    openApiType = c("string", "number", "integer", "boolean", "object"),
+    # note that openApiFormat is extensible - so match.arg should not be used on openApiFormat
+    openApiFormat = c("float", "double", "int32", "int64", "date", "date-time", "password", "byte", "binary"),
+    openApiRegex  = NULL,
+    parser = function(input) { input; }) {
 
-    for (plumberType in plumberTypes) {
-      plumberToApiTypeMap[[plumberType]] <<- apiType
+    # TODO  - some match.arg stuff?
+
+    entry <- list(
+      location = location,
+      openApiType = openApiType,
+      openApiFormat = openApiFormat,
+      # Q: Do we need to safe guard against special characters, such as `,`?
+      # https://github.com/rstudio/plumber/pull/532#discussion_r439584727
+      # A: https://swagger.io/docs/specification/serialization/
+      # > Additionally, the allowReserved keyword specifies whether the reserved
+      # > characters :/?#[]@!$&'()*+,;= in parameter values are allowed to be sent as they are,
+      # > or should be percent-encoded. By default, allowReserved is false, and reserved characters
+      # > are percent-encoded. For example, / is encoded as %2F (or %2f), so that the parameter
+      # > value quotes/h2g2.txt will be sent as quotes%2Fh2g2.txt
+      openApiRegex = openApiRegex,
+      openApiRegexArray = paste0("(?:(?:", openApiRegex, "),?)+"),
+      parser = parser,
+      # TODO - this won't work for strings that contain commas?
+      parserArray = function(x) {parser(stri_split_fixed(x, ",")[[1]])}
+    )
+
+    for (apiType in keys) {
+      apiTypesInfo[[apiType]] <<- entry
     }
-    # make sure it could be called again
-    plumberToApiTypeMap[[apiType]] <<- apiType
 
     invisible(TRUE)
   }
 
+      # list(
+      #   regex = regex,
+      #   converter = converter,
+      #   format = format,
+      #   location = location,
+      #   realType = apiType %||% realType,
+      #   # Q: Do we need to safe guard against special characters, such as `,`?
+      #   # https://github.com/rstudio/plumber/pull/532#discussion_r439584727
+      #   # A: https://swagger.io/docs/specification/serialization/
+      #   # > Additionally, the allowReserved keyword specifies whether the reserved
+      #   # > characters :/?#[]@!$&'()*+,;= in parameter values are allowed to be sent as they are,
+      #   # > or should be percent-encoded. By default, allowReserved is false, and reserved characters
+      #   # > are percent-encoded. For example, / is encoded as %2F (or %2f), so that the parameter
+      #   # > value quotes/h2g2.txt will be sent as quotes%2Fh2g2.txt
+      #   regexArray = paste0("(?:(?:", regex, "),?)+"),
+      #   converterArray = function(x) {converter(stri_split_fixed(x, ",")[[1]])}
+      # )
+
+
   addApiInfo(
-    "boolean",
-    c("bool", "boolean", "logical"),
-    "[01tfTF]|true|false|TRUE|FALSE",
-    as.logical,
+    c("boolean", "bool", "logical"),
+    openApiType = "boolean",
+    openApiRegex = "[01tfTF]|true|false|TRUE|FALSE",
+    parser = as.logical,
     location = c("query", "path")
   )
+
   addApiInfo(
-    "number",
-    c("dbl", "double", "float", "number", "numeric"),
-    "-?\\\\d*\\\\.?\\\\d+",
-    as.numeric,
-    format = "double",
+    c("number", "numeric"),
+    openApiType = "number",
+    openApiRegex = "-?\\\\d*\\\\.?\\\\d+",
+    parser = as.numeric,
     location = c("query", "path")
   )
+
   addApiInfo(
-    "integer",
-    c("int", "integer"),
-    "-?\\\\d+",
-    as.integer,
-    format = "int64",
+    c("dbl", "double"),
+    openApiType = "number",
+    openApiFormat = "double",
+    openApiRegex = "-?\\\\d*\\\\.?\\\\d+",
+    parser = as.numeric,
     location = c("query", "path")
   )
+
   addApiInfo(
-    "string",
-    c("chr", "str", "character", "string"),
-    "[^/]+",
-    as.character,
+    c("float"),
+    openApiType = "number",
+    openApiFormat = "float",
+    openApiRegex = "-?\\\\d*\\\\.?\\\\d+",
+    parser = as.numeric,
     location = c("query", "path")
   )
+
   addApiInfo(
-    "date-time",
-    c("POSIXct", "POSIXt"),
-    "[^/]+",
-    lubridate::as_datetime,
-    format = "date-time",
-    location = c("query", "path"),
-    realType = "string"
+    c("integer", "int"),
+    openApiType = "integer",
+    openApiFormat = "int64",
+    openApiRegex = "-?\\\\d+",
+    parser = as.integer,
+    location = c("query", "path")
   )
+
   addApiInfo(
-    "date",
-    c("Date"),
-    "[^/]+",
-    lubridate::as_date,
-    format = "date",
-    location = c("query", "path"),
-    realType = "string"
+    c("string", "str"),
+    openApiType = "string",
+    parser = as.character,
+    location = c("query", "path")
   )
+
   addApiInfo(
-    "object",
-    c("list", "data.frame", "df", "object"),
-    location = "requestBody"
+    c("date-time", "datetime"),
+    openApiType = "string",
+    # https://regex101.com/r/qH0sU7/1
+    openApiRegex = "^((?:(\\d{4}-\\d{2}-\\d{2})T(\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?))(Z|[\\+-]\\d{2}:\\d{2})?)$",
+    openApiFormat = "date-time",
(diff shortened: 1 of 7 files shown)


R/openapi-types.R
```diff
@@ -18,7 +18,7 @@ add_api_info_onLoad <- function() {
         converter = converter,
         format = format,
         location = location,
-        realType = realType,
+        realType = ifelse(is.null(realType), apiType, realType),
```
schloerke, line 21:
```suggestion
        realType = apiType %||% realType,
```

R/openapi-types.R
```diff
@@ -70,6 +70,24 @@ add_api_info_onLoad <- function() {
     as.character,
     location = c("query", "path")
   )
+  addApiInfo(
+    "date-time",
+    c("POSIXct", "POSIXt"),
+    "[^/]+",
+    lubridate::as_datetime,
```
schloerke, line 77: Is there a `{base}` method we can use? `{lubridate}` is currently not a direct dependency.

--------------------

If keeping `{lubridate}`... we'll need to wrap the method as to not cache the `{lubridate}` function at `{plumber}`'s installation time. 

Ex:
```r
function(x) { lubridate::as_datetime(x) },
```

R/openapi-types.R
```diff
@@ -70,6 +70,24 @@ add_api_info_onLoad <- function() {
     as.character,
     location = c("query", "path")
   )
+  addApiInfo(
+    "date-time",
```
schloerke, line 74: Let's keep the name similar to the `lubridate` method.

```suggestion
    "datetime",
```

R/openapi-types.R
```diff
@@ -70,6 +70,24 @@ add_api_info_onLoad <- function() {
...
+    c("POSIXct", "POSIXt"),
+    "[^/]+",
+    lubridate::as_datetime,
+    format = "date-time",
```
schloerke, line 78: (Continuation)
```suggestion
    format = "datetime",
```

schloerke (changes requested):
Looks good!

Requests: 
* Can you add the `date`/`datetime` formats to `annotations.Rmd`'s table in ``More details on `@param` ``?
* Add a NEWS entry ending in `(@slodge #891)`

schloerke:
For testing, please add tests where you see fit in [tests/testthat/test-zzz-openapi.R](https://github.com/rstudio/plumber/blob/76cecc58925897c91cb88203ab485a0c512a0d26/tests/testthat/test-zzz-openapi.R)

R/openapi-types.R
```diff
@@ -70,6 +70,24 @@ add_api_info_onLoad <- function() {
     as.character,
     location = c("query", "path")
   )
+  addApiInfo(
+    "date-time",
+    c("POSIXct", "POSIXt"),
```
schloerke, line 75: This would allow for people to use 
```r
#* @param x:POSIXct Description of my date `x`
```

Since we are not doing anything in particular about the date type object, let's rearrange the values.

```suggestion
    c("datetime", "date-time"), # Keep `datetime` for convenience
```

R/openapi-types.R
schloerke: I did not see the openapi definition of `date-time` in the issue. 

Resolving to remove the suggestion.

R/openapi-types.R
```diff
@@ -70,6 +70,24 @@ add_api_info_onLoad <- function() {
...
+    "[^/]+",
+    lubridate::as_datetime,
+    format = "date-time",
+    location = c("query", "path"),
+    realType = "string"
+  )
+  addApiInfo(
+    "date",
+    c("Date"),
```
schloerke, line 84: (Continuation)
```suggestion
    c("date"),
```

R/openapi-types.R
```diff
@@ -70,6 +70,24 @@ add_api_info_onLoad <- function() {
...
+    "[^/]+",
+    lubridate::as_date,
+    format = "date",
```
schloerke, line 87: Is there a {base} method we can use? `{lubridate}` is currently not a direct dependency.

R/openapi-types.R
```diff
@@ -70,6 +70,24 @@ add_api_info_onLoad <- function() {
     as.character,
     location = c("query", "path")
   )
+  addApiInfo(
+    "date-time",
+    c("POSIXct", "POSIXt"),
+    "[^/]+",
```
schloerke, line 76: Could test for full format. From https://regex101.com/r/qH0sU7/1 :

(untested)
```suggestion
    # https://regex101.com/r/qH0sU7/1
    "^((?:(\\d{4}-\\d{2}-\\d{2})T(\\d{2}:\\d{2}:\\d{2}(?:\\.\\d+)?))(Z|[\\+-]\\d{2}:\\d{2})?)$",
```

R/openapi-types.R
```diff
@@ -70,6 +70,24 @@ add_api_info_onLoad <- function() {
...
+    lubridate::as_datetime,
+    format = "date-time",
+    location = c("query", "path"),
+    realType = "string"
+  )
+  addApiInfo(
+    "date",
+    c("Date"),
+    "[^/]+",
```
schloerke, line 85: (continuation)

(untested)
```suggestion
    # https://regex101.com/r/qH0sU7/1
    "^\\d{4}-\\d{2}-\\d{2}$",
```

slodge:
@schloerke apologies for this... but I think it might be sensible to merge this one **after** we work out what to do with #889 

I don't think this has much use without the work in #889 ... and I think after committing #889 then there'll be other changes needed here (e.g. other tests for the converter)

Sorry - they're a bit inter-related...

R/openapi-types.R
slodge: I don't know.. 

I trust `lubridate` more than base. I feel like I have had issues where base methods haven't done as well with timezone issues... but I don't have specific examples, and I fully admit that I'm not a long-term R coder, and I know way more tidyverse than base R.

R/openapi-types.R
slodge: I've got to admit I don't really follow what the current parameters for addApiInfo mean... and I think #889 might have changed what they mean...

As far as I can see, if we are heading towards auto-conversion by default, then in an apiInfo there should be:

- the type key for the entry - this will be what appears next to the `@param` name - e.g. `@param my_field:int`
- the locations this can be found (I guess... although I've not used this much)
- the openapi type that will be displayed (one of https://swagger.io/docs/specification/data-models/data-types/ ?)
- (optional) the `format` specifier to go alongside the openapi type
- (optional) the R `parser` that will be applied to parse the incoming `character` value (this could possibly also return errors if parsing unsuccessful?)
- (optional) the regex that is applied in the swagger ui
- (optional) possible other keywords like `minimum`, `multipleOf`, 'minLength`, etc (but think they live in a future PR 👍 )

 Based on my understanding from this list... I don't really understand what `apiType`, `plumberTypes` and `realType` are
 - if doing a major release, can we get rid of (or deprecate) some of these fields?
 
Would it help to clean up `addApiInfo` and replace it with a function with a surface something like:
`
addApiInfo(
    key, 
    locations = c("body", "route", "query"), 
    openApiType = c("string", "number", "integer", "boolean", "object"),
    # note that openApiFormat is extensible - so match.arg should not be used on openApiFormat
    openApiFormat = c("float", "double", "int32", "int64", "date", "date-time", "password", "byte", "binary"), 
    openApiRegex  = NULL,
    parser = function(input) { list(error=NULL, result=input) } 
 ) {
 ...
 }
 `
 
That could then be called and configured with only one key for each entry (I think maybe plumber doesn't need this `for` loop any more - https://github.com/rstudio/plumber/blob/befc866cb0d9d24cedf8399c82305c2e2e306214/R/openapi-types.R#L34 ) 

Disclaimer: this is all coming very much from #889 and I admit I don't have any experience of the attachment/file and object inputs... they might cause more problems here?

Sorry... didn't mean to ask such "big questions"

R/openapi-types.R
slodge: I think this is my confusion here about what these parameters to `addApiInfo` mean... 

- are they openapi types? - in which case I guess `date` works
- or are they the R types? - in which case I think `Date` is better (`class(lubridate::today())`)

Currently the options I can see include lots of things which are openapi, lots which are R and some which are neither?

e.g. I can see all of:

```
    c("bool", "boolean", "logical"),
    c("dbl", "double", "float", "number", "numeric"),
    c("chr", "str", "character", "string"),
    c("POSIXct", "POSIXt"),
    c("list", "data.frame", "df", "object"),
```

I can see why we'd want to have more keys than are in the openapi spec - and I guess I'm ok with duplicates (e.g. dbl and double), but if there is a chance for a breaking change, then it might be nice to clean these up (especially **before** I add the very ugly POSIXct and POSIXt to the list 👍 )

R/openapi-types.R
slodge: I'll need to read the differences - https://lubridate.tidyverse.org/reference/as_date.html#:~:text=Called%20on%20a%20POSIXct%20object,%22UTC%22%20by%20default). - I think as.Date can work... but there are timezone gotchas (which are worse in base R I think)

R/openapi-types.R
slodge: I had a go at something like this on a different branch - https://github.com/slodge/plumber/commit/536dd297cc5a8ebef4bb6558e1b044091b0f2d7c

To keep backward portability, I kept all the synonyms - e.g. dbl and double

That change is not complete... several more bits to do including the file/object stuff - but wanted to send it in for discussion... Is this direction worth pursuing or is it too big a change for the core team? 

It's not something I feel we desparately need - it is just a cleanup of api spec definitions rather than anything functional...

schloerke: Much better variable names! Makes it much clearer as to what is happening. I'll admit it was fussy when I wrote it.

Nice work on the backwards compatibility. 

Would be happy to merge the changes once complete.

R/openapi-types.R
schloerke: I agree it is 100% better using lubridate. If we need it, we can use `function(d) { lubridate::as_date(d) }` and have lubridate as a Suggests in the DESCRIPTION file.

R/openapi-types.R
slodge: I've implemented a simpler regex and added some unit tests.

I tried your regex but got lost in trying to resolve all the optional components within a single group.

I felt sick when visiting https://stackoverflow.com/a/3143231/373321 and there were worse out there... I think what I have added is OK (you can review it...)

While here... I also wondered why `////d` is necessary here - not sure how we managed it, but we've made regex even more unreadable :)

R/openapi-types.R
slodge: Done - with some protection/fallback if lubridate not available

R/openapi-types.R
slodge: Resolving this as changes are included now

R/openapi-types.R
slodge: Thanks.

Have got all tests passing (first half of Argentina vs Australia wasn't exciting so I had spare time!)

Have merged those changes into this PR.

Would really like to drop `converterArray` depending on decisions here and in the other PR - https://github.com/rstudio/plumber/pull/889#discussion_r1038748298 - I guess we really only intend this ever to be used for `[int]` and in query paths...

I can merge the PRs into one big thing to review if that would make it easier?

R/openapi-types.R
slodge: Resolved - lubridate in as Suggests

R/openapi-types.R
slodge: Still open here:

- Add the date/datetime formats to annotations.Rmd's table in More details on `@param`  (and more to discuss?)
- Add a NEWS entry ending in (@slodge #891)
- Discuss/resolve this `converterArray` stuff

R/openapi-types.R
slodge: Am resolving this - but expecting it to be reviewed still

R/openapi-types.R
slodge: That's pretty much what I added - but without the ^ and $ and with those four slash d's -  `\\\\d`

Test added too - so resolving

R/openapi-types.R
```diff
@@ -2,114 +2,197 @@
...
     location = c("query", "path")
   )
+
+  # be careful not to trap lubridate on package installation
+  lubridate_available <- function() {
+    system.file(package = "lubridate") != "";
+  }
+
+  parse_datetime <- function(x) {
+    if (lubridate_available()) lubridate::as_datetime(x) else as.Date.POSIXct(x)
```
jcheng5, line 125: I don't love the idea of the behavior being subtly different depending on whether this package is installed (we do this sometimes in Shiny and it always seems to end up hurting). Is there a semantic difference between the two?

If there is, and we strongly prefer lubridate's behavior, I'd prefer to throw an error if lubridate is not available.

R/openapi-types.R
```diff
@@ -2,114 +2,197 @@
...
-    "integer",
-    c("int", "integer"),
-    "-?\\\\d+",
-    as.integer,
-    format = "int64",
+    c("date-time", "datetime"),
+    openApiType = "string",
+    # much more complex regexes are available... e.g. see https://stackoverflow.com/a/3143231/373321
+    # (if importing one of those, remember to use cc attribution)
+    openApiRegex = "\\\\d{4}-\\\\d{2}-\\\\d{2}T\\\\d{2}:\\\\d{2}:\\\\d{2}Z",
```
jcheng5, line 137: Is this intentionally limiting the timezone to `Z`?

slodge: I'm very happy to see it restricted... I encourage everyone to use UTC whenever they can (and if I can persuade the UK government to drop BST then it'd make my life so much easier...)

But from a [RFC 3339](https://www.rfc-editor.org/rfc/rfc3339#section-5.6) perspective, it could probably be extended.

As far as I understand it, this regex is only used in parsing path query variables (but I might have that wrong...). With that in mind, are these regexes needed at all?

If they are, then the comment points to [stackoverflow.com/a/3143231/373321](https://stackoverflow.com/a/3143231/373321) with a wealth of options :)

R/openapi-types.R
slodge: From 

> ## Compare to base R
>These are drop in replacements for [as.Date()](https://rdrr.io/r/base/as.Date.html) and [as.POSIXct()](https://rdrr.io/r/base/as.POSIXlt.html), with a few tweaks to make them work more intuitively.

>
> Called on a POSIXct object, as_date() uses the tzone attribute of the object to return the same date as indicated by the printed representation of the object. This differs from as.Date, which ignores the attribute and uses only the tz argument to [as.Date()](https://rdrr.io/r/base/as.Date.html) ("UTC" by default).

>
> Both functions provide a default origin argument for numeric vectors.

>
> Both functions will generate NAs for invalid date format. Valid formats are those described by ISO8601 standard. A warning message will provide a count of the elements that were not converted.

>
> as_datetime() defaults to using UTC.
 
That last point wins me over ... but should be possible to do in as.Date.POSIXct too?

I'd be very happy adding a permanent dependency to `lubridate` (would any plumber users really mind?), or switching to an error, or duplicating the lubridate code here (possibly with some cross-referencing to what open api spec says)

slodge:
Closing - too hard to maintain separate long-running PRs

Issue #819, opened 2021-08-11
Schema Definitions


kempspo:
Definitely useful to be able to define the schema of our objects, [like so](https://swagger.io/docs/specification/components/). Found it particularly hard to communicate POST body request formats without having to talk to other devs directly and figure out how it was built. 

```
components:
  schemas:
    User:
      type: object
      properties:
        id:
          type: integer
        name:
          type: string
```

meztez:
yes it would be great.

@kempspo What kind of abstraction would make sense to you?

kempspo:
To be honest, I don't have a specific opinion on how it should go, but I would think maybe some function like `pr_add_schema` that takes a Plumber API object, a name and a list that holds the schema object definition (I honestly don't know how to make this simpler for people to use). 

I think a way to refer to the schema from the `pr_handle` set of functions would also be helpful? 
```
requestBody:
        content:
          application/json:
            schema:
              $ref: '#/components/schemas/Pet'
```

Issue #763, opened 2021-02-05
Add ability to register a plumb block method


schloerke:
https://github.com/rstudio/plumber/blob/5d2070089849e05bb356537de266c2464b544a5f/R/plumb-block.R#L16-L243 

☝️ has a lot of repetitive behaviors and is closed off to other developers. 

It would be great if plumber (and others) could register the functions (similar to parsers and serializers) and allow them to parse a block independently. 

Example: 
```r
#* @mycustomplumbertag key value value2
```

Thoughts:
* Only a single line should be attempt at one time. (This is current behavior)
* Once a block parser is a _hit_, the remaining block parsers should not attempted.
* Don't know what should be supplied or be returned from the block parsers. There is a lot of global state that is currently being altered in the function above. This _state_ should be continued throughout the block's parsing.
* Should we be allowed to overwrite? Maybe a verbose argument? (like the parsers and serializers)
* "block parser" or "plumb block" is a confusing name. Need a better one.

Issue #222, opened 2017-12-21
Annotations for return type(s)?


JHibbard:
Is it possible to annotate endpoints with their [response type(s)/structure](https://swagger.io/docs/specification/2-0/describing-responses/)?  The [swagger codegen](https://twitter.com/wing328/status/904520974954926080) project released (Sep 2017) code to generate R clients from swagger specs but requires these endpoint specs to deserialize responses; this along with plumber would allow a top to bottom RESTful api solution for R!

trestletech:
You can document response codes (see https://github.com/trestletech/plumber/blob/master/inst/examples/11-car-inventory/plumber.R#L46), but currently not the response structure.

I haven't figured out a great way to enable the documentation of the response structure. I'm thinking there may be something built around R6 that would allow us to document the structure of each object that we're going to serialize in/out, but a.) jsonlite doesn't play nicely with R6 custom serialization at this point AFAIK, and b.) I'm not convinced that this is the best path forward.

How would you want to document the structure of your objects?

JHibbard:
Ah, that sounds tricky.

I'd like to define a named json schema that I can reuse; e.g. I might have a "cookie" definition that maps to a (swagger/open api) json schema with {'type': <string>, 'cost': <number>}.  This would allow me to have an endpoint that returns a single "cookie" and another that returns an array of "cookie" objects.  swagger/OpenAPI client sdk generators could then deserialize these JSON objects into language-specific types; e.g. "cookie" might be a list with named members in R and a dictionary in Python.  Without this extra type annotation client sdk's can't be generated.

A possible "quick fix" would be to allow a filter to be applied after an endpoint's function, on the response.  Then you could inject this information into the swagger.json before it gets returned to callers.

trestletech:
Yeah, I'd be tempted to just allow extensions/modifications to the Swagger generation at that point. 

My gut is that there's an elegant solution hiding in here somewhere using R6. It would be a little extra work on the API author to have to compose their APIs using these R6 objects, but I feel like the type safety you'd get in exchange would be worth it. 

Unfortunately, that might require some extensions to jsonlite to get that to work, as -- last I checked -- it wasn't supported.

svdwoude:
I'd be very interested in exploring this further and implementing this.

I'm not quite sure I completely understand the approach you have in mind with R6.
Could you provide a minimal example of the annotation you'd envision for the response schema? Or would you generate that response schema entirely based upon an 'example' output from the endpoint-function?

[...]

PR #892, opened 2022-12-04
Add schema definitions for request and response bodies

slodge:
This is a PR for #222 - providing schema definitions for response and request bodies

(Think also covers #819)

## PR task list:

Functional:
- [ ] Discuss design and make sure the maintainer is happy... 
- [ ] Discuss implementation and make sure the maintainer is happy... I think some names could be better (starting with `object`). I think there should be a way to do the object blocks without the empty function() {} calls? I think #763 might help make the block code less icky... You could look at more advanced scenarios too...
- [ ] I suspect the regexes "could be better"...
- [ ] Need to complete the programmatic `pr_` access

Admin:

- [x] Sign RStudio CLA 
- [ ] Deal with the other PRs first... there is going to be some merge hell here (sorry!)
- [ ] Update NEWS
- [ ] Update articles/doc
- [ ] Add tests
  - [ ] lots to do?
- [ ] Needs testing with some client swagger code gen in e.g. C#  
- [ ] Run `devtools::check()`

```r

devtools::load_all()


#* Do something
#* @post /doIt
#* @requestBody ARequestThing
#* @response 200 @body AResponseThing The success response
function(req) {
  browser()
  req$body
}

#* This is an object
#* @object ARequestThing
#* @prop first:string The filter map
#* @prop a:[logical] The filter map
#* @prop b:[string] The set of string to filter
function() {}

#* This is an object
#* @object AResponseThing
#* @prop life:string Just a description
#* @prop details:[object]:AResponseSubThing This gets advanced
function() {}

#* This is an object
#* @object AResponseSubThing
#* @prop id:int Just a description
#* @prop value:string More descriptions
function() {}

```

This generates openapi json like:
```
{
  "servers": [
    {
      "url": ""
    }
  ],
  "openapi": "3.0.3",
  "info": {
    "description": "Plumber example description.",
    "title": "Plumber Example API",
    "version": "1.0.0"
  },
  "paths": {
    "/doIt": {
      "post": {
        "summary": "Do something",
        "description": "",
        "responses": {
          "200": {
            "description": "The success response",
            "content": {
              "application/json": {
                "schema": {
                  "$ref": "#/components/schemas/AResponseThing"
                }
              }
            }
          },
          "500": {
            "description": "Internal Server Error",
            "content": {
              "application/json": {
                "schema": {
                  "type": "string"
                }
              }
            }
          },
          "default": {
            "description": "Default response."
          }
        },
        "parameters": [],
        "requestBody": {
          "content": {
            "application/json": {
              "schema": {
                "$ref": "#/components/schemas/ARequestThing"
              }
            }
          }
        }
      }
    }
  },
  "components": {
    "schemas": {
      "ARequestThing": {
        "type": "object",
        "properties": {
          "first": {
            "type": "string",
            "description": "The filter map",
            "required": false
          },
          "a": {
            "type": "array",
            "description": "The filter map",
            "required": false,
            "items": {
              "type": "boolean"
            }
          },
          "b": {
            "type": "array",
            "description": "The set of string to filter",
            "required": false,
            "items": {
              "type": "string"
            }
          }
        }
      },
      "AResponseThing": {
        "type": "object",
        "properties": {
          "life": {
            "type": "string",
            "description": "Just a description",
            "required": false
          },
          "details": {
            "description": "This gets advanced",
            "required": false,
            "items": {
              "$ref": "#/components/schemas/AResponseSubThing"
            },
            "type": "array"
          }
        }
      },
      "AResponseSubThing": {
        "type": "object",
        "properties": {
          "id": {
            "type": "integer",
            "format": "int64",
            "description": "Just a description",
            "required": false
          },
          "value": {
            "type": "string",
            "description": "More descriptions",
            "required": false
          }
        }
      }
    }
  }
}
```

That seems to work:

image

image


jcheng5:
1. I was a little surprised that the schema is not used to actually validate the input at runtime, especially on the request side. However, I also don't know of an easy way to validate json against a schema in R, so this might be a bigger lift than we ever want to do. If we do think this is something we might do in the future and would want to make it the default behavior at that point, it'd be good to pick a syntax today that is forward-compatible with a world where that is the case; for example, `@requestBody ARequestThing validate=false` (or something like that) being required today, and once that feature exists then `validate=true` becomes supported and `@requestBody ARequestThing` implies `validate=true`.

2. Instead of `@object`, how about `@schema`?

slodge:
> 1. I was a little surprised that the schema is not used to actually validate the input at runtime, especially on the request side. However, I also don't know of an easy way to validate json against a schema in R, so this might be a bigger lift than we ever want to do. If we do think this is something we might do in the future and would want to make it the default behavior at that point, it'd be good to pick a syntax today that is forward-compatible with a world where that is the case; for example, `@requestBody ARequestThing validate=false` (or something like that) being required today, and once that feature exists then `validate=true` becomes supported and `@requestBody ARequestThing` implies `validate=true`.

I like the idea of validation... 

...but I don't love the idea of adding `validate=false` to the code when there's no `validate=true` there yet (or realistically likely to be added in the next few years?)

Assuming plumber is moving to editions... then the `validate=` could be added in a future edition so could be left out for now?

**Aside:** Internally I've been doing some "hacking" on top of [rapiclient](https://github.com/slodge/rapiclient) to allow R clients to POST RequestBody calls - I've no plans on adding validation there either, but I'm really trying to tease it into creating example `list`s to help people write calls.

**Overall:** the functionality in this PR is definitely "version 1" - I think it's a good start, but doesn't cover all the complexity available in the openapi spec - so protecting future plumber here is v. sensible 👍 

> 2. Instead of `@object`, how about `@schema`?

I'd be worried that both `object` and `schema` have meanings in [OpenApi3 Spec](https://swagger.io/docs/specification/data-models/data-types/#object) so wonder if there's a third option ... if not, I'm very happy to avoid the word `object` - `schema` definitely feels better 👍

slodge:
Change made to @schema from @object

Now closing this PR - too difficult to maintain within our own source

Issue #837, opened 2021-10-18
`png` serializer with dynamic image size


logstar:
> If you wish to dynamically size images, you will need render and capture the graphical output yourself and return the contents with the appropriate `Content-Type` header. See the existing image renderers as a model of how to do this.

-- <https://www.rplumber.io/articles/rendering-output.html#customizing-image-serializers> viewed on 10/18/2021

According to the above documentation, a `png` serializer with dynamic image sizes can be implemented with the current `plumber` framework. This issue is a feature request for creating a friendly and stable interface for users to directly use `png` serializer with dynamic image sizes.

I attempted to implement a `png` serializer with dynamic image sizes, `dyn_param_png_serializer`, which is listed below. However, is `dyn_param_png_serializer` properly implemented? As `plumber::serializer_device` apparently also supports async execution, will `dyn_param_png_serializer` cause async errors? Will `dyn_param_png_serializer` likely be supported by future `plumber` versions? 

Additionally, expressions inside `dyn_param_png_serializer` cannot directly access variables in `/plot` endpoint function, e.g. `image_width` and `myData`, but why the passed `my_plot_func` can be called directly within `dyn_param_png_serializer`? Is there any pointers on the scope of variables in the context of `plumber` endpoint function definition, registration, runtime, etc?

```R
dyn_param_png_serializer <- function(plot_func, width, height, res) {
  # print(image_width) # <simpleError in print(image_width): object 'image_width' not found>
  tmpfn <- base::tempfile()

  grDevices::png(tmpfn, width = width, height = height, res = res)
  base::on.exit({base::unlink(tmpfn)}, add = TRUE)

  device_id <- grDevices::dev.cur()

  plot_func()

  grDevices::dev.off(device_id)

  fconn <- base::file(tmpfn, "rb")
  base::on.exit({base::close(fconn)}, add = TRUE)

  img <- base::readBin(fconn, "raw", n = base::file.info(tmpfn)$size)

  return(img)
}


#* Plot out data from the iris dataset
#* @param spec If provided, filter the data to only this species (e.g. 'setosa')
#* @get /plot
#* @serializer contentType list(type="image/png")
function(spec){
  myData <- iris
  title <- "All Species"

  # Filter if the species was specified
  if (!missing(spec)){
    title <- paste0("Only the '", spec, "' Species")
    myData <- subset(iris, Species == spec)
    image_width <- 1000
  } else {
    image_width <- 2000
  }

  my_plot_func <- function() {
[...]

PR #896, opened 2023-01-04
Dynamic Height/Width Serialization

slodge:
## Pull Request

First attempt at #837

## Still to do:

- [ ] Lots of QA - both manual and unit tests
- [ ] Lots of Docs
- [ ] Merging with the other open PRs - this is going to be a bit of a **mess** now :/
- [ ] News.md
- [ ] devtools::Check Package
- [ ] Review
- [ ] Possibly more granular control on parameters - do we want to turn things on/off per endpoint? do we want more options? (e.g. quality)? do we want less?

### To consider

* breaking backward compatibility - don't think it does - but it's possible...
* is hard to understand - possibly
* is hard to maintain in the future - possibly ... but no worse than the rest of plumber? there's some technical debt building up in the open api params I think

Have run:
```
devtools::load_all()
library(tidyverse)

#* Plot out data from the iris dataset
#* @param spec If provided, filter the data to only this species (e.g. 'setosa')
#* @get /plot
#* @serializer png
function(spec = "setosa"){
  myData <- iris
  title <- "All Species"

  # Filter if the species was specified
  if (!missing(spec)){
    title <- paste0("Only the '", spec, "' Species")
    myData <- subset(iris, Species == spec)
  }

  plot(myData$Sepal.Length, myData$Petal.Length,
       main=title, xlab="Sepal Length", ylab="Petal Length")
}
```

This shows as:
image

and:
image

Seems to work... 👍

### DIFF
--- a/R/openapi-spec.R
+++ b/R/openapi-spec.R
@@ -17,8 +17,18 @@ endpointSpecification <- function(routerEndpointEntry, path = routerEndpointEntr
   funcParams <- routerEndpointEntry$getFuncParams()
   # Get the plumber decoration defined endpoint params
   endpointParams <- routerEndpointEntry$getEndpointParams()
+
+  # if legacy plumber.staticSerializers is specified then ignore all req serializer
+  # based parameters
+  if (getOption("plumber.staticSerializers", default="FALSE") == "TRUE") {
+    serializerParams <- list()
+  } else {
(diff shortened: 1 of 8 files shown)


slodge:
(Have redone the push - as something went wrong with which email address was used in the commit)

PR #908, opened 2023-03-07
Vignette example missing Pipe

pinduzera:
Without the pipe operator, the Vignette API launch example doesn't work.


schloerke:
Thank you!

PR #922, opened 2023-11-16
Add a zip attachment example

meztez:
How to return multiple object in an archive. Adding an example. Not sure, if it should be added as a feature or the right API for it either.

```r
as_attachment <- function(value, filename, ..., .list = NULL) {
  ...
}
```
What if the user wants to use different serializers for each file?

At least this is an example for the current version of plumber.

### DIFF
--- a/inst/plumber/16-attachment/plumber.R
+++ b/inst/plumber/16-attachment/plumber.R
@@ -21,3 +21,32 @@ function() {
 function() {
   Sys.time()
 }
+
+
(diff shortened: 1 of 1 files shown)


schloerke:
> What if the user wants to use different serializers for each file?

They'll need to handle it themselves in the route

schloerke:
> At least this is an example for the current version of plumber.

Yes. Even if we made a zip serializer, they would need to construct a folder with the content ready to go. It still leaves a clunky user interaction, so an example of how it's possible would be better.

Issue #915, opened 2023-08-24
Calling POST from C# code does not process body correctly


MJSchut:
### System details

Output of `sessioninfo::session_info()()`:

```
> sessioninfo::session_info()
- Session info -----------------------------------------------------------------------------------------------------------------------------------------
 setting  value
 version  R version 3.6.3 (2020-02-29)
 os       Windows 10 x64 (build 19044)
 system   x86_64, mingw32
 ui       RStudio
 language (EN)
 collate  English_United States.1252
 ctype    English_United States.1252
 tz       Europe/Berlin
 date     2023-08-24
 rstudio  2023.06.1+524 Mountain Hydrangea (desktop)
 pandoc   3.1.1 @ C:/Users/Martijn.Schut/AppData/Local/Programs/RStudio/resources/app/bin/quarto/bin/tools/ (via rmarkdown)

- Packages ---------------------------------------------------------------------------------------------------------------------------------------------
 ! package       * version  date (UTC) lib source
 1 abind           1.4-5    2016-07-21 [1] RSPM (R 3.6.3)
 1 alluvial      * 0.1-2    2016-09-09 [1] RSPM (R 3.6.2)
 1 anytime       * 0.3.9    2020-08-27 [1] RSPM (R 3.6.3)
 1 arrow         * 9.0.0    2022-08-10 [1] RSPM (R 3.6.3)
 1 askpass         1.1      2019-01-13 [1] RSPM (R 3.6.3)
 1 assertthat      0.2.1    2019-03-21 [1] RSPM (R 3.6.3)
 1 automap       * 1.0-16   2022-04-05 [1] RSPM (R 3.6.3)
 1 AzureAuth     * 1.3.3    2021-09-13 [1] RSPM (R 3.6.3)
 1 AzureGraph      1.3.2    2021-11-28 [1] RSPM (R 3.6.3)
 1 AzureRMR        2.4.3    2021-10-23 [1] RSPM (R 3.6.3)
 1 AzureStor     * 3.7.0    2022-05-25 [1] RSPM (R 3.6.3)
 1 backports       1.4.1    2021-12-13 [1] RSPM (R 3.6.3)
 1 base64enc       0.1-3    2015-07-28 [1] CRAN (R 3.6.3)
 1 BBmisc          1.13     2022-09-29 [1] RSPM (R 3.6.3)
 1 bit             4.0.5    2022-11-15 [1] RSPM (R 3.6.3)
[...]

Issue #876, opened 2022-08-22
multipart boundary support surrounding quotes


meztez:
In:
https://github.com/rstudio/plumber/blob/92532ffc1d7188b2841235dc4416178c57e51b8e/R/parse-body.R#L521

plumber should be able to support boundary in surrounded double quotes. Wouldhave to modify regex or trim quotes before forwarding to webutils::parse_multipart.

https://www.rfc-editor.org/rfc/rfc7578#section-4.1

meztez:
```r
...
boundary <- gsub('^"|"$', "", boundary)
...
```

PR #924, opened 2023-11-17
Add support for quoted boundary for multipart request parsing.

meztez:
See #915, thanks @slodge for finding the problem and @MJSchut for pointing out the issue.

Per [RFC2046](https://www.ietf.org/rfc/rfc2046.txt)

The issue will be present in [jeroen/webutils](https://github.com/jeroen/webutils) too.

> WARNING TO IMPLEMENTORS:  The grammar for parameters on the Content-
>    type field is such that it is often necessary to enclose the boundary
>    parameter values in quotes on the Content-type line.  This is not
>    always necessary, but never hurts. Implementors should be sure to
>    study the grammar carefully in order to avoid producing invalid
>    Content-type fields.  Thus, a typical "multipart" Content-Type header
>    field might look like this:
> 
>      Content-Type: multipart/mixed; boundary=gc0p4Jq0M2Yt08j34c0p
> 
>    But the following is not valid:
> 
>      Content-Type: multipart/mixed; boundary=gc0pJq0M:08jU534c0p
> 
>    (because of the colon) and must instead be represented as
> 
>      Content-Type: multipart/mixed; boundary="gc0pJq0M:08jU534c0p"
> 
>    This Content-Type value indicates that the content consists of one or
>    more parts, each with a structure that is syntactically identical to
>    an RFC 822 message, except that the header area is allowed to be
>    completely empty, and that the parts are each preceded by the line

PR task list:
- [x] Update NEWS
- [x] Add tests

### DIFF
--- a/R/parse-body.R
+++ b/R/parse-body.R
@@ -519,6 +519,7 @@ parser_multi <- function() {
     if (!stri_detect_fixed(content_type, "boundary=", case_insensitive = TRUE))
       stop("No boundary found in multipart content-type header: ", content_type)
     boundary <- stri_match_first_regex(content_type, "boundary=([^; ]{2,})", case_insensitive = TRUE)[,2]
+    boundary <- stri_replace_first_regex(boundary, '^"([^"]+)"?', "$1")
     toparse <- parse_multipart(value, boundary)
 
     # set the names of the items as the `name` of each item
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # plumber (development version)
 
+* Add support for quoted boundary for multipart request parsing. (@meztez #924)
 
 # plumber 1.2.1
 
--- a/tests/testthat/test-parse-body.R
+++ b/tests/testthat/test-parse-body.R
@@ -261,6 +261,12 @@ test_that("Test multipart parser", {
   expect_true(is.raw(body_args[["img2"]][["ragnarok_small.png"]]))
   expect_gt(length(body_args[["img2"]][["ragnarok_small.png"]]), 100)
   expect_equal(body_args[["json"]], list(a=2,b=4,c=list(w=3,t=5)))
+
+  # Quoted boundary
+  req$HTTP_CONTENT_TYPE = "multipart/form-data; boundary=\"----WebKitFormBoundaryMYdShB9nBc32BUhQ\""
+  req$body <- req_body_parser(req, make_parser(c("multi", "json", "rds", "octet")))
+  expect_equal(names(req$body), c("json", "img1", "img2", "rds"))
+
 })
 
 


slodge:
Is there any advantage in splitting this into two regex operations? The original fix suggested was:

https://github.com/slodge/plumber/commit/f0e37b82b189d34fdaa2d8f6972a895193dd5d9e

meztez:
making sure quotes are either there or not, original pattern would still match if only one quote appears either at the front or back of the boundary

slodge:
I don't believe doublequotes are valid boundary characters - so that shouldn't be a problem:

image

from Appendix A of https://www.ietf.org/rfc/rfc2046.txt

Not a big problem/concern - so will leave just as comment 👍

meztez:
yeah, seems the fix I proposed in #876 was just a straight up gsub. Still, can't be too cautious, nothing is currently preventing a client from using one as the pattern does not check the boundary character or the boundary length.

R/parse-body.R
```diff
@@ -519,6 +519,7 @@ parser_multi <- function() {
     if (!stri_detect_fixed(content_type, "boundary=", case_insensitive = TRUE))
       stop("No boundary found in multipart content-type header: ", content_type)
     boundary <- stri_match_first_regex(content_type, "boundary=([^; ]{2,})", case_insensitive = TRUE)[,2]
+    boundary <- stri_replace_first_regex(boundary, '^"([^"]+)"?', "$1")
```
schloerke, line 522: I like your implementation that you did in jeroen/webutils#8. I compared this implementation, your issue suggestion, and the webutils PR. The simple `gsub` command was the fastest, but no guarantees a start and end quote exist. 

So we should stay with the current implementation.

```r
boundary <- "\"----WebKitFormBoundaryMYdShB9nBc32BUhQ\""
bench::mark(
  stri_replace_first_regex(boundary, '^"([^"]+)"$', "$1"),
  gsub("^\"|\"$", "", boundary),
  sub('^"(.*)"$', "\\1", boundary)
)
#> # A tibble: 3 × 13
#>   expression        min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory     time       gc
#>   <bch:expr>     <bch:> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>     <list>     <list>
#> 1 "stri_replace… 3.65µs 3.81µs   253590.        0B      0   10000     0     39.4ms <chr>  <Rprofmem> <bench_tm> <tibble>
#> 2 "gsub(\"^\\\"… 3.36µs 3.61µs   268517.        0B      0   10000     0     37.2ms <chr>  <Rprofmem> <bench_tm> <tibble>
#> 3 "sub(\"^\\\"(… 3.81µs 4.02µs   238138.        0B     23.8  9999     1       42ms <chr>  <Rprofmem> <bench_tm> <tibble>
```

Adjusting the regex to force a quote to exist at the end. (If it is improperly formed, I don't know how it'll be parsed in the first place.)

```suggestion
    # Remove surrounding quotes if they exist
    boundary <- stri_replace_first_regex(boundary, '^"([^"]+)"$', "$1")
```

meztez:
```r
```
> bench::mark(iterations = 1000000,
+   stri_replace_first_regex(boundary, '^"(.*)"$', "$1"),
+   gsub("^\"|\"$", "", boundary),
+   stri_replace_first_regex(boundary, '^"([^"]+)"$', "$1"),
+   sub('^"(.*)"$', "\\1", boundary)
+ )
# A tibble: 4 × 13
  expression    min median `itr/sec` mem_alloc `gc/sec`  n_itr  n_gc total_time result memory     time      
  <bch:expr> <bch:> <bch:>     <dbl> <bch:byt>    <dbl>  <int> <dbl>   <bch:tm> <list> <list>     <list>    
1 "stri_rep… 3.81µs 4.42µs   210046.        0B     7.35 999965    35      4.76s <chr>  <Rprofmem> <bench_tm>
2 "gsub(\"^… 4.08µs 4.78µs   194017.        0B     8.34 999957    43      5.15s <chr>  <Rprofmem> <bench_tm>
3 "stri_rep… 4.82µs 6.31µs   151239.        0B     3.02 999980    20      6.61s <chr>  <Rprofmem> <bench_tm>
4 "sub(\"^\… 5.67µs 6.28µs   146406.        0B     6.73 999954    46      6.83s <chr>  <Rprofmem> <bench_tm>
# ℹ 1 more variable: gc <list>

R/parse-body.R
```diff
@@ -519,6 +519,8 @@ parser_multi <- function() {
     if (!stri_detect_fixed(content_type, "boundary=", case_insensitive = TRUE))
       stop("No boundary found in multipart content-type header: ", content_type)
     boundary <- stri_match_first_regex(content_type, "boundary=([^; ]{2,})", case_insensitive = TRUE)[,2]
+    # Remove surrounding quotes if they exist
+    boundary <- stri_replace_first_regex(boundary, '^"([^"]+)"$', "$1")
```
meztez, line 523:
```suggestion
    boundary <- stri_replace_first_regex(boundary, '^"(.*)"$', "$1")
```

schloerke:
Addressing https://github.com/rstudio/plumber/pull/924#issuecomment-1815788444 by comparing a single call, two calls, and webutils implementation...

```r
content_type <- "multipart/form-data; boundary=\"----WebKitFormBoundaryMYdShB9nBc32BUhQ\""; 
bench::mark(
  stri_match_first_regex(content_type, "boundary=\"?([^; \"]{2,})\"?", case_insensitive = TRUE)[,2],
  {
    boundary <- stri_match_first_regex(content_type, "boundary=([^; ]{2,})", case_insensitive = TRUE)[,2]
    boundary <- stri_replace_first_regex(boundary, '^"(.*)"$', "$1")
  },
  {
    m <- regexpr('boundary=[^; ]{2,}', content_type, ignore.case = TRUE)
    boundary <- sub('boundary=','',regmatches(content_type, m)[[1]])
    sub('^"(.*)"$', "\\1", boundary)
  }
)
# A tibble: 3 × 13
  expression        min median `itr/sec` mem_alloc `gc/sec` n_itr  n_gc total_time result memory     time       gc
  <bch:expr>    <bch:t> <bch:>     <dbl> <bch:byt>    <dbl> <int> <dbl>   <bch:tm> <list> <list>     <list>     <list>
1 "stri_match_…  8.77µs   10µs    90181.        0B     18.0  9998     2      111ms <chr>  <Rprofmem> <bench_tm> <tibble>
2 "{ boundary … 12.01µs 12.5µs    77089.        0B     15.4  9998     2      130ms <chr>  <Rprofmem> <bench_tm> <tibble>
3 "{ m <- rege… 19.76µs 20.7µs    45794.        0B     18.3  9996     4      218ms <chr>  <Rprofmem> <bench_tm> <tibble>
```

There is a 2.5µs speed increase. So, we should use the single statement as both calls are being done.

R/parse-body.R
```diff
@@ -519,6 +519,8 @@ parser_multi <- function() {
     if (!stri_detect_fixed(content_type, "boundary=", case_insensitive = TRUE))
       stop("No boundary found in multipart content-type header: ", content_type)
     boundary <- stri_match_first_regex(content_type, "boundary=([^; ]{2,})", case_insensitive = TRUE)[,2]
+    # Remove surrounding quotes if they exist
+    boundary <- stri_replace_first_regex(boundary, '^"(.*)"$', "$1")
```
schloerke, line 523:
```suggestion
    # Remove surrounding quotes if they exist
    boundary <- stri_match_first_regex(content_type, "boundary=\"?([^; \"]{2,})\"?", case_insensitive = TRUE)[,2]
```

meztez: @schloerke Do you want to enforce the 70 character delimiter limit?

...{2,70}?

schloerke:
Does this also fix #915? I can't tell from the issue

meztez:
> Does this also fix #915? I can't tell from the issue

It should. The C# httpclient surrounds the boundary in dquotes.

PR #930, opened 2023-11-18
Prevent parseUTF8 from overwriting srcfile attribute

### LINKED ISSUE #916: in parseUTF8() on Windows with enc "unknown", src is ignored / / overwritten
```R
> sessioninfo::session_info()
─ Session info ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 setting  value
 version  R version 4.3.1 (2023-06-16 ucrt)
 os       Windows 10 x64 (build 19045)
 system   x86_64, mingw32
 ui       RStudio
 language (EN)
 collate  English_Canada.utf8
 ctype    English_Canada.utf8
 tz       America/Toronto
 date     2023-09-07
 rstudio  2023.06.1+524 Mountain Hydrangea (desktop)
 pandoc   2.12 @ C:\\Users\\iris\\anaconda3\\Scripts\\pandoc.exe

─ Packages ───────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
 package     * version  date (UTC) lib source
 cli           3.6.1    2023-03-23 [1] CRAN (R 4.3.0)
 crayon        1.5.2    2022-09-29 [1] CRAN (R 4.3.0)
 essentials  * 0.3.0.13 2023-08-01 [1] local
 jsonlite      1.8.5    2023-06-05 [1] CRAN (R 4.3.1)
 later         1.3.1    2023-05-02 [1] CRAN (R 4.3.0)
 lifecycle     1.0.3    2022-10-07 [1] CRAN (R 4.3.0)
 magrittr      2.0.3    2022-03-30 [1] CRAN (R 4.3.0)
 plumber       1.2.1    2022-09-06 [1] CRAN (R 4.3.1)
 promises      1.2.0.1  2021-02-11 [1] CRAN (R 4.3.0)
 R6            2.5.1    2021-08-19 [1] CRAN (R 4.3.0)
 Rcpp          1.0.11   2023-07-06 [1] CRAN (R 4.3.1)
 rlang         1.1.1    2023-04-28 [1] CRAN (R 4.3.1)
 rstudioapi    0.15.0   2023-07-07 [1] CRAN (R 4.3.1)
 sessioninfo   1.2.2    2021-12-06 [1] CRAN (R 4.3.1)
 stringi       1.7.12   2023-01-11 [1] CRAN (R 4.3.0)
 swagger       3.33.1   2020-10-02 [1] CRAN (R 4.3.1)
 this.path   * 2.0.0.7  2023-09-07 [1] local
 webutils      1.1      2020-04-28 [1] CRAN (R 4.3.1)

 [1] C:/Users/iris/AppData/Local/R/win-library/4.3
 [2] C:/Program Files/R/R-4.3.1/library

──────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────────
>
```

```R
if (.Platform$OS.type != "windows") {
    stop("windows only bug")
} else local({
    FILE.R <- tempfile(fileext = ".R")
    on.exit(unlink(FILE.R))
    writeLines("5 + 6", FILE.R)
    exprs <- plumber:::parseUTF8(FILE.R)
    cat("original filename: ", FILE.R, "\n", sep = "")
    cat("srcfile  filename: "); print(attr(exprs, "srcfile"))
})
```

In `parseUTF8()` the comments claim that on Windows with encoding `"unknown"` the file must be re-encoded to native. I would agree with this, though I would think this is no longer necessary since the C runtime on Windows is now ucrt and the native encoding is now UTF-8. Regardless, it writes the lines into a new file in the native encoding.

It then claims that despite parsing a different file, the source reference is pointed to the original file. This is incorrect since `parse()` overwrites its argument `srcfile` when `file` is a character string and `keep.source = TRUE`.

To avoid `srcfile` being overwritten when passed to `parse()`, you should use `parse(keep.source = FALSE)` (unintuitively).

Here is some code showing the fix working as intended:

```R
my_parseUTF8 <- function (file)
{
    lines <- plumber:::readUTF8(file)
    src <- srcfilecopy(file, lines, isFile = TRUE)
    file <- tempfile()
    on.exit(unlink(file))
    writeLines(lines, file)
    parse(file, keep.source = FALSE, srcfile = src)
}


if (.Platform$OS.type != "windows") {
    stop("windows only bug")
} else local({
    FILE.R <- tempfile(fileext = ".R")
    on.exit(unlink(FILE.R))
    writeLines("5 + 6", FILE.R)
    exprs <- my_parseUTF8(FILE.R)
    cat("original filename: ", FILE.R, "\n", sep = "")
    cat("srcfile  filename: "); print(attr(exprs, "srcfile"))
})
```

meztez:
Fix #916, related to `parseUTF8` return value attribute `srcfile` on Windows.

PR task list:
- [x] Update NEWS
- [x] Add tests

### DIFF
--- a/NEWS.md
+++ b/NEWS.md
@@ -1,5 +1,6 @@
 # plumber (development version)
 
+* Fix #916, related to `parseUTF8` return value attribute `srcfile` on Windows. (#930) 
 
 # plumber 1.2.1
 
--- a/R/utf8.R
+++ b/R/utf8.R
@@ -66,15 +66,13 @@ parseUTF8 <- function(file) {
   # **can** be encoded natively on Windows (might be a bug in base R); we
   # rewrite the source code in a natively encoded temp file and parse it in this
   # case (the source reference is still pointed to the original file, though)
-  keepsource <- TRUE
   if (isWindows() && enc == 'unknown') {
     file <- tempfile(); on.exit(unlink(file), add = TRUE)
     writeLines(lines, file)
-    keepsource <- FALSE
   }
 
   # keep the source locations within the file while parsing
-  exprs <- try(parse(file, keep.source = keepsource, srcfile = src, encoding = enc))
+  exprs <- try(parse(file, keep.source = FALSE, srcfile = src, encoding = enc))
   if (inherits(exprs, "try-error")) {
     stop("Error sourcing ", file)
   }
--- a/R/utf8.R
+++ b/R/utf8.R
@@ -66,13 +66,15 @@ parseUTF8 <- function(file) {
   # **can** be encoded natively on Windows (might be a bug in base R); we
   # rewrite the source code in a natively encoded temp file and parse it in this
   # case (the source reference is still pointed to the original file, though)
+  keepsource <- TRUE
   if (isWindows() && enc == 'unknown') {
     file <- tempfile(); on.exit(unlink(file), add = TRUE)
     writeLines(lines, file)
+    keepsource <- FALSE
   }
-  
+
   # keep the source locations within the file while parsing
-  exprs <- try(parse(file, keep.source = TRUE, srcfile = src, encoding = enc))
+  exprs <- try(parse(file, keep.source = keepsource, srcfile = src, encoding = enc))
   if (inherits(exprs, "try-error")) {
     stop("Error sourcing ", file)
   }
--- a/tests/testthat/test-utf8.R
+++ b/tests/testthat/test-utf8.R
@@ -1,0 +1,4 @@
+test_that("parseUTF8 has same srcfile attribute as file arg input", {
+  filename <- test_path("files/plumber.R")
+  expect_equal(attr(parseUTF8(file), "srcfile")$filename, filename)
+})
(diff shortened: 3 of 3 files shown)


ArcadeAntics:
This fixes the issue of the source file pointing to the correct file, but you've basically made a `srcfilecopy` just for `parse` to overwrite it with the exact same input. You need to directly supply `keep.source = FALSE` always.

ArcadeAntics:
Yes, perfect! Thank you so much!

meztez:
> This fixes the issue of the source file pointing to the correct file, but you've basically made a `srcfilecopy` just for `parse` to overwrite it with the exact same input. You need to directly supply `keep.source = FALSE` always.

I saw that, I thought maybe there was a reason for it? I will let @schloerke comment on this behaviour since it seems the line came from a commit he authored, but I've taken your comment into account for this PR.

Shiny has keep.source set to FALSE with similar instructions in `sourceUTF8`.
https://github.com/rstudio/shiny/blob/b1297395a9365c97972f243ee492232d70af281f/R/utils.R#L1355

PR #934, opened 2023-11-25
Fall back to sys env var if R option doesn't exist

meztez:
…umber`.

PR task list:
- [x] Update NEWS
- [x] Add tests

Setting an option that requires a function using an environment variable with most likely fail. Environment variable are characters.


R/options_plumber.R
```diff
@@ -89,3 +91,24 @@ options_plumber <- function(
...
+    env_name <- toupper(chartr(".", "_", x))
+    res <- Sys.getenv(env_name)
+    if (res == "") {
+      default
+    } else {
+      if (res %in% c("TRUE", "FALSE")) {
+        return(as.logical(res))
+      }
+      res
+    }
```
schloerke, line 111: Cosmetic

```suggestion
      return(default)
    }

    if (res %in% c("TRUE", "FALSE")) {
      return(as.logical(res))
    }
    res
```

schloerke (changes requested):

R/options_plumber.R
```diff
@@ -93,22 +93,21 @@ options_plumber <- function(
 }
 
 #' Get an option value, alternatively look in environment for value
-#' @rdname options_plumber
+#' @describeIn options_plumber DOCS
```
schloerke, line 96: Going to revert this in favor of future PR using `{options}`.

```suggestion
#' @rdname options_plumber
```

schloerke:
> I'll do a second PR, we can work from there to choose which one makes the most sense.

Yes, it looks promising! https://dgkf.github.io/options/articles/options.html

We could probably soft deprecate `options_plumber()` (as it was mainly a place for documentation for all options) or just have `options_plumber()` use `options::opt`/`options::opts` internally. We could then mark `options_plumber` as `@keywords internal` (and drop from other docs) to favor the `{options}` package usage.

man/options_plumber.Rd
```diff
@@ -76,3 +87,8 @@ When \code{NULL}, secret is not validated. Otherwise, Plumber compares secret wi
 by settings this option to \code{FALSE}. Defaults to \code{TRUE}}
 }
 }
+\section{Functions}{
+\itemize{
+\item \code{get_option_or_env()}: DOCS
+
+}}
```
schloerke, line 94:
```suggestion
```

PR #936, opened 2023-11-28
Attempting to use options package.

meztez:
Pending a doc ordering issue in options package.

https://github.com/dgkf/options/pull/7

PR task list:
- [ ] Update NEWS
- [ ] Add tests


meztez:
I'm thinking using the options package would maybe open up an attack vector, because of this line here:
https://github.com/dgkf/options/blob/a2c110229f844de49b3dbb56bb0a19542b92899a/R/envvars.R#L168

I'm going to ice this one for now.

schloerke:
The `option_spec()` handles the [extra args supplied to `define_option.character()`](https://github.com/dgkf/options/blob/a2c110229f844de49b3dbb56bb0a19542b92899a/R/define_option.R#L69). 

So we could use a helper function to define the options (setting the `envvar_fn` value for every option):

```r
# Copied from https://github.com/dgkf/options/blob/a2c1102/R/envvars.R#L1-L4
# Sets up pretty printing
fn_with_desc <- function(f, desc) {
  attr(f, "desc") <- desc
  f
}
# Keep sys env var as raw string only
plumber_envvar_raw <- fn_with_desc(
  function(raw, name, ...) {
    raw
  },
  "string"
)
define_plumber_option <- function(...) {
  options::define_option(
    ...,
    envvar_fn = plumber_envvar_raw
  )
}
```

```r
# Ex usage
define_plumber_option(
  option = "port",
  default = NULL,
  desc = paste(
    "Port Plumber will attempt to use to start http server.",
    "If the port is already in use, server will not be able to start."
  )
)
```

Issue #950, opened 2024-04-26
fix request object table layout in routing vignette


maxheld83:
[the request object table](https://www.rplumber.io/articles/routing-and-input.html#the-request-object) wraps weirdly for the last columns

<img width="1538" alt="Screenshot 2024-04-26 at 18 26 06" src="https://github.com/rstudio/plumber/assets/5372770/4dc356fb-1d2a-47a8-a427-309d7175790c">

This is macOS Safari 17.4.1 (actual size, no zoom).
Same in Chrome 124.
So probably a markdown issue.

PR #951, opened 2024-04-26
fix table spacing

maxheld83:
closes https://github.com/rstudio/plumber/issues/950

This is just cosmetics, so don't think the checklist (`NEWS.md` etc.) applies.

I just applied the tidyverse style guide, which also makes the table look better.

*before*:
Screenshot 2024-04-26 at 18 26 00

*after*:
Screenshot 2024-04-26 at 19 03 45


schloerke:
Thank you!

PR #971, opened 2025-01-23
Add note on annotating required parameters

### LINKED ISSUE #957: Required parameter annotation
The plumber docs (as far as I can tell) do not show how to mark a parameter as required (or optional) without using the programmatic usage.

I would prefer to use the comment annotation as it is far easier to write and read. 

Marking a parameter as optional is alluded to in the examples at https://www.rplumber.io/articles/annotations.html#equivalent-programmatic-usage-1. This is only shown for the programatic usage, however.

thomasp85:
Fix #957

This PR only adds to the annotation vignette. If there are other places it would make sense to add a remark around required parameters let me know


schloerke:
Docs are fairly sparse at ~~some~~ many points given the broad topic area of `{plumber}`. This looks good for now!

Thank you!

PR #976, opened 2025-01-24
Update default debug behaviour

### LINKED ISSUE #962: enh: Change the default value of `debug` to a FALSE-like sentinel value
There are multiple reports of confusion around `debug` and the default error handler. 
* Errors are displayed locally when `interactive()`
* Errors are not displayed when deployed

Default `debug` value:
https://github.com/rstudio/plumber/blob/7f51e24bb6e6050c7743ed6c98015ecf42b50527/R/plumber.R#L1338-L1340

## Proposal

Change the default value of `debug` to behave like a `FALSE` value. 

To maintain least surprise (default to same locally, display a message with instructions on how to turn in on:
```r
Error reporting has been turned off by default. See `pr_set_debug` for more details. 
To disable this message, setting a debug value.
```

## Tasks

- [ ] Change the default value of `debug` to behave like a `FALSE` value. 
- [ ] Add `rlang::inform(.frequency="once", .frequency_id="pr_set_debug_message")` message. 
	- This should only be displayed if `rlang::is_interactive()` is `TRUE`.
- [ ] Add `#* @plumber` example within `pr_set_debug()` docs to show to to set the value within a plumbed file
- [ ] News entry of changed behavior

thomasp85:
Fix #962 

As discussed in the issue, this PR changes the default behaviour of debugging to no longer be contingent on whether plumber is running interactively or not. It will now always be false if not set (either in `$run()` or with `pr_set_debug()`), but will give a message (once per session) if using the default (FALSE) and running interactively()


R/plumber.R
```diff
@@ -267,6 +266,14 @@ Plumber <- R6Class(
         on.exit(unmount_docs(self, docs_info), add = TRUE)
       }
 
+      if (!isTRUE(quiet) && inform_debug && rlang::is_interactive()) {
+        rlang::inform(
+          "Error reporting has been turned off by default. See `pr_set_debug()` for more details.\nTo disable this message, set a debug value.",
```
schloerke, line 271: Thank you

schloerke (approved):

PR #977, opened 2025-01-24
Actually use apiPath when mounting the docs

### LINKED ISSUE #836: global basepath
It would be helpful to be able to specify a global base path e.g. as an argument to `pr_run()`, which would be taken into account by all routers including the documentation and spec urls.

I'm currently working around this by just creating an empty root router `pr()` and using `pr_mount()` to add my actual router under a base path. This is not a complete solution since I would also like to serve my documentation under that same base path.

I found `options(plumber.apiPath)`, which looks like it's working, but the docs are actually served under `http://127.0.0.1:5297/__docs__/`
```
>  plumber::options_plumber(apiPath = "/x")
>  plumber::pr() %>% plumber::pr_run()
Running plumber API at http://127.0.0.1:5297
Running swagger Docs at http://127.0.0.1:5297/x/__docs__/
```

thomasp85:
Fix #836 

The `apiPath` option was never used during the actual mounting of the docs router - only for reporting the url of it.

This is now fixed. 

@schloerke can you please advise me where tests for this bug (and the fix) are best placed?

### DIFF
--- a/R/plumber.R
+++ b/R/plumber.R
@@ -254,7 +254,6 @@ Plumber <- R6Class(
         old_wd <- setwd(dirname(private$filename))
         on.exit({setwd(old_wd)}, add = TRUE)
       }
-
       if (isTRUE(docs_info$enabled)) {
         mount_docs(
           pr = self,
@@ -264,7 +263,7 @@ Plumber <- R6Class(
           callback = swaggerCallback,
           quiet = quiet
         )
-        on.exit(unmount_docs(self, docs_info), add = TRUE)
+        #on.exit(unmount_docs(self, docs_info), add = TRUE)
       }
 
       on.exit(private$runHooks("exit"), add = TRUE)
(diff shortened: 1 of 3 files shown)


R/plumber.R
```diff
@@ -264,7 +263,7 @@ Plumber <- R6Class(
           callback = swaggerCallback,
           quiet = quiet
         )
-        on.exit(unmount_docs(self, docs_info), add = TRUE)
+        #on.exit(unmount_docs(self, docs_info), add = TRUE)
```
schloerke, line 266: I believe this still needs to be there as the docs will want to be mounted again on a followup run.

schloerke (changes requested):

schloerke:
Instead of altering just the docs, should we adjust how `pr$handle()` and `pr$mount()` will prepend the path portion of https://github.com/rstudio/plumber/blob/5980d1e2ebab3a63c058813931cd0e16e6690e16/R/ui.R#L12-L22

If we do not apply it internally to the two methods, we should still use the path portion of the `api_url` when mounting the docs.

PR #983, opened 2025-02-15
Update vignette's link 

cgiachalis:
New link address is : `https://www.restapitutorial.com/introduction/httpmethods` instead of  `http://www.restapitutorial.com/lessons/httpmethods.html`


schloerke:
Thank you for the updated link!

PR #986, opened 2025-02-18
bug: Validate APIs using `@redocly/cli`

schloerke:
`swagger-cli` was throwing an error about being deprecated. The error suggested to use `@redocly/cli`. 

`@redocly/cli` has an opinionated lint method. I've disabled rules in which I do not believe they play nicely with plumber. We can re-enable them at a later date.

Enhancements:
* Added `operationId` in the shape of `PATH-VERB` to each operation in the swagger doc.

```r
validate_api_spec()
```

PR task list:
- [x] Update NEWS
- [x] Add tests


schloerke:
Merging as it is not a critical path and resolves bad output from tests

PR #1007, opened 2025-12-11
fix(docker): Use `pak` to install plumber

schloerke:
All docker builds are failing, independent of plumber version. Switching to pak in hopes it installs any missing system deps

Screenshot 2025-12-11 at 4 33 32 PM


schloerke:
macos failure unrelated to PR. Merging as docker jobs are fixed

